PR #56, opened 2018-04-13
I-D Template

mnot:
I think this does most everything to get over to @martinthomson's I-D template. Martin you might want to look, as I suspect I've done horrible things.

After merge, TODO:

* [ ] Setup circle integration
* [ ] Verify gh-pages update works 
* [ ] Put pages on `gh-pages`, not `master`
* [ ] Verify submission works
* [ ] Move Docker package away from mnot/tt
* [ ] Automate additional lints / tasks.


mnot:
@reschke I've fixed your issues. Thanks.

WRT diffs -- I've added the `diffs` target from the current repo as `diff_auth48'; the template already has a `diff` target that does what you want (try it once we've tagged -00).

@fielding we'll have it sorted by the time you taxes are done, don't worry.

tools/redefinitions.mk
```diff
@@ -0,0 +1,15 @@
+saxpath = "$(HOME)/java/saxon-9-7/saxon9he.jar"
```
martinthomson: Maybe you should just warn if this isn't set in the environment.  The circle build sets it and anyone using this can as well.

tools/redefinitions.mk
```diff
@@ -0,0 +1,15 @@
...
+xreffer    = $(tools)/xreffer.xslt
+stylesheet = $(tools)/myxml2rfc.xslt
+
+clean::
+	rm -f diff*.html
+
+%.cleanxml: %.xml $(tools)/clean-for-DTD.xslt abnf
+	$(saxon) $< $(xreffer) | $(saxon) - $(tools)/clean-for-DTD.xslt > $@
+
+%.htmltmp: %.xml $(stylesheet) abnf
```
martinthomson: I can see why you want to override these.  I think that this is reasonable.

tools/core.mk
```diff
@@ -0,0 +1,106 @@
+
+rfcdiff = rfcdiff --width 78 --stdout
+reduction  = $(tools)/clean-for-DTD.xslt
```
martinthomson: You don't use this here.

tools/core.mk
```diff
@@ -0,0 +1,106 @@
+
+rfcdiff = rfcdiff --width 78 --stdout
```
martinthomson: I'd take a PR for that change.

tools/core.mk
```diff
@@ -0,0 +1,106 @@
...
+MESSAGING   = $(draftname)-messaging
+RANGE       = $(draftname)-range
+SEMANTICS   = $(draftname)-semantics
+
+CORE_XML = $(MESSAGING).xml $(SEMANTICS).xml
+TARGETS_ABNF= $(addprefix $(bd)/,$(CORE_XML:.xml=.abnf))
+TARGETS_ABNFAPPENDIX= $(TARGETS_ABNF:.abnf=.abnf-appendix)
+TARGETS_PARSEDABNF= $(TARGETS_ABNF:.abnf=.parsed-abnf)
+
+ABNF =    $(TARGETS_ABNF) \
```
martinthomson: It looks like these files are still being checked-in to the repo.  That seems problematic.

tools/core.mk
```diff
@@ -0,0 +1,106 @@
...
+$(bd)/%.iana-warn-codes: %.xml $(tools)/extract-warn-code-defs.xslt
+	$(saxon) $< $(tools)/extract-warn-code-defs.xslt > $@
+
+$(bd)/%.cache-directives: %.xml $(tools)/extract-cache-directives.xslt
+	$(saxon) $< $(tools)/extract-cache-directives.xslt > $@
+
+outlineALL.html: $(TARGETS_XHTML) $(tools)/extractOutline.xslt
+	$(saxon) $(tools)/extractOutline.xslt $(tools)/extractOutline.xslt > $@
+
+httpbis.abnf: $(TARGETS_ABNF)
```
martinthomson: Is this in .gitignore?

martinthomson: If it's a build product, you will know it changes when the source changes.  Having to review changes with generated output and source is difficult.

mnot:
Well, we've got the other build products included...

tools/docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM ubuntu:16.04
+LABEL maintainer="Martin Thomson <martin.thomson@gmail.com>"
```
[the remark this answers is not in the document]
mnot: fixed.

tools/redefinitions.mk
```diff
@@ -0,0 +1,15 @@
...
+saxon = java -classpath $(saxpath) net.sf.saxon.Transform -l -versionmsg:off -ext:on
+tools = tools
+
+xreffer    = $(tools)/xreffer.xslt
+stylesheet = $(tools)/myxml2rfc.xslt
+
+clean::
+	rm -f diff*.html
+
+%.cleanxml: %.xml $(tools)/clean-for-DTD.xslt abnf
```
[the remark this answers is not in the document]
mnot: fixed

tools/core.mk
mnot: fixed

tools/core.mk
reschke: It is difficult, but it's a guard against unintended ABNF changes.

martinthomson: You can use CI for that.

tools/redefinitions.mk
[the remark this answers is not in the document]
martinthomson: but only in one place (the dependency needs tweaking too)

PR #106, opened 2018-06-25
remove fragment from http(s) URI ABNF (closes #103)

### LINKED ISSUE #103: RFC 7230: ABNF for http(s) URIs contains fragment
See https://www.rfc-editor.org/errata/eid4251 and https://www.rfc-editor.org/errata/eid4252

reschke:
....then discuss fragments in separate section.


royfielding:
Do we really need to add that section? It seems to contradict itself. I'd rather just remove them, or (at most) add a note saying that fragment is an optional component of URI-reference and not part of the scheme definition.

reschke:
The intent was to provide a place we can point people to when they ask about fragment identifiers...

What is the contradiction? Maybe we can fix it...

royfielding:
The section itself is a contradiction

   http and https Fragment Idenfifiers

They aren't http and https fragment identifiers.  They are URI reference fragment identifiers, which is why they weren't included in RFC2616's definition of URLs.  Changing that in 7230 was just a mistake.

We already talk about fragments in the URI section:

  https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#uri

and in target:

  https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#target.resource

We can add more text there (or in URI), if it is needed.

reschke:
>  Creating a new section only encourages folks to think that the fragment is part of the URI scheme definition, which is how we fell into this trap in the first place.

The section consists of three sentences:

1. The introduction, derived from the text taken out of the http URI scheme registration.
2. Prose explaining that fragment identifiers are only allowed in certain protocol elements.
3. A statement clarifying that the fragment component is not part of the URI scheme.

The last point should actually *avoid* what you are concerned with.

We can of course argue where to put these bits. I strongly believe that a separate section is best so it can easily be referred to.

royfielding:
If we put a new section under resources with a heading that implies fragments are part of the scheme definition, it places an unnecessary spotlight on a technical mistake which no longer exists in the spec. It also misleads folks who only read the headings and not the fine print (we know they exist).

The current protocol should be cleanly defined.  Changes like this one should be explained in the "Changes since RFC7230" section.

mnot:
I was envisioning a small paragraph under each chunk of ABNF that explains what's happening, nothing more.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -811,6 +805,29 @@ Content-Type: text/plain
 </t>
 </section>
 
+<section title="Fragment Identifiers on http(s) URIs" anchor="uri.fragment.identifiers">
```
royfielding: Title should be "Fragment Identifiers on http(s) URI References"

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -811,6 +805,29 @@ Content-Type: text/plain
...
+  <iref item="Fragment Identifiers"/>
+<t>
+   Fragment identifiers allow for indirect identification
+   of a secondary resource, independent of the URI scheme, as defined in
+   <xref target="RFC3986" x:fmt="of" x:sec="3.5"/>.
+   Only some URI-based HTTP protocol elements allow the inclusion of fragment
```
royfielding: Some protocol elements that refer to a URI allow inclusion of a fragment, while others do not.

royfielding: They are distinguished by use of the ABNF rule <URI-reference> for elements where fragment is allowed; otherwise, a specific rule that excludes fragments is used (e.g., target-resource).

PR #107, opened 2018-06-25
Re-Introduce BWS in chunk-ext (closes #101)

### LINKED ISSUE #101: RFC 7230: chunk-ext ABNF needs to allow more WS
See https://www.rfc-editor.org/errata/eid4667 and https://www.rfc-editor.org/errata/eid4825

reschke:


reschke:
That's what the verified erratum for https://www.rfc-editor.org/errata/eid4667 says.

If we want to do something else, we should discuss in in #101 or in the referenced mailing list thread.

royfielding:
We should probably commit this first and then fix it as a new issue later.

Issue #61, opened 2018-04-24
Deprecate Accept-Charset


annevk:
It's my understanding that the IETF recommends exclusive use of UTF-8. It seems therefore that `Accept-Charset` can go. (Browsers already do this and Fetch requires it, FWIW.)

reschke:
Please clarify what exact change you are proposing.

FWIW, I agree that exclusive use of UTF-8 is preferrable, but this is not a new protocol. We can not "un-define" this header field. We *could* add prose explaining why it's still there, and why it may be bad to base new HTTP applications on it. Thinking of it, this might be input for BCP56bis?

annevk:
You can recommend user agents to not send it and servers not to rely on it existing.

reschke:
The latter part is already covered by the spec, no?

WRT former, I certainly agree for browsers, but again, this is a 25 year old protocol, and there may be legitimate uses in the wild.

mnot:
We can deprecate it, we've done that elsewhere.

reschke:
It's fully optional. I don't see any merit in deprecation. We can explain why it's less important than 25 years ago, though.

mnot:
It's also fingerprinting surface area. That's a security/privacy concern, which changes things a bit.

reschke:
Well, the header field can't "go". It has a defined meaning and it is in use, and that should be documented.

We can advice against it's use, if that's what you mean by "deprecating".

reschke:
FWIW, we should document fingerprint surface in general for all request header fields (the recent discussion about webp comes to mind).

mnot:
I don't think that was proposed. I think what's on the table is:

* Putting some prose in explaining the downsides, and
* Changing its status in [the registry](https://www.iana.org/assignments/message-headers/message-headers.xhtml#perm-headers) to "obsoleted."

mnot:
+1 to that. I've just added some text to BCP56bis asking applications to consider that, but specific guidance for headers would be great.

mnot:
+1 to that. I've just added some text to BCP56bis asking applications to consider that, but specific guidance for headers would be great.

reschke:
> The User-Agent header field might contain enough information to uniquely identify a specific device, usually when combined with other characteristics, particularly if the user agent sends excessive details about the user's system or extensions. However, the source of unique information that is least expected by users is proactive negotiation (Section 5.3), including the Accept, Accept-Charset, Accept-Encoding, and Accept-Language header fields.

[...]

PR #113, opened 2018-06-29
Clarify that the charset parameter value is case-insensitive due to the definition in RFC 2046 (closes #13)

### LINKED ISSUE #13: RFC 7231 content-type charset param case sensitivity
https://greenbytes.de/tech/webdav/rfc7231.html#media.type

> The type, subtype, and parameter name tokens are case-insensitive. Parameter values might or might not be case-sensitive, depending on the semantics of the parameter name. The presence or absence of a parameter might be significant to the processing of a media-type, depending on its definition within the media type registry.

...it then continues showing examples that imply that the charset value is case-insensitive (which is correct), but it would be helpful if it cited RFC 2046 which defines charset's case  sensitivity.

reschke:


royfielding:
It's an example. The fact that charset names are case-insensitive is both well-known and defined (correctly) one inch below that.  Adding an out-dated reference does not help the example.  If this is somehow a problem (for whatever reason), just delete that one line of the example.

reschke:
@royfielding  - not convinced. The section you are referring to is about the "charset" ABNF production, not the parameter called "charset" in media types. These are not the same; the former is token-shaped and only used in "Accept-Charset", the latter is a parameter value.

Maybe there's more to clarify here. Maybe <https://greenbytes.de/tech/webdav/draft-ietf-httpbis-semantics-02.html#charset> should be eliminated where it is right now - it's only used by Accept-Charset (see also #61)

reschke:
More feedback appreciated.

I would recommend that we apply this pull request and also consider removing the "charset" ABNF production, given the fact that it is only used in one header field definition.

royfielding:
This is confusing too many things.

First, we have a set of four examples of things that are equivalent.  If you want to remove one of them just because we haven't yet defined that in the spec (two inches down), that's fine.  Just delete that line. It is not worth the added explanation.

Second, the section on charset defines the charset property as a relevant semantic for media types and normatively refers to the correct spec: RFC6365. The reference to how to register charset values is informative. This is a subsection of media types. We could change the order here and define charset as a parameter before a subsection on equivalent examples (if we want to keep all four examples).

Third, the rule "charset" is defined here for historical reasons because that is where I think a reader would look for it. If you want to move the ABNF fragment into Accept-Charset instead, that's fine with me as well.  But they are not two different things. The charset parameter and the charset token are both defined as case-insensitive because of the semantics here.

I am really not good at making interactive changes via pull requests, so it may be easier to commit a first draft and iterate within git instead of via PRs.

reschke:
Actually, RFC 6365 is "Terminology Used in Internationalization in the IETF", and we are referring to it to define the term "character encoding scheme".

royfielding:
and it is also the normative definition of "[charset](https://tools.ietf.org/html/rfc6365#page-8)" (the older RFC is only the registration mechanism).

royfielding:
I think I see what you mean, though. For the ABNF definition of charset, the normative definition is in RFC2978 as

>  mime-charset = 1*mime-charset-chars
>   mime-charset-chars = ALPHA / DIGIT / "!" / "#" / "$" / "%" / "&" / "'" / "+" / "-" / "^" / "_" /  "`" / "{" / "}" / "~"
>  ALPHA        = "A".."Z"    ; Case insensitive ASCII Letter
>  DIGIT        = "0".."9"    ; Numeric digit

PR #125, opened 2018-07-19
Add status code  422 (closes #123)

### LINKED ISSUE #123: Include status 422?
It's commonly used, but defined in RFC 4918 (WebDAV),, which continues to confuse people.

Should we include it in the core specs?

reschke:
The text is almost the same as in RFC 4918, except that I removed the somehow bogus claim that status code 400 wouldn't be right.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -6288,6 +6288,11 @@ Expect: 100-continue
    <c>
       <xref target="status.417"/>
    </c>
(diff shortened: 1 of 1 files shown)


reschke:
> Are we updating the IANA registry to point to this document?

Yes. For all codes.

> The only slight concern I have here is that this may unintentionally reinforce the view that "real" status codes are required to be in the core document. If we can find a way to discourage that view, that would be great (but that's a separable issue).

That occurred to me as well.

PR #154, opened 2018-10-15
Explain the limited utility of *

### LINKED ISSUE #46: * in Accept-*
`*` in [`Accept`](https://httpwg.org/specs/rfc7231.html#header.accept), [`Accept-Charset`](https://httpwg.org/specs/rfc7231.html#header.accept-charset) and [`Accept-Encoding`](https://httpwg.org/specs/rfc7231.html#header.accept-encoding) request headers is of very limited use, because:

* With a qvalue higher than another value, it says "send anything in preference to something specific", which isn't useful
* With a qvalue lower than another media type, it says "send a more specific value in preference to something else", which is already implied by HTTP
* With a qvalue of 0, it says "don't ever send other value" -- but a server can always ignore that, because this is an optional-to-implement extension
* In media types (e.g., `text/*`), it has been cargo culted a lot, but doesn't actually give the server information it can use. 

Can we deprecate `*` in these headers, or at least restrict its use / give better guidance?

mnot:
Fixes #46.


mnot:
Yeah, I wasn't happy about that either, particularly. However, if we consolidated it (e.g., into 6.4.1 or 8.4.1), I seriously doubt it will be read, since it's disconnected from the text that people will go to as a reference.

I'll take another try, creating a more obvious reference to the generic section. I'll also edit some of the examples that lead people down the path we're trying to avoid.

mnot:
Yeah, the problem is that we devote a fair amount of text in each of those headers to the `*` mechanism, so it needs to go next to each of those bits. Take a look in context.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5251,11 +5251,11 @@ Expect: 100-continue
    The example
 </t>
 <figure><artwork type="example">
-  Accept: audio/*; q=0.2, audio/basic
+  Accept: audio/mp3; q=0.2, audio/basic
```
royfielding: I don't understand why this is being changed.

royfielding:
I am a little lost with regard to this change.  Yes, it makes sense to note *somewhere* that wildcard accept parameters are not very useful in general browsing considerations. However, saying that doesn't change the protocol, nor does it lessen the need to explain what things like "audio/*" mean within examples. There are still many utilities, servers, and APIs that do make use of those features.

draft-ietf-httpbis-semantics-latest.xml
mnot: What's the use case for asking for `audio/*` vs. omitting the header field?

mnot:
The change doesn't disallow `*/*` and similar; it just doesn't hold them out as exemplars (since many people -- even implementers -- glance at examples to understand a spec, for better or worse).

draft-ietf-httpbis-semantics-latest.xml
martinthomson: The pattern I've seen is: `Accept: image/webp, */*; q=0.5`.  Though why `*/*` rather than `image/*` is not clear.  The point being that this is going into something that can handle a range of image formats - presumably too many to list in the request - but there is still a desire to express a preference for one or two particular formats.

s/image/audio/g as needed.

mnot: What information does that header contain that isn't conveyed by `Accept: image/webp`? The server is already explicitly allowed to ignore the header as it sees fit.

mnot:
@royfielding take a look at #189; it's a bit rough.

royfielding:
Yes, let's go with #189

draft-ietf-httpbis-semantics-latest.xml
royfielding: The original motivator was external viewers like xv that were configured in the browser to handle image/* as an extension (the browser itself had no idea what image formats were actually acceptable -- it was just instructed to accept all of them). This can also show up in audio-only environments (screen readers) where an audio alternative is preferred if available.

Also, originally, the server was not allowed to ignore; it was just that only a few implemented negotiation, and even then only for specific resources. That spec text is just trying to capture reality that the user agent has no control over which resources are actually negotiable, and when that negotiation happens, which makes it appear like the server can ignore. We tried (and failed) to address that with 406.

royfielding: Did I mention that I hate content negotiation?

PR #173, opened 2018-11-13
mention cache-control: immutable

### LINKED ISSUE #69: CC: immutable
... might be worth mentioning in Conditional Requests.

mnot:
Fixes #69


mnot:
How about replacing the current proposed text with:

> Note that protocol extensions can modify the conditions under which revalidation is triggered. Fo recompile, the "immutable" cache directive (defined by [RFC8246]) instructs clients to forgo revalidation of fresh responses in some circumstances.

MikeBishop:
The first part seems reasonable.  I have a little trouble with "Fo recompile," and I'm not sure what "some circumstances" is alluding to -- the circumstance I can think of where you wouldn't is that the response is no longer fresh, but you've already specified freshness.

mnot:
Bah, autocorrect (and perhaps, being sick with resolving issues). Trying again:

> Note that protocol extensions can modify the conditions under which revalidation is triggered. For example, the "immutable" cache directive (defined by [RFC8246]) instructs caches to forgo revalidation of fresh responses even when requested by the client.

mnot:
@royfielding that work for you?

Issue #42, opened 2017-11-23
Header Field Registry


mnot:
... is currently interleaved with News and SMTP, which is less than ideal.

See:
  https://mnot.github.io/I-D/httpbis-header-registry/

royfielding:
Strictly speaking, it's an IANA instruction, but those instructions are in some other RFC.  I think we would need the responsible list to agree first, which I guess means dispatch.

mnot:
Yes. I'm talking to Alexey about it.

mnot:
Discussed in DISPATCH Bangkok, no problem from them.

I've updated and submitted my draft, see:
  https://tools.ietf.org/html/draft-nottingham-httpbis-header-registry-00

mcmanus:
in bkk: make it so

mnot:
As per discussion in #188, re-opening to remind us to remove this:

> Entries with a status of 'deprecated' are to have a status of 'obsoleted'.

PR #177, opened 2018-11-29
Establish a separate header registry

mnot:
For #42

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -11045,6 +11045,7 @@ Content-Encoding: gzip
 <ul>
   <li>In <xref target="status.308"/>, include status code 308 from RFC 7538 (<eref target="https://github.com/httpwg/http-core/issues/3"/>)</li>
   <li>Rework <xref target="operation"/> to be more version-independent (<eref target="https://github.com/httpwg/http-core/issues/142"/>)</li>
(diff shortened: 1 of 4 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9545,10 +9590,40 @@ Content-Encoding: gzip
...
+   Message Header Registries with the protocol 'http' are to be moved to it,
+   with the following changes applied:
+</t>
+
+<ol>
+   <li><t>The 'Applicable Protocol' field is to be omitted.</t></li>
+   <li><t>Entries with a status of 'standard', 'experimental', or
+   'informational' are to have a status of 'permanent'.</t></li>
+   <li><t>Entries with a status of 'deprecated' are to have a status of
+   'obsoleted'.</t></li>
```
jyasskin: I'd expected to see a mention of 'obsoleted' in the field.name.registry section if it's going to be used. Should the "Provisional entries can be removed by the Expert(s)" permission instead on in addition say the Expert can mark entries 'obsoleted'?

mnot: Any registry can have obsoleted entries; it comes with the territory (and the registration section is targeted at people registering new ones). 

For provisional, the point is to reclaim the entries if they aren't used. The experts *can* mark them obsoleted if they like; IANA instructions are much less prescriptive than you might think.

reschke:
I believe the registry URI should actually say "http-header-fields". Can I change that?

PR #178, opened 2018-11-29
Caution against treating data on a connection as part of a not-yet-issued request

### LINKED ISSUE #26: Parsing content after a 204 response
When data is written to a connection following a 204 response, the user agent may interpret the data as it pleases. In web browsers today, this means:

- The Chromium and Edge web browsers will inspect the first 4 bytes for the
  start of a valid HTTP response. If found, they will parse the data that
  follows as a new response (and if any of those first four bytes bytes are
  invalid they will be discarded). If more than four invalid bytes are
  encountered, the browsers abort parsing and interpret the data as a
  HTTP/0.9 response. This is consistent with their general response parsing
  behavior (i.e. without a preceding a 204 response)
- The Firefox web browser may inspect 1 kilobyte of data or more (the exact
  number has been variable in my testing) for a valid response. If found, it
  will discard any preceding invalid data. This tolerant behavior is only
  observable following a 204 response; otherwise, Firefox seems to parse in the
  same way as Chromium and Edge.
- the Safari web browser, upon receiving any invalid data, makes no attempt to
  recover and discards the remaining data

This variation has led to instability in automated tests written for [the Web Platform Tests project](https://github.com/w3c/web-platform-tests)--see [issue 5037](https://github.com/w3c/web-platform-tests/pull/5037).

I originally reported this inconsistency in [issue 5227](https://github.com/w3c/web-platform-tests/pull/5227), where @mnot provided the following context (from [RFC7230 section 3.3.3](http://httpwg.org/specs/rfc7230.html#message.body.length)):

> If the final response to the last request on a connection has been completely
> received and there remains additional data to read, a user agent MAY discard
> the remaining data or attempt to determine if that data belongs as part of
> the prior response body, which might be the case if the prior message's
> Content-Length value is incorrect. A client MUST NOT process, cache, or
> forward such extra data as a separate response, since such behavior would be
> vulnerable to cache poisoning.

Mark followed up by saying:

> What I think's being requested is a recommendation for how much data should
> be discarded before the client gives up; possibly a minimum. It feels kind of
> analogous to when we established the minimum URL length that should be
> supported by implementations, so it's not completely off base.
>
> That said, this is truly a corner case; the right answer is "don't do that."
> Anyone depending on interop in this case is doing it wrong to start with.

Though I agree with @annevk: "I think ideally HTTP defines how to parse HTTP." Can the specification language be made more explicit for expected behavior in this situation?

Thanks for your consideration!

mnot:
Fixes #26.

### DIFF
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -1743,7 +1743,8 @@ https://www.example.org
    If a HTTP/1.1 client receives data on a connection that doesn't have any
    outstanding requests, it &MUST-NOT; consider them to be a response to a
    not-yet-issued request; it &SHOULD; close the connection, since message
-   delimitation is now ambiguous.
+   delimitation is now ambiguous, unless the data consists only of one or
+   more CRLF (which can be discarded, as per <xref target="message.parsing"/>).
 </t>
 </section>
 
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -1739,6 +1739,11 @@ https://www.example.org
    ordered request that has not yet received a final (non-<x:ref>1xx</x:ref>)
    response.
 </t>
+<t>
+   If a HTTP/1.1 client receives data on a connection that doesn't have any
+   outstanding requests, it &MUST-NOT; consider them to be a response to a
+   not-yet-issued request; it &SHOULD; silently discard them.
+</t>
 </section>
 
 <section title="Persistence" anchor="persistent.connections">
@@ -3348,6 +3353,7 @@ Upgrade: HTTP/2.0
 <ul>
    <li>In <xref target="status.line"/>, explain why the reason phrase should be ignored by clients (<eref target="https://github.com/httpwg/http-core/issues/60"/>).</li>
    <li>Add <xref target="associating.response.to.request"/> to explain how request/response correlation is performed (<eref target="https://github.com/httpwg/http-core/issues/145"/>)</li>
+   <li>In <xref target="associating.response.to.request"/>, caution against treating data on a connection as part of a not-yet-issued request (<eref target="https://github.com/httpwg/http-core/issues/26"/>)</li>
 </ul>
 </section>
 
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -1742,8 +1742,8 @@ https://www.example.org
 <t>
    If a HTTP/1.1 client receives data on a connection that doesn't have any
    outstanding requests, it &MUST-NOT; consider them to be a response to a
-   not-yet-issued request; it &SHOULD; silently discard them, and SHOULD
-   close the connection, since message delimitation is now ambiguous.
+   not-yet-issued request; it &SHOULD; close the connection, since message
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1739,6 +1739,11 @@ https://www.example.org
    ordered request that has not yet received a final (non-<x:ref>1xx</x:ref>)
    response.
 </t>
+<t>
+   If a HTTP/1.1 client receives data on a connection that doesn't have any
+   outstanding requests, it &MUST-NOT; consider them to be a response to a
+   not-yet-issued request; it &SHOULD; silently discard them.
```
MikeBishop: ...and SHOULD perhaps close the connection, since it might now be out-of-sync with the server about where the message boundaries are.

royfielding: Is it really necessary to say it SHOULD silently discard them? What it should do in this case is dependent on the nature of the client, not on the protocol.

Yes, the connection MUST be closed. OTOH, do implementations ever check to see if data has arrived before they send a first request? My guess is this would depend on what connection management and IO frameworks are being used to do connection management. Let's try to avoid making simple implementations impossible.

mcmanus: how does the should interact with 408? 408 seems like a pretty common pattern for this

mnot: I removed the "silent discard." Regarding 408s -- you might encounter one, but the server is going to be closing the connection anyway, right?

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1739,6 +1739,12 @@ https://www.example.org
    ordered request that has not yet received a final (non-<x:ref>1xx</x:ref>)
    response.
 </t>
+<t>
+   If a HTTP/1.1 client receives data on a connection that doesn't have any
```
MikeBishop:
```suggestion
   If an HTTP/1.1 client receives data on a connection that doesn't have any
```

wtarreau:
I also suggest closing here : the main reason for getting data without a request is an error reported by the server before closing (e.g. 408). Another reason may be a server implementation issue (extra CRLF, inconsistent multiple content-length, inconsistent content-length+transfer-encoding, user-fed incorrect error message). In all these cases the message delimitation is wrong and unrecoverable. Thus the client trying to "fix" it will in the best case be useless (close following), and in the worst case have nasty side effects.

In haproxy the only case we try to address is the extra CRLF that a few old servers send slightly after data. If extra bytes are sent immediately after data, we close. If the extra CRLF happens slightly later, we find it while parsing the next response and the CRLF is silently skipped.

mnot:
Everyone OK with this now?

wtarreau:
Well, like Roy, I'd even prefer a MUST close, but I can live with a SHOULD.

royfielding:
It doesn't seem to account for the extra CRLF case.  Maybe just add "... unless the data is limited to CRLF"?

mnot:
CR and CRLF?

wtarreau:
I was thinking the same but hesitating. BTW @mnot it would be LF and CRLF. But quite frankly we could stick to CRLF only since it's the only known bogus pattern widely deployed in field.

mnot:
updated.

PR #181, opened 2018-11-29
Start #165: list non-cached and non-updated header fields

### LINKED ISSUE #165: Updating stored headers
7234 requires stored headers to be updated upon a `304` or `HEAD` response. 

However, there are a number of issues:

* It doesn't specify that H1 connection-related headers should be omitted
* Some browsers omit other headers (e.g., `x-content-*`, `content-*` and `x-webkit-*` - see [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=174301) and [Webkit](https://bugs.webkit.org/show_bug.cgi?id=72414)
* Apache has a whitelist of headers; see [bug](https://bz.apache.org/bugzilla/show_bug.cgi?id=61820)
* Intermediaries are reluctant to do this for performance reasons; e.g., [squid](https://bugs.squid-cache.org/show_bug.cgi?id=7)

It seems like a lot of the bugs referenced above originated from the [confusing language around omitting entity headers in RFC2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.5.3); that was removed in RFC7234 (see [current language](https://httpwg.org/specs/rfc7234.html#freshening.responses)).

I think we need to:

* [ ] Specify the list of H1 connection-oriented headers to be omitted
* [ ] Talk about other headers that need to be omitted (coordinating especially with browsers, who seem to have weird behaviour here)
* [ ] Contemplate any advice/warnings/requirements we can give to an intermediary cache about how they should handle this if they don't want the performance hit

[Mozilla's list of headers they don't update](https://dxr.mozilla.org/mozilla-central/source/netwerk/protocol/http/nsHttpResponseHead.cpp#879) might be a good starting point for the first two (see also [related bug](https://bz.apache.org/bugzilla/show_bug.cgi?id=61820)).

There are also [browser tests](https://wpt.fyi/results/fetch/http-cache/304-update.html?aligned&label=stable) and an [outstanding PR](https://github.com/web-platform-tests/wpt/pull/12316).

Note that this is a security issue, because some browsers filter out headers starting with `Content-`, which includes `Content-Security-Policy` -- i.e., a browser that has an old copy of the response won't see the new CSP header on a 304 response.

(this might also apply to 206 responses, since there's header combination there too)

jyasskin:
Debate continues about exactly which headers should be included in these
lists, so my goal in this change is just to record the status quo
without expressing any opinion that it's a good status quo.

@MattMenke2, Chrome's failing the "update Content-Test-Header" test in https://wpt.fyi/results/fetch/http-cache/304-update.html?aligned&label=stable, but I don't see code in https://cs.chromium.org/chromium/src/net/http/http_response_headers.cc to do that. Any idea what's going on?

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -384,6 +384,39 @@
    prohibited from storing such responses.
 </t>
 
+<section title="Storing Response Header Fields" anchor="storing.headers">
+<t>
+   When storing a response, a cache &MUST; omit header fields in the following
+   categories:
+</t>
+<ul>
+   <li>Hop-by-hop header fields listed in the Connection header field (<xref
+      target="header.connection" />).</li>
+   <li>Header fields listed in the no-cache response directive in the Cache-Control
+      header field (<xref target="cache-response-directive.no-cache" />).</li>
+   <li>Header fields whose name appears in the following list:
+      <ul>
+         <li>Clear-Site-Data</li>
+         <li>Connection</li>
+         <li>Keep-Alive</li>
+         <li>Proxy-Authenticate</li>
+         <li>Proxy-Connection</li>
+         <li>Public-Key-Pins</li>
+         <li>Set-Cookie</li>
+         <li>Set-Cookie2</li>
+         <li>Strict-Transport-Security</li>
+         <li>Trailer</li>
+         <li>Transfer-Encoding</li>
+         <li>Upgrade</li>
+         <li>WWW-Authenticate</li>
+      </ul>
+   </li>
+</ul>
+<t>
+   TODO: The above list is not final.
+</t>
+</section>
+
 <section title="Storing Incomplete Responses" anchor="incomplete.responses">
 <t>
    A response message is considered complete when all of the octets indicated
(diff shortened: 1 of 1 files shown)


MattMenke2:
I'll look into that, though I may not have time until next week.

jyasskin:
This should probably also update the header registry to declare how new headers are cached, once #177 is merged.

My guess is to add a field whose value is "uncached", "may-cache", "may-update", or "may-cache and may-update", but that'll depend on exactly the shape of the lists that get approved here.

MattMenke2:
It looks like the test passes on the latest Chrome Canary channel, though it is definitely failing on Chrome stable.  I don't see any failure expectation in Chrome's git history, so I guess it was fixed in Chrome before the wpt test landed, or I'm just really bad at looking through git history, which is certainly a very real possibility.

Either way, I don't think it's worth digging into that failure more, since it will presumably start passing when Chrome 72 makes it to stable, and possibly even Chrome 71.

MattMenke2:
Sorry - actually I reviewed the CL to fix that test.  It was in Chrome 72.  The test was inspired by Chrome's need to fix handling of that exact case.

PR #186, opened 2019-01-08
Reference MIME sniffing and X-CTO

### LINKED ISSUE #51: MIME sniffing
7231 [says](http://httpwg.org/specs/rfc7231.html#header.content-type):

> In practice, resource owners do not always properly configure their origin server to provide the correct Content-Type for a given representation, with the result that some clients will examine a payload's content and override the specified type. Clients that do so risk drawing incorrect conclusions, which might expose additional security risks (e.g., "privilege escalation"). Furthermore, it is impossible to determine the sender's intent by examining the data format: many data formats match multiple media types that differ only in processing semantics. Implementers are encouraged to provide a means of disabling such "content sniffing" when it is used.

It seems like this should at least mention:
* [MIME sniffing](https://mimesniff.spec.whatwg.org)
* [x-content-type-options: nosniff](https://fetch.spec.whatwg.org/#x-content-type-options-header)

mnot:
Fixes #51.


reschke:
(see review comments)

mnot:
@reschke I don't see any...

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -10018,6 +10021,24 @@ Content-Encoding: gzip
   <seriesInfo name="ACM Transactions on Internet Technology" value="1(2)"/>
 </reference>
 
+<reference anchor="mimesniff" target="https://mimesniff.spec.whatwg.org">
+   <front>
+      <title>MIME Sniffing</title>
+      <author>
+         <organization>WHAT Working Group</organization>
```
annevk: Can this be WHATWG? Or do you also write IE Task Force?

royfielding: updated to WHATWG

PR #187, opened 2019-01-08
Move parameter into its own section.

### LINKED ISSUE #45: Header normalization rules
From httpwg/http-extensions#282 (see that for more context).

Are the following header fields considered equal or equivalent or are they different?

```
Header: value; p=pval
Header: value; p="pval"
Header: value; p="Pval"
Header: value; P="Pval"
```
Some specializations of the header syntax permit both quoted-string and token for parameter values. If the string value after removal of quotes is the same as the token, is it the same value?

What about case folding for parameter names? Do we do that?

mnot:
Fixes #45.

This places it as a subsection of 1.2 Syntax Notation; the intent is to also move 11. ABNF List Extension: #rule and 4.3. Whitespace there, along with any other truly generic syntax.


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -309,6 +309,42 @@
...
+      transmitted either as a token or within a quoted-string. The quoted and
+      unquoted values are equivalent. For example, the following examples are
+      all equivalent, but the first is preferred for consistency:
+   </t>
+   <figure><artwork type="example">
+     charset=utf-8
+     charset=UTF-8
+     Charset="utf-8"
+     charset="utf-8"
+   </artwork></figure>
```
reschke: Why are the first two ones equivalent?

mnot: Good point.

mnot: Fixed, thx.

mnot:
Do you mean losing the example of media type value case insensitivity, or something else?

reschke:
> Do you mean losing the example of media type value case insensitivity, or something else?

Yes.

mnot:
So, the spec is still there - you feel we need an explicit example of that?

reschke:
Yes, as it's an FAQ. I believe we had a separate ticket about this; let me have a look.

mnot:
@reschke I think your concern is addressed by the latest commit.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2159,53 +2205,34 @@ Host: www.example.org
...
+   For example, the following media types are equivalent in describing HTML
+   text data encoded in the UTF-8 character encoding scheme, but the first is
+   preferred for consistency (the "charset" parameter value is defined as
+   being case-insensitive in <xref target="RFC2046" x:fmt="," x:sec="4.1.2"/>):
 </t>
 <figure><artwork type="example">
   text/html;charset=utf-8
   Text/HTML;Charset="utf-8"
   text/html; charset="utf-8"
-</artwork></figure>
```
reschke: Why was this removed?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2225,12 +2252,12 @@ Host: www.example.org
 </t>
 <aside>
   <t>
-    &Note; in practice, charset names are furthermore restricted by the
```
reschke: I would prefer this to be a separate PR.

mnot:
LGTM.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1650,14 +1697,13 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
 <t>
-   Many header fields use a format including (case-insensitively) named
-   parameters (for instance, <x:ref>Content-Type</x:ref>, defined in
-   <xref target="header.content-type"/>). Allowing both unquoted (token) and quoted
-   (quoted-string) syntax for the parameter value enables recipients to use
-   existing parser components. When allowing both forms, the meaning of a
-   parameter value ought to be independent of the syntax used for it (for an
-   example, see the notes on parameter handling for media types in
-   <xref target="media.type"/>).
+   Many header fields (including <x:ref>Content-Type</x:ref>, defined in
```
reschke: s/including/for instance/?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1572,6 +1582,9 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
                  / <x:ref>DIGIT</x:ref> / <x:ref>ALPHA</x:ref>
                  ; any <x:ref>VCHAR</x:ref>, except delimiters
 </sourcecode>
+</section>
+
+<section title="Quoted Strings and Comments">
```
reschke: I think it would be better to have separate subsections for strings and comments - having them in the same implies some relation, no?

reschke:
I think we're almost there. @mnot if you're ok with my proposed changes I can update the PR.

PR #188, opened 2019-01-08
Obsolete accept-charset

### LINKED ISSUE #61: Deprecate Accept-Charset

mnot:
Fixes #61

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ $(bd)/%.redxml: %.xml $(reduction)
 	$(xml2rfc) $< -o $@
 
 $(bd)/%.abnf: %.xml lib/extract-artwork.xslt
-	$(saxon) $< lib/xreffer.xslt | $(saxon) - lib/extract-artwork.xslt type="abnf2616" >$@
+	$(saxon) $< lib/xreffer.xslt | $(saxon) - lib/extract-artwork.xslt type="abnf7230" >$@
 
 $(bd)/%.parsed-abnf: $(bd)/%.abnf
 	$(bap)/bap -i $(bap)/core.abnf -X rfc7405 < $< | LC_COLLATE=C sort | $(bap)/bap -k -i $(bap)/core.abnf -X rfc7405 -l 69 >$@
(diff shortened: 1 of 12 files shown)


mnot:
If the preceding text isn't adequate, can you suggest something?

wtarreau:
Don't you fear that with this "fingerprinting" thing we open a huge can of worms if presented as the main concern? Accept-language is even worse in this regard and is important. Probably we should primarily focus on the uselessness and the waste of bandwidth that results from sending a header field with a well-knownn value ? What about something like this : 

   The Accept-Charset header field is not sent by most agents nowadays since its absence
   indicates that all well-known charsets are supported, which does not deserve wasting
   request bandwidth to explicitly state a known default. In addition clients are encouraged
   not to send a list of their supported charsets, as it makes it easier for servers to
   fingerprint them.

reschke:
Hm, all well-known charsets are supported? Isn't it more relevant that *UTF-8* is universally supported, thus there's really no need for anything else?

annevk:
Yeah, I'd say something like "Now UTF-8 has ubiquitous support and wide adoption, there's no need for Accept-Charset to use up bandwidth and in fact most user agents no longer transmit it."

wtarreau:
LGTM

wtarreau:
@reschke maybe "deprecated" would be more appropriate ? It still retains some value in some specific places but better avoid it for new stuff.

mnot:
I'm not crazy about defining fine distinctions between "obsoleted" and "deprecated."

mnot:
If the defining document remains at 7231, it *will* be obsoleted.

But yeah, I'm not going to die on this hill. Updating PR shortly.

mnot:
Actually, re-jigging the HTTP header registry in #42 resulted in this in [IANA considerations](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#header.field.registration):

> Entries with a status of 'deprecated' are to have a status of 'obsoleted'.

I take it that folks want that backed out, to preserve that state?

wtarreau:
The change looks good to me.

Regarding this : 
> Entries with a status of 'deprecated' are to have a status of 'obsoleted'.
Well, there is no reason for the deprecated to last only for a short duration or a single iteration of a spec. I see "deprecated" as "not recommended for new designs". The explanation in the change proposal makes it sufficiently clear so that one can decide whether to use it or not in a specific context.

reschke:
Yes, I believe keeping the distinction between deprecated and obsoleted would be good.

PR #189, opened 2019-01-08
Rearrange conneg and talk about *

### LINKED ISSUE #46: * in Accept-*

mnot:
Fixes #46.

(alternative approach)

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -5182,7 +5182,7 @@ Expect: 100-continue
 <t>
    &Note; In practice, using wildcards in content negotiation has limited
    practical value, because it is seldom useful to say, for example, "I
-   prefer image/* more or less than (some other specific value). Clients can
+   prefer image/* more or less than (some other specific value)". Clients can
    explicitly request a <x:ref>406 (Not Acceptable)</x:ref> response if a
    more preferred format is not available by sending Accept: */*;q=0, but
    they still need to be able to handle a different response, since the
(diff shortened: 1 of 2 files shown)


reschke:
On 2019-01-08 07:02, Mark Nottingham wrote:
> So, leave #rule as is, and add a new "Generic Syntax" section?

Yes.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5157,6 +5157,38 @@ Expect: 100-continue
...
+<t>
+   Each of these header fields defines a wildcard value (often, "*") to
+   select unspecified values. If no wildcard is present, all values not
+   explicitly mentioned in the field are considered "not acceptable" to the
+   client.
+</t>
+<t>
+   &Note; In practice, using wildcards in content negotiation has limited
+   practical value, because it is seldom useful to say, for example, "I
+   prefer image/* more or less than (some other specific value). Clients can
```
martinthomson:
```suggestion
   prefer image/* more or less than (some other specific value)". Clients can
```

PR #200, opened 2019-02-26
Obsolete Content-MD5

### LINKED ISSUE #93: Content-MD5
The deprecation of Content-MD5 in 7231 was a bit too subtle; it's only mentioned in a change note, and the registry wasn't updated.

mnot:
Fixes #93

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -9584,8 +9584,13 @@ Content-Encoding: gzip
    appropriate link.
 </t>
 <t>
-   After that is complete, please update the registry with the
+   After that is complete, please update the new registry with the
    header field names listed in the table of <xref target="field.names"/>.
+</t>
+<t>
+   Finally, please update the "Content-MD5" entry in the new registry to have
+   a status of 'obsoleted', and a reference of RFC7231, Appendix B, as this
+   instruction was omitted from that specification.
 </t>
 </section>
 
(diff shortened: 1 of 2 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9584,9 +9584,14 @@ Content-Encoding: gzip
    appropriate link.
 </t>
 <t>
-   After that is complete, please update the registry with the
+   After that is complete, please update the new registry with the
```
reschke: unrelated change..., would prefer it to appear in a separate commit :-)

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9584,9 +9584,14 @@ Content-Encoding: gzip
...
    header field names listed in the table of <xref target="field.names"/>.
 </t>
+<t>
+   Finally, please update the "Content-MD5" entry in the new registry to have
+   a status of 'obsoleted', and a reference of RFC7231, Appendix B, as this
+   instruction was omitted from that specification.
+</t>
 </section>
```
reschke: change

 RFC7231, Appendix B

to

\<xref target="RFC7231" x:fmt="of" x:sec="B"/>

reschke:
I think this is ready.

royfielding:
I think the registry is supposed to reference the obsolete spec where it was last defined, not the spec that mentions it briefly in a change note. In this case, it should be referring to RFC2616 Section 14.15.

reschke:
@royfielding - I think the IANA registry should point to both - pull request updated.

PR #201, opened 2019-02-26
DNS disclosure security considerations

### LINKED ISSUE #10: .onion names
Note the requirements in RFC6761 regarding .onion names in RFC7230bis Sections 2.7.1 and 2.7.2.

mnot:
Closes #10

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -9298,7 +9298,7 @@ Content-Encoding: gzip
    documented in the Special-Use Domain Names registry {{!RFC6761}} which
    might specify that some protocol other than DNS be used for resolution for
    names within a particular domain. If those special requirements are not
-   honoured diligently then security, privacy and interoperability problems
+   honoured diligently then security, privacy, and interoperability problems
    might well result.
 </t>
 <t>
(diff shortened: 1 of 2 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9280,6 +9280,42 @@ Content-Encoding: gzip
...
+   page, and possibly even unique to the user -- this can result in the
+   disclosure of a substantial amount of information. Possible mitigations
+   include supporting encrypted DNS protocols, such as {{?RFC8484}}. See also
+   {{?RFC7626}}.
+<t>
+   Furthermore, such resolution needs to adhere to any special requirements
+   documented in the Special-Use Domain Names registry {{!RFC6761}} which
+   might specify that some protocol other than DNS be used for resolution for
+   names within a particular domain. If those special requirements are not
+   honoured diligently, security, privacy and interoperability problems might
```
annevk: Writing "then security" would improve clarity a bit.

Also, I'd recommend using an Oxford comma, but maybe that's consistently not done for HTTP?

royfielding:
Wow, that's ugly. Do we really have to include an obscure example of domain resolution that nobody in our audience is going to implement?

royfielding:
I meant: do we really need to include that third paragraph? It is irrelevant to HTTP.  Actually, none of them are relevant to HTTP, but the first two are at least responsive to the issue, whereas the third assumes the reader can't read the RFCs being referenced.

If we do need to include it, then I sure hope the IESG is prepared for the entire HTTP security considerations section to be included in every other RFC the IETF publishes that uses or is configured via HTTP, because that's the kind of pain that deserves to be spread evenly.

mnot:
@royfielding you mean the fourth para? The third para is the point here.

mnot:
Assuming that's what you meant, I'm happy dropping the example para; it was just the suggestion we got from-list.

royfielding:
Yeah, just the example paragraph -- I apparently lost count across the diffs.

PR #205, opened 2019-03-26
Change reason-phrase to be optional but non-empty

reschke:
...instead of requrin…g it and allowing it to be empty. Also add a note about what the absence of reason-phrase means in practice (#197)


royfielding:
@reschke I changed the normative paragraph to target the requirement to servers and make it more readable. Go ahead and merge if you approve.

Issue #164, opened 2018-10-18
ABNF recipient list production #element does not allow null element prior to first element


reschke:
See <https://www.rfc-editor.org/errata/eid5257>.

Mailing list discussion starts here: <https://lists.w3.org/Archives/Public/ietf-http-wg/2018JanMar/0131.html>

reschke:
See also #14

reschke:
A simple fix for this would be to simplify the list expansion to just map `\#element`, and to put all other constraints in prose (the current approach doesn't work anyway due to the rules for empty list elements which are not counted).

reschke:
Proposal for recipient list rule mapping:

~~~
#element => *( OWS "," ) *( [ OWS element ] OWS "," )
~~~

More eyes requested :-) @royfielding @mnot

reschke:
The current proposal is incorrect as it does not allow a singleton element.

reschke:
Maybe:

~~~
#element => *( OWS "," ) [ element ] *( OWS "," [ OWS element ] ) *( OWS "," )
~~~

reschke:
PR updated

royfielding:
And maybe

    1#element => element *( OWS "," OWS [ element ] )

right?

annevk:
@royfielding that'd outlaw `Multiple-Element-Header: , somevalue`, which seems wrongish?

royfielding:
Closed by commit e475f36 for the recipient case. Note that this may become obsolete if we switch to only providing ABNF for what is good to send.

PR #209, opened 2019-03-26
simplify list rule mapping for consumers not to address cardinality

reschke:
(#164)

work in progress, feedback appreciated.

1. is the actual rule mapping correct?
2. prose
3. in  the convered ABNF, should I add the cardinality as comment?


reschke:
re 3) turns out to be hard as long as we have productions that mix list with something else (see separate issue), because we can't inline comments *inside* and ABNF production

reschke:
@mnot - your turn to review...

PR #216, opened 2019-04-03
Add link to cache tests

annevk:


mnot:
IETF doesn't "do" official conformance tests, so we'll have to be very careful about how this is done. Will chat with some folks.

reschke:
I note that https://github.com/httpwg/wiki/wiki/Structured-Headers links to tests...

mnot:
I don't think we should link to these; they aren't official, and are still in a lot of flux. Maybe from the WG home page, if we can get agreement to do that (I'd like to iterate on them more before bringing it up though).

Issue #212, opened 2019-03-27
byte-range-set definition allows OWS but probably doesn't work in practice


royfielding:
In Semantics, the definition of byte-range-set is

    byte-range-set  = 1#( byte-range-spec / suffix-byte-range-spec )

implying that they are a list, like header field values, with optional whitespace.
However, all of the examples are things like

        bytes=0-0,-1
        bytes=500-600,601-999
        bytes=500-700,601-999

and I seriously doubt that recipients are prepared for OWS around the commas.

Related to #196 and #210

royfielding:
Hmm, it does seem to work in practice, at least for spaces between commas.

mnot:
Discussed in Montreal; allow it, and change example to illustrate.

royfielding:
An example has been added in 7594239b46ef2c755b2e91a4572d86707d5f1021

Issue #210, opened 2019-03-27
unnecessary use of 1#element for header field definitions


royfielding:
When working on #164 I noticed that our existing use of 1#element (header fields requiring at least one list item) is kind of pointless.  The following fields use it:

     Cache-Control   = 1#cache-directive
     Transfer-Encoding = 1#transfer-coding
     Connection        = 1#connection-option
     Upgrade          = 1#protocol
     Trailer = 1#field-name
     Via = 1#( received-protocol RWS received-by [ RWS comment ] )
     Content-Encoding = 1#content-coding
     Content-Language = 1#language-tag
     If-Match = "*" / 1#entity-tag
     If-None-Match = "*" / 1#entity-tag
     Accept-Charset = 1#( ( charset / "*" ) [ weight ] )
     Accept-Language = 1#( language-range [ weight ] )
     Vary = "*" / 1#field-name
     WWW-Authenticate = 1#challenge
     Proxy-Authenticate = 1#challenge

It is also used (incorrectly?) within Byte Ranges fields as

     byte-range-set  = 1#( byte-range-spec / suffix-byte-range-spec )
     acceptable-ranges = 1#range-unit / "none"

We should audit these because it makes more sense to handle empty fields as non-error conditions.

mnot:
If you mean that we should change them to `#rule` rather than `1#rule`, I tend to agree -- but IIRC there was a reason this was done. @reschke?

reschke:
See #164. This change would reduce the complexity of the list rule. We'd just have to put the "minimally one instance" constraint into prose.

mnot:
I think Roy's point is that we don't need to limit to a minimum number of elements -- even in prose.

reschke:
Well, for at least some of these, we may have to say what it means if there is none.

mnot:
From Roy's list, I've ticked the ones that I think that's true for:

- [ ] Cache-Control   = 1#cache-directive
- [ ] Transfer-Encoding = 1#transfer-coding
- [ ] Connection        = 1#connection-option
- [ ] Upgrade          = 1#protocol
- [ ] Trailer = 1#field-name
- [ ] Via = 1#( received-protocol RWS received-by [ RWS comment ] )
- [ ] Content-Encoding = 1#content-coding
- [ ] Content-Language = 1#language-tag
- [x] If-Match = "*" / 1#entity-tag
- [x] If-None-Match = "*" / 1#entity-tag
[...]

Issue #196, opened 2019-02-01
Range header field ABNF


reschke:
The Range header field ABNF:
~~~
  Range = byte-ranges-specifier / other-ranges-specifier
  other-ranges-specifier = other-range-unit "=" other-range-set
  other-range-set = 1*VCHAR
~~~
This is another misguided attempt to restrict syntax using ABNF in presence of extension points.

For instance,
~~~
bytes=abc
~~~

matches `other-ranges-specifier`.

The proper fix for this is to rewrite the ABNF so it's properly layered - a generic one, plus an overlay for the only range unit we define.

royfielding:
I am very far down the rabbit hole on this one.

royfielding:
Fixed by 7594239b46ef2c755b2e91a4572d86707d5f1021

Issue #11, opened 2015-12-22
Ranges and Content and Transfer Encoding


martinthomson:
RFC 7233 does not mention content encoding at all.  Same for transfer encoding.  I assume that is because this is completely unspecified and therefore completely unreliable, however, for my sanity...

My reading is that a 206 response includes ranges of the encoded message, and that the content-encoding applies to the complete message body prior to being split into ranges.  Thus, if I had a "x2" content encoding that turned "Hello World!" into "HHeelllloo  WWoorrlldd!!", asking for bytes 3-5 would get you "eel" and not "llo".

The text in Section 4.1 suggests that you would not include a Content-Encoding header field if the client used If-Range on the expectation that they already know.  That seems pretty dangerous, but it's consistent with the idea that you are repairing a larger message.

On the other hand, I have to assume that a Transfer-Encoding applies _after_ the range request.

enygren:
This is also potentially weird in the case where a client specifies multiple Accept-Encodings and a Range request, such as "Accept-Encoding: gzip, br".  In these cases, the bytes returned will depend on which Content-Encoding is selected by the server.  (As minimal operational guidance, clients presumably will need to be very careful about this and may need to limit which encodings they accept for subsequent ranges requests based on what was returned from the first, being willing to fall back to downloading the whole object on failure.)

mnot:
[RFC7233 Section 3.1](http://httpwg.org/specs/rfc7233.html#header.range) says:

> The "Range" header field on a GET request modifies the method semantics to request transfer of only one or more subranges of the selected representation data, rather than the entire selected representation data.

[RFC7231 Section 3.1.2.2](http://httpwg.org/specs/rfc7231.html#header.content-encoding) says:

> Unlike Transfer-Encoding (Section 3.3.1 of [RFC7230]), the codings listed in Content-Encoding are a characteristic of the representation; the representation is defined in terms of the coded form, and all other metadata about the representation is about the coded form unless otherwise noted in the metadata definition. 

That gives you all the information you should need. Could it be clearer? Probably. A hyperlink to defining terms would probably help.

reschke:
I believe this issue can be closed.

PR #217, opened 2019-04-04
Refactor the range specifier grammar

royfielding:
This refactors range-spec to simplify and reduce distinction with range unit extensions. Moving the extension grammar into range specifiers (and removing comma) reduces the overlapping rules and allows the stuff specific to byte ranges to be specified separately.

Related to #11, #196, #210, #212


reschke:
This refactors things further than I expected, in that it also introduces consistency for range sets across range units. I'm not totally sure we can go that for (will think about it).

reschke:
FWIW, <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range> appears to support Roy's proposal.

royfielding:
The consistency for range sets is saying that if you have a comma-separated list of range specifiers then it means you want more than one range. It's hard to imagine any scenario in which some future range unit would want

    Range: unit=set,set,set

to have any other meaning. And that is clearly the intent of the original ABNF.

I note that the MDN articles use a space after each comma. Perhaps we need to test that?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2628,49 +2634,103 @@ Host: www.example.org
...
-  <x:ref>byte-ranges-specifier</x:ref> = <x:ref>bytes-unit</x:ref> "=" <x:ref>byte-range-set</x:ref>
-  <x:ref>byte-range-set</x:ref>  = 1#( <x:ref>byte-range-spec</x:ref> / <x:ref>suffix-byte-range-spec</x:ref> )
-  <x:ref>byte-range-spec</x:ref> = <x:ref>first-byte-pos</x:ref> "-" [ <x:ref>last-byte-pos</x:ref> ]
-  <x:ref>first-byte-pos</x:ref>  = 1*<x:ref>DIGIT</x:ref>
-  <x:ref>last-byte-pos</x:ref>   = 1*<x:ref>DIGIT</x:ref>
+<sourcecode type="abnf7230"><iref primary="true" item="Grammar" subitem="ranges-specifier"/><iref primary="true" item="Grammar" subitem="range-set"/><iref primary="true" item="Grammar" subitem="range-spec"/>
+  <x:ref>ranges-specifier</x:ref> = <x:ref>range-unit</x:ref> "=" <x:ref>range-set</x:ref>
+  <x:ref>range-set</x:ref>        = <x:ref>range-spec</x:ref> *( "," <x:ref>range-spec</x:ref>)
```
royfielding: Perhaps it is better to leave this as

   <x:ref>range-set</x:ref>        = 1#<x:ref>range-spec</x:ref>

until we have decided #212

reschke:
Looks good in general.

We can decouple this from #212 by going back to the list syntax for now (as proposed).

I still believe we have a (good) normative change here, in that we now say how ranges can be interpreted for units other than "bytes". This needs to be documented.

PR #219, opened 2019-04-12
Clarify when trailers can be generated

### LINKED ISSUE #16: Trailers
7230 says:

> The trailer fields are identical to header fields, except they are sent in a chunked trailer instead of the message's header section.

and:

> When a chunked message containing a non-empty trailer is received, the recipient MAY process the fields (aside from those forbidden above) as if they were appended to the message's header section.

As discussed at the 2016 HTTP Workshop as part of @annevk's Fetch presentation, this isn't necessarily sensible.

mnot:
Fixes #16.


mnot:
See #220.

mnot:
1. Yes; see #220 
2. Only if there's something that can appear in a trailer that isn't currently marked as such.

reschke:
> Only if there's something that can appear in a trailer that isn't currently marked as such.

Just checking: with the current PR, we end up with no fields being allowed as trailers, right?

mnot:
in core specs, yes; if we want to allow some, we could open issues (but personally I'd want to see implementations honouring them, or at least planning to).

We probably need to reconcile this with [this comment](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.comment.3).

reschke:
Would be good to identify at least one and insert the required language. ETag comes to mind.

mnot:
Alas, no one does anything with `ETag` in a trailer AFAIK.

`Server-Timing` is the only example I'm aware of that's actually specified.

reschke:
Well. Chicken and egg. Previously it didn't *need* to be specified.

Of the header fields defined in the core specs, "Etag" is the first (and maybe only) that comes to mind to allow in trailers. Thus I'd argue that we should do this now.

mnot:
I think we should discuss it with the WG to make sure folks are comfortable with that.

> On 8 Jul 2019, at 8:31 pm, Julian Reschke <notifications@github.com> wrote:
> 
> Well. Chicken and egg. Previously it didn't need to be specified.
> 
> Of the header fields defined in the core specs, "Etag" is the first (and maybe only) that comes to mind to allow in trailers. Thus I'd argue that we should do this now.
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

--
Mark Nottingham   https://www.mnot.net/

reschke:
It think the issue here manifests itself in the issue title: it claims that this is a clarification, while it really is a normative change. Yes, it may reflect reality and be a good change to do, but it *is* a change.

We need to summarize this change in the "Changes" section, and address the consequences (header fields potentially not allowed in trailers anymore).

mnot:
@royfielding would it address your concern if we added a para that stated why some fields are not allowed as trailers (as per the previous text)?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1711,6 +1711,78 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
 </section>
 </section>
 
+<section title="Trailer Fields" anchor="trailer.fields">
+   <iref item="trailer fields"/>
+   <iref item="trailers"/>
+
+<section title="Purpose" anchor="trailers.purpose">
+<t>
+   In some cases specific to the transfer coding or HTTP version, additional
```
mnot: "In some HTTP versions, additional..."

or

"In some HTTP versions (e.g., using transfer codings in HTTP/1.1)..."

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1711,6 +1711,78 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
+   metadata can be sent after the initial header section has been completed
+   (during or after transmission of the payload body) that might be
+   dynamically generated while the body is being sent, such as a message
```
mnot: "that might be dynamically generated while the body is being sent" makes this sentence really wordy. Suggest deleting or moving it.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1711,6 +1711,78 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
+   integrity check, digital signature, or post-processing status.
+   For example, the chunked coding in HTTP/1.1 allows a trailer section after
```
mnot: New para?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1711,6 +1711,78 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
+   the payload body (<xref target="chunked.trailer.section"/>) which can contain
+   trailer fields: field names and values that share the same syntax and
+   namespace as header fields but are received after the header section.
+</t>
+<t>
+   Trailer fields ought to be processed and stored separately from the fields
+   in the header section to avoid overriding the message semantics known at
```
mnot: Hmm. Is "overriding" right? "conflicting" seems closer to me...

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1711,6 +1711,78 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
+   Many header fields cannot be processed outside the header section because
+   their evaluation is necessary prior to receiving the message body, such as
+   fields that describe message framing, routing, authentication,
+   request modifiers, response controls, or payload format.
+   A sender &MUST-NOT; generate a trailer field unless the sender knows the
+   corresponding header field name's definition permits the field to be sent
+   in trailers.
+</t>
+<t>
+   Because of the potential for trailer fields to be discarded in transit, a
```
mnot: It feels like we should say a little bit more about this; right now we're relying on the reader to infer that some HTTP versions don't support trailers, that intermediaries can drop them, and that some implementations just drop them on the floor. Probably worth including in "purpose" above.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1711,6 +1711,78 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
+  <x:ref>Trailer</x:ref> = 1#<x:ref>field-name</x:ref>
+</sourcecode>
+<t>
+   For example, a sender might indicate that a message integrity check will
+   be computed as the payload is being streamed and provide the final
+   signature as a trailer field. This allows a recipient to perform the same
+   check on the fly as the payload data is received.
+</t>
+<t>
+   A sender that intends to send a non-empty trailer section &SHOULD; generate a
```
mnot: `trailer-section` is HTTP/1.1-specific potentially, no? Might be better to just say "A sender that intends to send one or more trailers..."

mnot:
Also, as I mentioned on the issue, older implementations might still merge trailers, so recipients SHOULD disregard headers that are only defined to appear in trailers.

Maybe in "Limitations"?

draft-ietf-httpbis-semantics-latest.xml
royfielding: yep, first alternative is better because next sentence uses chunked as example.

draft-ietf-httpbis-semantics-latest.xml
royfielding: deleted

draft-ietf-httpbis-semantics-latest.xml
royfielding: it's just 2 sentences and looks better after the above deleted.

draft-ietf-httpbis-semantics-latest.xml
royfielding: contradicting

draft-ietf-httpbis-semantics-latest.xml
royfielding: intends to generate, yes.

mnot:
LGTM. @reschke ?

PR #221, opened 2019-04-12
Adjust text defining what an incomplete response is

### LINKED ISSUE #25: Content framing and refresh
It'd be good to say something about how weakly-framed content (e.g., no Content-Length in 1.0; no CL or chunked in 1.1) should be handled differently by caches, especially during reload.

From https://github.com/httpwg/http-extensions/issues/325

mnot:
Fixes #25.


mnot:
@royfielding yes, as per issue discussion I was reconsidering this. See updated PR.

mnot:
yeah, I like that. Will update.

PR #222, opened 2019-04-12
adjust requirements for retries

### LINKED ISSUE #27: Retries
As discussed quite a bit, retries in HTTP need to be better defined. In particular, HTTP does not offer a guarantee that any particular request will not be automatically retried.

mnot:
Fixes #27.


mnot:
See #27; the issue is not just about moving the text.

reschke:
I understand that, I just thought it would be good if we could split this into two changes; that makes it easier later on...

reschke:
See #230 - I'm on it.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3926,23 +3930,29 @@ Content-Range: exampleunit 11.2-14.3/25
...
-   actually idempotent, regardless of the method, or some means to detect that
-   the original request was never applied. For example, a user agent that
-   knows (through design or configuration) that a POST request to a given
-   resource is safe can repeat that request automatically.
-   Likewise, a user agent designed specifically to operate on a version
-   control repository might be able to recover from partial failure conditions
-   by checking the target resource revision(s) after a failed connection,
-   reverting or fixing any changes that were partially applied, and then
-   automatically retrying the requests that failed.
+   A client &SHOULD-NOT; automatically retry a request with a
```
reschke: This changes "user-agent" to "client", but a subsequent use of "user-agent" stays. Intentional?

mnot: Yes. This allows non-proxy intermediaries to do this, which they already do.

reschke:
I think I'm ok with the actual changes, but we need a summary for the "Changes from RFC 723*" section.

Proposal:

> Requirements on when requests can be retried (and by whom) have been slightly relaxed (see ...).

mnot:
@reschke I think this is ready to go.

reschke:
We still need to document this as a normative change from RFC 723*.

mnot:
Done.

mnot:
LGTM.

Issue #224, opened 2019-05-06
Status code 422 uses the term `entity`.


ioggstream:
## I expect

The term entity to be dismissed, in favor of `representation`

## Instead 

It's used here, as 422 is C&P from webdav

https://github.com/httpwg/http-core/blob/0fd2e719c0fd8ceb7ea0c11a0b3134584dc8fbc1/draft-ietf-httpbis-semantics-latest.xml#L7946

PR #225, opened 2019-05-06
Fix: #224. Use representation instead of entity

ioggstream:
## This PR

Fixes 422 definition as HTTP Semantics replaces entity with representation since RFC7231.

## Notes

We could refine this, stating if 422 applies only to resource representation
or may refer to request-target too (eg. query parameters).


ioggstream:
@LPardue in-topic for digest :)

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7943,9 +7943,9 @@ Content-Range: bytes */47022
   <x:anchor-alias value="422 (Unprocessable Entity)"/>
```
ioggstream: Should we retain the Entity for backward compatibility?

reschke:
See https://github.com/httpwg/http-core/pull/226

PR #226, opened 2019-05-15
Fix description of status code 422 not to use "entity" (also tune prose) (see #224)

reschke:
…se) (see #224)


reschke:
I chose "Payload" in the reason phrase for consistency with 413...

We *do* use "enclosed representation" in many other places, such as in the definitions of PUT, 401, and Content-Location.

Maybe use "payload" throughout?

(I guess we have way more terminology to clean up)

reschke:
@ioggstream it's not incorrect, but sure we can tune things even more...

Issue #62, opened 2018-04-30
ABNF message format


reschke:
We currently have:

~~~
  HTTP-message   = start-line
                   *( header-field CRLF )
                   CRLF
                   [ message-body ]
~~~

Maybe it would be more consistent to remove CRLF from start-line, and then to have

~~~
  HTTP-message   = start-line CRLF
                   *( header-field CRLF )
                   CRLF
                   [ message-body ]
~~~

mnot:
LGTM.

reschke:
...waiting until https://github.com/httpwg/http-core/issues/197 is resolved...

PR #232, opened 2019-07-06
Move the trailing CRLF from the line grammars into the message format (#62)

reschke:


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -414,11 +415,11 @@
   <x:anchor-alias value="request-line"/>
 <t>
    A request-line begins with a method token, followed by a single
-   space (SP), the request-target, another single space (SP), the
-   protocol version, and ends with CRLF.
+   space (SP), the request-target, another single space (SP), aned ends
```
mnot: s/aned/and/

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -249,14 +249,15 @@
 <iref item="headers"/>
 <iref item="header field"/>
 <t>
-   All HTTP/1.1 messages consist of a start-line followed by a sequence of
+   All HTTP/1.1 messages consist of a start-line followed by a CRLF and a
```
royfielding: I think this should be:

> All HTTP/1.1 messages consist of an initial line of control data (terminated by CRLF) followed by optional header fields, an empty line terminating the header section, and an optional message body.

Alternatively, go ahead and merge this PR and we can make an editorial change later.

reschke: Yes, I'd prefer to do that separately, as it looses the RFC 5322 reference...

Issue #119, opened 2018-07-05
Include RFC 7694


reschke:
From the abstract:

> In HTTP, content codings allow for payload encodings such as for compression or integrity checks. In particular, the "gzip" content coding is widely used for payload data sent in response messages.

> Content codings can be used in request messages as well; however, discoverability is not on par with response messages. This document extends the HTTP "Accept-Encoding" header field for use in responses, to indicate the content codings that are supported in requests.

mnot:
Discussed in Montreal; support for doing this.

Issue #6, opened 2015-09-03
Security considerations: discuss BREACH in the context of gzip encoding


reschke:
See https://lists.w3.org/Archives/Public/ietf-http-wg/2015JulSep/0295.html.

royfielding:
I don't know what this issue is about. Was it intended for CICE?

wtarreau:
The attack was described here : http://breachattack.com/

In short, if you can force a client to emit requests and affect user input that are reflected in compressed responses, you can guess some parts of the response by observing its length only, hence bypass encryption. If the response contains some secrets (tokens to access another site, etc) then they can be guessed this way.

I've only been aware of one of our users who was worried about this and wanted to disable Gzip compression to stay safe... I guess nobody really cares because there are too many "if" for the attack to be effective. But by mentioning it we could at least encourage implementations to leave the choice to the user when to compress or not. For example in haproxy we still don't provide an easy way to decide whether to compress or not on a per-request basis, and we should work on it at least to mitigate this attack (e.g. disable compression on responses to POST requests maybe or on non-cacheable responses).

martinthomson:
As Willy says, this general class of attacks relies on the combination of generic compression (like gzip) and a mix of user-controlled content and secrets.

There are a lot of "if"s on this, but the web context is probably the most fraught.  CICE doesn't come into this too much, because an attacker generally only can either control the entire request (but won't have the ability to access secrets in doing so) or just the URL and maybe the request body.  None of the request contexts really have this unique blend of attacker-controlled and secret information.

However, for responses, it is common for user-generated content to be mixed with secrets.  For instance, pages routinely put CSRF tokens into HTML that also contain user-generated messages.  Compress those and you can do interesting things. There are plenty of ways in which the attack can be foiled, even accidentally, but also plenty of ways in which information can leak out.  I suspect that the authors of the several papers in these areas struggled a little to find vulnerable sites because of this.

All that said, it's worth having some text explaining the basic setup.  If you need a volunteer, I might be able to do something.

royfielding:
Sorry, I should have been more specific. I know what the breach attack is about. What I don't know is the intended target of this issue. The IESG comment was on the CICE draft.

I will guess that @reschke created this issue because we are expected to address the attack somewhere in the core protocol so that we don't have to deal with describing it as a special case in all extension drafts. But where? Oh, never mind, that's why the issue title says security considerations.

reschke:
[...]

PR #238, opened 2019-08-26
include RFC 7694 (#119)

reschke:
This includes the extension, but leaves out:

- examples
- deployment considerations
- security considerations (but see https://github.com/httpwg/http-core/issues/6)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5708,9 +5725,20 @@ Expect: 100-continue
...
-   target="content.codings"/>). An "identity" token is used as a synonym for
+   The "Accept-Encoding" header field can be used to indicate preferences
+   regarding the use of content-codings (<xref target="content.codings"/>).
+</t>
+<t>
+   When used by a user agent in a request, it applies to the content codings
+   acceptable in the response.
+</t>
+<t>
+   On the other hand, when used by a server in a response, it provides
```
mnot: nit: "On the other hand" is a bit wordy

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5759,6 +5787,37 @@ Expect: 100-continue
    is listed as acceptable, the origin server &SHOULD; send a response
    without any content-coding.
 </t>
+<t>
+   When the Accept-Encoding header field is present in a response, it indicates
+   what content codings the resource was willing to accept in the associated
+   request. The field value is evaluated the same way as in a request.
```
mnot: "associated request" doesn't seem appropriate here; we use that elsewhere to mean the request in a specific exchange, and so by definition this information can't ever be used (unless we want to get into the exact semantics of what a repeated / retried request is -- but that's another windmill).

I'd suggest something like "... what content codings the resource is currently willing to accept in requests similar to that associated with the response." 

And below: "Note that this information is a hint only; the server's willingness to accept an encoding can be influenced by many factors, including resource identity, time, request method, and other factors."

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5759,6 +5787,37 @@ Expect: 100-continue
...
+   include an Accept-Encoding header field in that response, allowing
+   clients to distinguish between issues related to content codings and media
+   types. In order to avoid confusion with issues related to media types,
+   servers that fail a request with a 415 status for reasons unrelated to
+   content codings <bcp14>MUST NOT</bcp14> include the Accept-Encoding header
+   field.
+</t>
+<t>
+   The most common use of Accept-Encoding is in responses with a
+   <x:ref>415 (Unsupported Media Type)</x:ref> status code, in response to
```
mnot: Maybe it's time for a new status code...

draft-ietf-httpbis-semantics-latest.xml
reschke: just remove it?

draft-ietf-httpbis-semantics-latest.xml
reschke: But that's exactly what it means here (the request leading to the response being discussed)...

I'm ok with the rephrased note, but it really doesn't seem to say anything different (the current text in the PR is from <https://greenbytes.de/tech/webdav/rfc7694.html#rfc.section.3.p.3>.

draft-ietf-httpbis-semantics-latest.xml
mnot: I think so.

draft-ietf-httpbis-semantics-latest.xml
mnot: I see. It might be good to explicitly state something like "Clients can use this information to optimistically encode future requests. However, because this information is specific to the associated request..."

reschke: Not sure here. Clients always could use content-codings in requests. All we add here is some signalling from server to client that gives more information. And the paragraph already clearly says what it applies to, and when it could change.

Can you propose some concrete text change?

reschke:
Hm. My understanding is that content encodings are fully bidirectional (and have been all the time in theory), thus this would be exactly the right thing to do.

In your proposal, would we have two different sections for Accept-Encoding (request and response)?

royfielding:
What I should have said is that defining a bunch of information about response header fields inside the main section on request header fields doesn't work. Moving it to a new section within the general discussion of content negotiation might work.

I have gone ahead and done that here. The section on Accept-Encoding still contains both request and response definitions, which seems fine.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4236,6 +4240,57 @@ Content-Range: exampleunit 11.2-14.3/25
...
+    which content types are accepted in PATCH requests.
+  </t>
+</aside>
+</section>
+
+<section title="Quality Values" anchor="quality.values">
+  <x:anchor-alias value="weight"/>
+  <x:anchor-alias value="qvalue"/>
+<t>
+   The content negotiation fields
```
mnot: Either "The content negotiation fields defined by this specification" or "some content negotiation fields."

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4096,12 +4098,14 @@ Content-Range: exampleunit 11.2-14.3/25
...
    be made visible within the protocol:
-   "proactive", where the server selects the representation based
-   upon the user agent's stated preferences, and "reactive" negotiation,
+   "proactive" negotiation, where the server selects the representation based
+   upon the user agent's stated preferences, "reactive" negotiation,
    where the server provides a list of representations for the user agent to
-   choose from. Other patterns of content negotiation include
+   choose from, and "request payload" negotiation, where the user agent
+   selects the representation for a future request based upon the server's
+   stated preferences. Other patterns of content negotiation include
```
mnot: "stated preferences in past responses." ?

mnot:
Looks better, left two small-ish comments.

draft-ietf-httpbis-semantics-latest.xml
reschke: done

draft-ietf-httpbis-semantics-latest.xml
reschke: done

Issue #47, opened 2018-02-27
Guidelines for header field names


mnot:
Is it useful to give some guidance about the use of prefixes like:

* Accept-
* Allow-
* Expect-

etc.? In particular, we could say that you can't depend on the semantics of a prefix, but it's a good idea to align your use with existing prefixes.

mnot:
I think so, but probably separately, since the ones above don't trigger any automatic handling, whereas `Sec-` does.

royfielding:
I don't think it is true that they don't trigger automatic handling -- that happens for Accept-* and Content-*, for example.

I also don't think we should encourage application-specific prefixes unless we can also accurately define what application-specific means and avoid repeating the mistake of x-prefixes. In general, it is better to encourage broad standardization and ignore the ones that failed, rather than encourage prefixes and then have to make a standard out of them after deployment.

I do like the tweak to encourage shorter names.

mnot:
> I don't think it is true that they don't trigger automatic handling -- that happens for Accept-* and Content-*, for example.

How so? 

> I also don't think we should encourage application-specific prefixes unless we can also accurately define what application-specific means and avoid repeating the mistake of x-prefixes. In general, it is better to encourage broad standardization and ignore the ones that failed, rather than encourage prefixes and then have to make a standard out of them after deployment.

"limited-use, such as a header confined to a single application or use case"?

MikeBishop:
More generally, if I come along and define something new in an existing family -- let's say `Accept-Species:`, to indicate I only want resources that include cat pictures -- is there any logic that will treat it differently because it's Accept-*, even if unknown?  Will I start seeing `Content-Species:` headers in responses?  I'm dubious, though I could obviously be wrong.

Even `Sec-` I find questionable, because it's not the HTTP layer that's attributing any specific meaning to that prefix.

At the HTTP layer, we don't care what format a header name takes; it's merely an opaque key in a dictionary.  However, as a best practice, aligning your naming with similar headers aids developer comprehension and debugging.

royfielding:
> "limited-use, such as a header confined to a single application or use case"?

yes, that's much better.

PR #240, opened 2019-09-02
Rewrite header field name guidelines

mnot:
For #47.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -4414,8 +4414,6 @@ Proxy-Authorization: basic aGVsbG86d29ybGQ=
    in a machine or human-readable representation. A standard format for such a
    representation is not defined by this specification, but might be defined by
    future extensions to HTTP.
-   A server &MUST; generate a <x:ref>Content-Length</x:ref> field with a value
-   of "0" if no payload body is to be sent in the response.
 </t>
 <t>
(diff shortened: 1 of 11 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1489,14 +1489,27 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
-   (The "X-" prefix idiom has been extensively misused in practice; it was
-   intended to only be used as a mechanism for avoiding name collisions inside
-   proprietary software or intranet processing, since the prefix would ensure
-   that private names never collide with a newly registered Internet name; see
-   <xref target="BCP178"/> for further information).
+   Authors of specifications defining new fields are advised to choose a short
+   but descriptive field name. Short names avoid needless data transmission
+   (especially when a header field is sent often); descriptive names avoid
+   confusion and "squatting" on names that might have broader uses.
+</t>
```
reschke: Isn't that only true for HTTP/1.1?

mnot: No; [HQ]Pack aren't magical, they have limited-size state tables that can be consumed by overly long header names.

reschke: Yes, that's understood. But I think this is still will cause people tp be confused, in particular as it says "when ... sent often".

mnot: OK, removing the parenthetical.

PR #241, opened 2019-09-02
Change response cacheability to only require understanding in certain…

### LINKED ISSUE #120: Status codes and caching
7234 [says](https://httpwg.org/specs/rfc7234.html#response.cacheability) that a response can't be stored if the cache doesn't understand its status code.

Need to dig around in the history, but I wonder if that's right.

mnot:
… circumstances

Fixes #120.


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -331,7 +331,10 @@
   <li><t>The request method is understood by the cache, and</t></li>
   <li><t>the response status code is final (see <xref
   target="associating.response.to.request"/>), and</t></li>
-  <li><t>the response status code is understood by the cache, and</t></li>
+  <li><t>the response status code is understood by the cache, if it is 206,
```
reschke: It took me a moment to parse this. Why do 206 and 304 show up here now? Is this adressing the same issue? Maybe a separate paragraph about what status codes a cache MUST understand would help?

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1480,6 +1483,15 @@
 </t>
 </section>
 
+<section title="no-store" anchor="cache-response-directive.must-understand">
+   <iref item="must-understand (cache directive)" primary="true" />
+<t>
+   The "must-understand" response directive indicates that a cache &MUST; 
+   understand and implement the caching semantics of the response to be able
+   to store it; if it does not, it &MUST-NOT; store the response.
+</t>
```
reschke: I'd replace the second bcp14 keyword by prose; it's the same requirement, no?

draft-ietf-httpbis-cache-latest.xml
mnot: Yes, this was the interesting part. Partial content and validation responses are two parts of HTTP where a cache has to be aware of the semantics of the response to store it; if they aren't, bad things can happen -- and those responses often do carry caching metadata. So calling them out explicitly seemed wise.

This is definitely part of this issue; it's assumed in the current text, but now needs to be explicit because of the change. 

Would it clarify to split these out into a separate bullet point, with a small amount of explanatory text?

draft-ietf-httpbis-cache-latest.xml
mnot: Fixed.

draft-ietf-httpbis-cache-latest.xml
reschke: yes.

reschke:
The combination of "unless..." with the conditional list items still scares me.

Maybe define "needs-to-be-understood" status codes upfront, and then simplify the list?

mnot:
The only place that "unless" occurs is in the change notes.

reschke:
No, it's in the statement that introduces the list.

mnot:
That sounds like a separate issue; that text is from 7234.

reschke:
Yes, but I think that combining this with conditionals in the list makes things hard to understand.

mnot:
There are already conditionals in the list; this hasn't caused any confusion to date (and the overwhelming feedback we got in the reviews of 7234 was that this form was much easier to understand).

If you still feel that way, please raise a separate issue.

mnot:
I don't think that helps; the grouping of the _and_ and _or_ is ambiguous (which is why I split it into two list items).

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -331,7 +331,11 @@
   <li><t>The request method is understood by the cache, and</t></li>
   <li><t>the response status code is final (see <xref
   target="associating.response.to.request"/>), and</t></li>
-  <li><t>the response status code is understood by the cache, and</t></li>
+  <li><t>the response status code is understood by the cache, if it is 206 or
+  304 (because these responses need special handling by caches), and</t></li>
```
MikeBishop: Pre-existing issue:  You're using commas inside a comma-delineated list; the list should be using semicolons.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1480,6 +1484,15 @@
 </t>
 </section>
 
+<section title="no-store" anchor="cache-response-directive.must-understand">
```
MikeBishop:
```suggestion
<section title="must-understand" anchor="cache-response-directive.must-understand">
```

mnot:
@royfielding any other suggestions? I'd like to merge this soon.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1480,6 +1484,15 @@
 </t>
 </section>
 
+<section title="must-understand" anchor="cache-response-directive.must-understand">
+   <iref item="must-understand (cache directive)" primary="true" />
+<t>
+   The "must-understand" response directive indicates that a cache &MUST; 
+   understand and implement the caching semantics of the response to be able
+   to store it; if it does not, it cannot store the response.
+</t>
```
reschke: Does this apply to just the status code, or to more elements of the response?

mnot: I think it applies to the response status code, and whatever else it requires. Make sense? If so, will clarify.

royfielding:
@mnot I think this is now fixed but needs your careful review

mnot:
@royfielding honestly, I really don't like it; it's more confusing.

Can I suggest we revert that and split some of the conditional / complex requirements out into separate paragraphs (with their own 2119 language), rather than making one big list?

royfielding:
Sure, but keep in mind that this entire list is already duplicating the requirements stated in more detail in later sections. Perhaps we should simply replace it with an informative checklist pointing to each relevant section?

draft-ietf-httpbis-cache-latest.xml
reschke: ...which makes me think that the cache control directive should be called "must-understand-status"...

PR #243, opened 2019-09-02
Recommend minimum URI lengths.

### LINKED ISSUE #169: URI length
[Messaging section 3](https://httpwg.org/http-core/draft-ietf-httpbis-messaging-latest.html#request.target) says:

> Various ad hoc limitations on request-line length are found in practice. It is recommended that all HTTP senders and recipients support, at a minimum, request-line lengths of 8000 octets.

The primary intent of that text (added in 7230) AIUI was to encourage implementations to support longer URIs. As such, I was looking for it in semantics, not messaging.

I think the best thing would be to add text somewhere in [the semantics section about URIs](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#uri) saying that implementations should support HTTP(S) URIs of at least 8000 characters.

mnot:
Fixes #169.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -781,6 +781,12 @@ Content-Type: text/plain
    an error; it is likely being used to obscure the authority for the sake of
    phishing attacks.
 </t>
+<t>
+   It is &RECOMMENDED; that all senders and recipients support, at a
+   minimum, http URIs with lengths of 8000 octets. Note that this implies
+   some structures and on-wire representations (for example, the request line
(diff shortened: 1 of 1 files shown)


mnot:
I think we could refactor to have one section for both URI schemes, but that's  different issue.

reschke:
One section for both schemes would make it harder to reference a specific one.

I think adding a section on limits would fit the overall structure well.

mnot:
Yeah, that should work; PTAL. 

While we're there, the split between 2.4 and 2.5 is odd; to me, it'd make more sense if 2.5 became 2.4.1, and 2.5.* became 2.4.* afterwards.

Issue #233, opened 2019-07-09
Caching Security Considerations


mnot:
... needs a rewrite. 

I'm thinking of splitting up into distinct sections, e.g.,

* Cache Poisoning
* Timing Attacks
* Caching of Sensitive Information
* Information Leakage

... with appropriate mitigations in each (e.g., double keying).

PR #244, opened 2019-09-02
Refactor security considerations

mnot:
For #233.


mnot:
@reschke it didn't so much introduce it as much as it just changed the nature of the example.

reschke:
Well, before the term "timing attack" did not appear in the spec...

PR #246, opened 2019-09-02
update port references

### LINKED ISSUE #36: Port registration
Just a reminder to update the port registration appropriately.

mnot:
Fixes #36.


mnot:
No, all entries for 80 and 443 that use TCP or UDP. I have no opinion about SCTP, and don't want to take over authority for it in these specs.

I think the instruction in the PR is sufficient; all we're doing is changing (or adding) references.

reschke:
My experience with IANA is that they want precise instructions. It also helps readers/reviewers understand what's happening.

reschke:
See https://github.com/httpwg/http-core/pull/246/commits/8835112d77b9c3d218c9032fa17f29bdc0761af9 - which begs the question whether we should also update assignee/contact for "www-http" which are currently assigned to TimBL.

mnot:
Works for me. @reschke you now OK?

My instinct is to leave www-http alone.

PR #247, opened 2019-09-02
Move text about CRLF up

### LINKED ISSUE #68: message parsing robustness
... is awfully far from the definition of the message format; IME people read the start of section 3 in isolation and don't see the text below. Maybe move 3.5 up?

mnot:
Fixes #68.


mnot:
As suggested in the issue. Looking at it, more text could be moved up, or the entire section could be moved up (e.g., as a subsection of where this text went), but this seemed reasonable.

mnot:
-1; this needs to be where CRLF is used, not a different section.

mnot:
Like that? Wasn't sure what you meant by "moving HTTP version up..."

royfielding:
> Like that? Wasn't sure what you meant by "moving HTTP version up..."

I meant literally move the section titled "HTTP Version" up above the section titled "Message" and then move "Parsing" to a subsection of "Message". I'll see if I can tweak it.

reschke:
Would this close #68?

PR #249, opened 2019-09-03
Bring RFC2818 into semantics (Mnot 236)

### LINKED ISSUE #236: Bring RFC2818 into semantics
As discussed in #194.

mnot:
Fixes #236


mnot:
I *think* I've brought the relevant parts across into what seems like the appropriate sections; haven't modified text very much at all. The text that I didn't bring over didn't seem to add much, but I'm happy to be talked around on that.

N.B. we'll need to update the reference from 5246 to 8446 in semantics, but that's separable (and I assume we'll do a reference sweep when we're closer to done anyway).

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -2009,6 +2009,33 @@ https://www.example.org
...
+   send its closure alert, generating an "incomplete close".  Note that
+   an implementation which does this &MAY; choose to reuse the session.
+   This &SHOULD; only be done when the application knows (typically
+   through detecting HTTP message boundaries) that it has received all
+   the message data that it cares about.
+</t>
+<t>
+   As specified in <xref target="RFC8446"/>, any implementation which receives a
+   connection close without first receiving a valid closure alert (a
+   "premature close") &MUST-NOT; reuse that session.  Note that a
```
MikeBishop: I don't know that I've noticed this requirement before.  Do implementations actually make this distinction?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9628,15 +9630,85 @@ Content-Encoding: gzip
...
+   not bar.foo.a.com. f*.com matches foo.com but not bar.com.
+</t>
+<t>
+   In some cases, the URI is specified as an IP address rather than a
+   hostname. In this case, the iPAddress subjectAltName must be present
+   in the certificate and must exactly match the IP in the URI.
+</t>
+<t>
+   If the hostname does not match the identity in the certificate, user
+   oriented clients MUST either notify the user (clients MAY give the
```
MikeBishop: `user-oriented`, no?

reschke: is a "user(-)oriented client" different from a "user agent"?

mnot:
ACK'd. I think additional issue should be separate. PTAL.

draft-ietf-httpbis-messaging-latest.xml
royfielding: we can look into this when we trim this down later

PR #250, opened 2019-09-03
Tighten language around GET request bodies (Mnot 202)

### LINKED ISSUE #202: Tighten language around GET request bodies
I'm active on Stack Overflow a lot, and one thing that occasionally pops up is: Can a GET or DELETE request have a request body?

My answer to this currently stands at a score of -1: https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request/54735278#54735278

But I think it's correct.

This is the wording for `DELETE`:

> A payload within a DELETE request message has no defined semantics;
> sending a payload body on a DELETE request might cause some existing
> implementations to reject the request.

Some people interpret 'no defined semantics' as, "I can define the semantics myself". That interpretation is somewhat similar to how someone might look at a `POST` request and argue: the HTTP standard doesn't really say what the effect of a `POST` request is.

So English and well, semantics aside... My understanding is that the intent for those paragraphs is so that generic parsers are possible, and a generic parsers don't need to know specifics about the methods to figure out whether to parse the body or not.

However, a real world effect is that:

1. People misinterpret the paragraph and assign a new meaning.
2. Others outright reject requests with a body.
3. Some of those in group #2 are intermediaries and cause problems for group #1.

I think the specification could benefit from stronger wording. I think even a sentence like this would help:

> Implementations SHOULD not add a payload to DELETE requests.

But I would argue that, given that this is creating some interop issues that this is even more preferable:

> Implementations SHOULD reject DELETE requests with a payload.

mnot:
Fixes #202.


mnot:
I'm fine with that; it was borderline.

mnot:
Change note added.

Creating a new issue doesn't make any difference; they're just how we track what we need to do. 

IIRC there was one person who mentioned an existence proof of a DELETE body -- which is very analogous to the situation with GET. We know that some people do them, but we also know that they're not interoperable, and this is not an intended extension of the protocol.

reschke:
I strongly believe that the issues for GET and DELETE are *very* different, and that handling them in a single ticket is harmful.

Can we please focus on GET first?

mnot:
Please update the PR and create the issue, then.

reschke:
will do.

asbjornu:
Would it be possible to mention the problems around caching with a body on `GET` requests? As `GET` with a body is in use by large vendors such as Elasticsearch (see elastic/elasticsearch#16024) and their interpretation of the spec being that this is "just fine" (see https://github.com/elastic/elasticsearch-php/issues/737#issuecomment-378590773), I think a stronger language should be used to discourage implementations from using a body with `GET` requests.

This also brings up what we discussed at the HTTP WG around creating a header to declare that the HTTP body should also be considered a part of the cache key, as well as the possibility of creating a new `SEARCH` method ([currently drafted](https://tools.ietf.org/html/draft-snell-search-method-01) by @jasnell).

reschke:
@asbjornu - the pull request has this:

>  A payload within a GET request message has no defined semantics. A client SHOULD NOT generate a body in a GET request. Sending a payload body on a  GET request might cause some existing implementations to reject the request.

IIUC, you'd like this also to mention other failure modes such as GET request suceeding, but result might be inproperly cached?

(the other points IMHO are out of scope for the HTTP core revision; maybe they should be tracked in the http-extensions repo instead).

asbjornu:
@reschke:

> IIUC, you'd like this also to mention other failure modes such as GET request suceeding, but result might be inproperly cached?

Yes, exactly. Just as an informational warning so implementers understand the dangers they are putting themselves into by ignoring "SHOULD NOT".

> (the other points IMHO are out of scope for the HTTP core revision; maybe they should be tracked in the http-extensions repo instead).

Should I create an issue for `SEARCH` and one for a cache header in the [`httpwg/http-extensions` repository](https://github.com/httpwg/http-extensions)?

reschke:
1) Ack. @mnot ?

2) Yes.

asbjornu:
@reschke: Ok, httpwg/http-extensions#942 and httpwg/http-extensions#943 created.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4068,9 +4068,10 @@ Content-Range: exampleunit 11.2-14.3/25
    <x:ref>Range</x:ref> header field in the request (<xref target="header.range"/>).
 </t>
 <t>
-   A payload within a GET request message has no defined semantics;
-   sending a payload body on a GET request might cause some existing
-   implementations to reject the request.
+   A payload within a GET request message has no defined semantics. A client
+   &SHOULD-NOT; generate a body in a GET request. Sending a payload body on a
+   GET request might cause some existing implementations to reject the
+   request.
```
reschke: "... or might cause surprising effects in HTTP caches which ignore the payload."

mnot: perhaps "... or might cache caches to reuse the response erroneously."

PR #252, opened 2019-09-03
Exact separator used when combining header fields (Mnot 148)

### LINKED ISSUE #148: Exact separator used when combining header fields
Fetch and XMLHttpRequest have for the longest time used comma-followed-by-space (0x2C 0x20) as separator when multiple header fields are combined into a single header field.

This is also what at least Firefox and Chrome do from code inspection:

* https://searchfox.org/mozilla-central/source/netwerk/protocol/http/nsHttpHeaderArray.h#256
* https://cs.chromium.org/chromium/src/net/http/http_response_headers.cc?l=465

I think this is because of https://tools.ietf.org/html/rfc2616#section-4.2 encouraging values to start with a single space and folks conceptually still seeing it as multiple values.

Now https://tools.ietf.org/html/rfc7230#section-3.2.2 has a much more explicit definition, which no longer allows for a space as far as I can tell, "breaking" with what implementations are doing.

Most of the time this difference is insignificant, but anything with quoted-string or equivalent will be able to tell what was used if those configuring the headers like to play games.

I'd like HTTP to either allow 0x2C 0x20 as an alternative separator (and have Fetch mandate it for browsers) or simply adopt 0x2C 0x20 as separator for everyone.

I have some tests around this as well, but they only affect APIs, which theoretically could have a different non-HTTP-conformant view, but that'd seem really bad: https://github.com/web-platform-tests/wpt/pull/13471. See https://github.com/whatwg/fetch/pull/813 and https://github.com/whatwg/fetch/issues/752 for further context.

cc @ddragana @MattMenke2 @youennf

mnot:
Fixes #148.


mnot:
That would take us to something like:

> A recipient &MAY; combine multiple header fields with the same field name into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field value (after stripping surrounding OWS) to the combined field value in order, separated by COMMA. For consistency, implementations SHOULD follow the COMMA with a SP.

I don't think that's materially better than what's currently in the PR, but if it will get us there, sure.

martinthomson:
Is the surrounding OWS part of that consistency thing?  I mean, you could just drop the comma in and leave spaces in place.

reschke:
"stripping surrounding OWS" is new too (to be discussed), and not needed as field-value by definition has not leading/trailing WS.

Adding an SP can't be a SHOULD, because it would potentially make existing implementations non-compliant, with no good reason for that.

royfielding:
For consistency, a sender ought to generate the same meaningless whitespace as everyone else.

(just kidding -- I don't see a need to spec this other than as a note somewhere about caching and consistency)

MikeBishop:
> Adding an SP can't be a SHOULD, because it would potentially make existing implementations non-compliant, with no good reason for that.

@reschke, maybe there are different existing principles on what's non-compliant, but the point of a SHOULD is that you're not non-compliant if you do something different.  I don't think a recommendation breaks anything; if it were a MUST, we'd obviously have a problem.

reschke:
@MikeBishop  - "SHOULD" ought to come with a reason; here I don't see a compelling one. It really doesn't matter.

MikeBishop:
Well, if it doesn't matter but we're documenting common practice, maybe we just say that?  Changing the last sentence to "Many implementations insert a SP after the comma to improve readability." is correct, but non-normative.

mnot:
The point of the issue AIUI is *not* to just document practice, but to have a normative way to combine header instances, so that a header whose value is split across multiple lines is combined in a deterministic way.

In particular, if a header contains a whitespace-sensitive string.

Some seem to be arguing that this isn't important to specific because splitting a header value across multiple instances isn't allowed. Others seem to be arguing that this is seen in the wild, so it's good to specify behaviour. 

Personally, I don't see a problem with documenting a SHOULD here; it is very widespread practice, and I think we ought to document things as precisely as we can, within reason.

reschke:
I don't think "widespread" is sufficient for a normative change, *unless* there's clearly a problem being solved. Is there?

I agree that there is a potential issue with broken messages being interpreted differently when using different separators, but I don't think that's sufficient reason for introducing a new normative requirement.

mnot:
"widespread" wasn't being used to motivate the change; it was establishing that it would be aligned with current practice. The motivation is in the next sentence -- but I do think they're disagreement about how to value it.

We should really move discussion over to #148.

reschke:
AFAICT, this is OBE. See https://github.com/httpwg/http-core/pull/309

Issue #159, opened 2018-10-16
"Message Body" needs to be in Semantics


mnot:
[this text](
https://httpwg.org/http-core/draft-ietf-httpbis-messaging-latest.html#message.body) isn't version-specific, and is referenced from Semantics 2.1 Client/Server messaging. 

I'd put it in Semantics 4. Message Abstraction.

mnot:
Proposal is rough, but ready for feedback.

PR #253, opened 2019-09-03
Move version non-specific parts of message body into semantics

mnot:
For #159.

### DIFF
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -952,7 +952,7 @@ https://www.example.org
    message differ for requests and responses.
 </t>
 <t>
-   The presence of a message body in a HTTP/1.1 request is signaled by a
+   The presence of a message body in a request is signaled by a
    <x:ref>Content-Length</x:ref> or <x:ref>Transfer-Encoding</x:ref> header
    field. Request message framing is independent of method semantics,
    even if the method does not define any use for a message body.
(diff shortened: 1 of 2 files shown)


mnot:
Roughed in; PTAL.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -939,39 +939,29 @@ https://www.example.org
...
 </sourcecode>
 <t>
-   The rules for when a message body is allowed in a message differ for
-   requests and responses.
+   The rules for determining when a message body is present in a HTTP/1.1
+   message differ for requests and responses.
 </t>
 <t>
-   The presence of a message body in a request is signaled by a
+   The presence of a message body in a HTTP/1.1 request is signaled by a
```
reschke: Not sure we need to say HTTP/1.1 everywhere - this is the topic of this spec, right?

mnot: I took that one out, but left the first one to remind readers that this is 1.1-specific.

mnot:
I don't any requirements were lost, they're just incorporated by reference, since they apply to more than one version of HTTP.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -939,17 +939,17 @@ https://www.example.org
...
+   message body is identical to the payload body unless a transfer coding has
+   been applied, as described in <xref target="header.transfer-encoding"/>.
 </t>
 <sourcecode type="abnf7230"><iref primary="true" item="Grammar" subitem="message-body"/>
   <x:ref>message-body</x:ref> = *OCTET
 </sourcecode>
 <t>
-   The rules for when a message body is allowed in a message differ for
-   requests and responses.
+   The rules for determining when a message body is present in a HTTP/1.1
```
MikeBishop:
```suggestion
   The rules for determining when a message body is present in an HTTP/1.1
```

PR #254, opened 2019-09-04
Alternative Services (Mnot 143)

### LINKED ISSUE #143: Alternative Services
We need to figure out how Alt-Svc affects Semantics.

mnot:
Fixes #143


reschke:
But then the reference to RFC 6454 is informative...

royfielding:
> But then the reference to RFC 6454 is informative...

I meant that Mark's reference is defining origin as RFC6454. That's normative and would belong in the normative section. If it is intended to be an informative reference, it has to be presented as an example of how HTTP's notion of authority and origin server impacts HTML's notion of the origin for a given scripting context. They are not the same thing.

The tail doesn't wag the dog (references are defined as norm/inform based on the way that they are used, not the appendix in which they appear).

Issue #28, opened 2017-06-20
Describe Vary operation more clearly


mnot:
This confuses many people.

mnot:
> ## 2. Overview of Cache Operation
>
> The primary cache key consists of the request method and target URI. However, since HTTP caches in common use today are typically limited to caching responses to GET, many caches simply decline other methods and use only the URI as the primary cache key.
>
> If a request target is subject to content negotiation, its cache entry might consist of multiple stored responses, each differentiated by a secondary key for the values of the original request's selecting header fields (Section 4.1).

... but the `cache key` terms aren't used anywhere else except 4.3.4. It would be good to use this consistently in 3, 4 and especially 4.1.

mnot:
OTOH, since POST can be reused for GET, the definition of "cache key" is misleading. Perhaps the best thing to do would be to remove these terms completely.

mnot:
It would also be good to note that in practice, some implementations "flap" between having Vary and none; this should be clarified and mitigated.

mnot:
I think this has improved enough to close the issue; if anyone has a suggestion for further improvement, please comment.

PR #260, opened 2019-10-01
Describe Vary operation more clearly (Mnot 28)

mnot:
For #28

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -289,17 +289,6 @@
    reusing a stored response inappropriately, rather than mandating that
    caches always store and reuse particular responses.
 </t>
-<t>
-   Caches store HTTP responses corresponding to prior requests that used the
-   same cache key. Most commonly, this is a successful result of a retrieval
-   request: i.e., a <x:ref>200 (OK)</x:ref> response to a GET request, which
-   contains a representation of the resource identified by the request target
-   (<xref target="GET"/>). However, it is also possible to store 
-   redirects, negative results (e.g., <x:ref>404 (Not Found)</x:ref>),
-   incomplete results (e.g., <x:ref>206 (Partial Content)</x:ref>), and
-   responses to methods other than GET if the method's definition allows such
-   caching and defines something suitable for use as a cache key.
-</t>
 <iref item="cache key" />
 <t>
    The base <x:dfn>cache key</x:dfn> consists of the request method and target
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -558,6 +563,13 @@
    response (as determined by the <x:ref>Date</x:ref> header field) is used, as
    per <xref target="constructing.responses.from.caches"/>.
 </t>
+<t>
+   Note that in practice, some resources might omit appropriate Vary header
```
reschke: please stick to "header field" until we decide to change that...

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -558,6 +563,13 @@
...
+   values on some representations, or otherwise have Vary headers whose values
+   aren't stable. Caches &MAY; apply heuristics to identify such resources and
+   act appropriately (typically, applying the most recent Vary header value to
+   all responses that share the same target URI).
+</t>
```
reschke: FWIW, isn't that sort-of required? It would be the result of the server changing Vary dimensions, in which case the latest would definitvely the right one to consider.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -291,29 +290,35 @@
    caches always store and reuse particular responses.
 </t>
 <t>
-   Each <x:dfn>cache entry</x:dfn> consists of a cache key and one or more
-   HTTP responses corresponding to prior requests that used the same key. The
-   most common form of cache entry is a successful result of a retrieval
+   Caches store HTTP responses corresponding to prior requests that used the
+   same cache key. Most commonly, this is a successful result of a retrieval
```
royfielding: I find this more confusing than the original. How about if we delete the first sentence and move the rest of this paragraph below the discussion of cache key?

draft-ietf-httpbis-cache-latest.xml
royfielding: Yeah, I don't see the point of adding this. In general, the origin server chooses what dimensions of variance it wants a cache to care about, and anything not mentioned is assumed to be "okay". The intent is that the most recent acceptable fresh response will be reused by the cache, followed by most recent acceptable stale response. Further heuristics are always possible, but that isn't an interop requirement.

mnot: Many many servers are incorrectly implemented, and oscillate between different vary values, e.g., omitting them on the "default" representation. If a cache only pays attention to the latest or the values on the cached responses in isolation, it will behave incorrectly. Yes, this is the server's fault, but it seems reasonable to warn implementers about this widely experienced condition.

draft-ietf-httpbis-cache-latest.xml
mnot: done

draft-ietf-httpbis-cache-latest.xml
mnot: Text revised, PTAL

PR #266, opened 2019-11-20
Define origin and authority for http and https URIs

### LINKED ISSUE #237: Implications of Establishing Authority
From #194:
 
> Rewrite establishing authority to make the implications on clients and servers more clear, and referencing it from appropriate places (e.g., the Host header definition).

royfielding:
Fixes #237 and #194 with the hope of being useful for quicwg/base-drafts#253

I suspect that we also need to move "Initiating HTTP over TLS" down to the section on Routing, but that can be done separately.


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -788,33 +768,53 @@ Content-Type: text/plain
...
-</t>
-<t>
-   All of the requirements listed above for the "http" scheme are also
-   requirements for the "https" scheme, except that TCP port 443 is the
-   default if the port subcomponent is empty or not given,
-   and the user agent &MUST; ensure that its connection to the origin
-   server is secured through the use of strong encryption, end-to-end,
-   prior to sending the first HTTP request.
+   identifiers within a hierarchical namespace governed by a potential origin
+   server capable of establishing a secure and verifiable TLS connection on
```
martinthomson: I think that we need to be more precise about "secure and verifiable" here.  Using terms of art, you would say "a TLS connection that is authenticated on behalf of the identified authority and provides confidentiality and integrity protection (`<xref...`)."

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -788,33 +768,53 @@ Content-Type: text/plain
...
+   identifier to an IP address, establishing a TCP connection to that address
+   on the indicated port, securing the connection end-to-end by successfully
+   initiating TLS over TCP, and sending an HTTP request message to the server
+   over that secured connection containing the URI's identifying data
+   (<xref target="http.message"/>).
+</t>
+<t>
+   The client &MUST; ensure that its connection to the origin server is
+   secured through the use of strong encryption, end-to-end, prior to sending
+   the first HTTP request.
```
martinthomson: "secured" again is a little vague, though you do define it, it's not enough to use encryption.  Maybe instead say that the client `&MUST;` ensure that all requests it sends and all responses it accepts are only sent to a server that it authenticated as being authoritative and that those messages have confidentiality and integrity protection.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+</t>
+<sourcecode>
+   { "https", "example.com", "443"}
+</sourcecode>
+<t>
+   which can also be described as the normalized URI prefix with port always
+   present:
+</t>
+<sourcecode>
+   https://example.com:443
```
martinthomson: Is the intent here to supplant RFC 6454 rather than reference it?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+   within that namespace are mapped to resources. In turn, how the origin
+   responds to valid requests, consistently over time, determines the
+   semantics that users will associate with a URI, and the usefulness of
+   those semantics is what ultimately transforms these mechanisms into a
+   "resource" for users to reference and access in the future.
+</t>
+<t>
+   The general term "authority" is also used as an abstract reference to
+   whatever entity controls a given resource. A single authority might control
+   only parts of an origin, an entire origin, or many different origins.
```
martinthomson: Though a single logical authority might control a subset of the resources attributed to an origin, or even resources that span multiple origins, I don't know that this is a helpful statement to make here.  You can recognize that "authority" is a concept that HTTP uses both in this more abstract sense, but the protocol only concretely cares about authority in a very specific sense.  That is, the concrete definition of authority in HTTP is based on the origin of the resource alone.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+   origin server verified to be under the same authority.
+</t>
+<t>
+   What the client is relying on for "https" authority is that there exists
+   some chain of trust (possibly configured by exception) that verifies the
+   certificate provided. Hence, the authority is defined only by the host,
+   the port impacts which certificate is chosen (assuming that certificate
+   matches the host identifier), and the client trusts that match according
+   to the secure handshake and certificate validation described in
+   <xref target="https.initiation"/>.
```
martinthomson: I'm having a really hard time parsing this last sentence here.  Are you trying to say that the chain of trust is conveyed from some trust anchor (which is usually pre-arranged or configured), through a chain of certificates [citation here to RFC 5280 perhaps?] to a final certificate that binds a private key to host name for the origin.  The secure handshake and certificate validation in `<xref target="https.initiation"/>` describe how that certificate can be used to authorize a secure connection.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+</t>
+<t>
+   Note that the "https" scheme does not rely on TCP and the port number for
+   establishing authority, even though it provides for use of HTTP over TLS
+   over TCP via that port as the default mechanism for obtaining a
+   certificate that can be used for verifying authority and providing access
```
martinthomson: You don't really use TCP and the port number to obtain a certificate, but the specification defines a default mechanism using TCP, the port number, and TLS to establish an authenticated connection for the purposes of access to the origin's resources.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+   to the origin's resources. The peers establishing a secure connection
+   cannot trust transport parameters outside the secured channel, but the
```
martinthomson: > The protection offered by the secure channel does not typically provide integrity protection for the port number, but the value serves to distinguish between different origin namespaces on the same host.  Though it is potentially subject to modification, the port identifier is passed as part of the fields that identify a target resource.

I wanted to say `Host:` here, but h2 uses `:authority` by preference, so...

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+   client will make a DNS query and see that it contains the same server IP
+   address. A server sending the ORIGIN frame removes this restriction
+   <xref target="RFC8336"/>.
+</t>
+<t>
+   In addition to the client's verification, the origin server is responsible
+   for verifying that that requests it receives over a connection correspond
+   to resources for which it actually wants to be the origin. If a network
+   attacker causes connections for port N to be received at port Q, checking
+   the Host header field is necessary to ensure that the attacker can't cause
```
martinthomson: ..., the `:authority` pseudo-header field, (or the port field from a request in absolute form?).

h2 recommends use of `:authority` over `Host`.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+   default port for that scheme is used. For example, the URI
+</t>
+<sourcecode>
+   https://Example.Com/happy.js
+</sourcecode>
+<t>
+   would have the origin
+</t>
+<sourcecode>
+   { "https", "example.com", "443"}
```
martinthomson:
```suggestion
   { "https", "example.com", "443" }
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+   the listener (if any) on the indicated TCP port at that IP address.
+   If host is a registered name, the registered name is an indirect identifier
+   for use with a name resolution service, such as DNS, to find an address for
+   an appropriate origin server.
+</t>
+<t>
+   Access to "http" services is not limited to the authority mechanism.
+   For example, the Alt-Svc header field <xref target="RFC7838"/> allows an
+   origin server to explicitly alias its own namespace to an alternative
+   service at a different origin. Access to "http" identified resources might
```
MikeBishop: I think this is not quite correct.  Alt-Svc permits the origin server to identify other endpoints which are also authoritative for that origin.  There's not necessarily any aliasing to a different origin going on at those alternative servers -- they receive the origin's hostname as part of the request and might well use it to find the appropriate resource.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+   also be provided by other protocols outside the scope of this document.
+</t>
+<t>
+   The "https" scheme, in contrast, associates authority based on the ability
```
MikeBishop: It's unclear whether "in contrast" means versus the use of Alt-Svc, or versus the "http" scheme.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+   of a server to use the private key associated with a certificate that
+   the client considers to be trustworthy for the identified host.
+   If a server presents a certificate that verifiably applies to the host,
+   along with proof that it controls the private key, then a client will
+   accept a secured connection to that server as being authoritative for all
+   origins with the same host. However, each origin remains distinct (from
```
MikeBishop: > a client will accept a secured connection to that server as being authoritative for all origins with the same host.

This is a very expansive statement, and I think it's broader than what you intended to say here.  I've connected over TLS to `www.example.com` on port 443 and verified the certificate.  Is that connection now valid to make requests for:
- `https://www.example.com/`?  Clearly.
- `https://www.example.com:8443/`?  I think this is the expansion from current practice that you intended here.
- `http://www.example.com/`?  This matches "all origins with the same host," but we decided in RFC 8164 that we wouldn't do this without explicit opt-in.
- `ftp://www.example.coml`?  This matches "all origins with the same host," but I think it's pretty clearly out-of-scope.

At the same time, we might want to make a similar statement about other hostnames in the certificate.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+   origins with a different scheme or port) unless explicitly aliased by an
+   origin server verified to be under the same authority.
```
MikeBishop:
```suggestion
   origin server authoritative for that origin.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+</t>
+<t>
+   What the client is relying on for "https" authority is that there exists
```
MikeBishop:
```suggestion
   Authority in the "https" scheme relies on
```

draft-ietf-httpbis-semantics-latest.xml
MikeBishop: This paragraph is tough to parse.  Also, the fact that a server might choose to use different certificates on different ports is both an implementation detail and not really necessary.  How about:
> Hence, authority is defined only by the host (assuming that the certificate matches the host identifier) and scheme.  The client validates the certificate and handshake as described in <xref target="https.initiation"/>.  The port partitions multiple namespaces controlled by the server and serves as a convenience when attempting to establish a connection.

Note that this is similarly expansive, reinforcing that a request for `https://www.example.com:8443` send on a TCP connection to port 443 is totally legitimate.

**P.S.**  I see that you get into this further down, so some of this might be superfluous instead.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +848,152 @@ Content-Type: text/plain
...
+   cannot trust transport parameters outside the secured channel, but the
+   port number remains useful (beyond the default mechanism) as an identifier
+   to distinguish different origin namespaces on the same host. The port
+   identifier is therefore applicable even when the protocols being used for
+   access do not include TCP.
+</t>
+<t>
+   In HTTP/1.1 and earlier, the only URIs for which the client will assign
+   authority are those that contain a TLS server_name extension matching the
+   origin's host.
```
MikeBishop: I know what you're trying to say, but tying this to the SNI doesn't work.  It's legitimate for a client not to send that extension.

draft-ietf-httpbis-semantics-latest.xml
royfielding: done

draft-ietf-httpbis-semantics-latest.xml
royfielding: I tried several variations and finally had to go with an explicit definition for "secured" in this context. Trying to detail it three times in the same section was hopeless.

draft-ietf-httpbis-semantics-latest.xml
royfielding: Yep, thanks for the careful review. I can delete the paragraph and just use entity below.

draft-ietf-httpbis-semantics-latest.xml
royfielding: My intent is to supply the normative bits specifically for HTTP origin, which a later update to 6454 might use as part of its larger description of web origins for HTML scripting. I move the informational reference to 6454 up to this section. It's a little confusing because the web origin concept is based on the HTTP origin server definitions but applied to many other references as well.

draft-ietf-httpbis-semantics-latest.xml
royfielding: Umm, okay, I guess that's what I meant to say. Thanks for clarifying and providing a reference.

draft-ietf-httpbis-semantics-latest.xml
royfielding: I think this is missing the point I was trying to make, so I have redrafted it.

draft-ietf-httpbis-semantics-latest.xml
royfielding: Right, I have included parts of that in the new description.

draft-ietf-httpbis-semantics-latest.xml
royfielding: Changed to "effective request URI"

draft-ietf-httpbis-semantics-latest.xml
royfielding: fixed

draft-ietf-httpbis-semantics-latest.xml
royfielding: You've lost me.  I thought this (from Martin) was talking about the TLS certificate matching the same name as the origin host. What is the right way to say that?

draft-ietf-httpbis-semantics-latest.xml
royfielding: Yes, I think it was overtaken by events.

draft-ietf-httpbis-semantics-latest.xml
royfielding: Tweaked slightly, but I am deliberately trying to describe the client mechanism here, whereas the other paragraphs covered a neutral perspective.

draft-ietf-httpbis-semantics-latest.xml
royfielding: fixed

draft-ietf-httpbis-semantics-latest.xml
royfielding: fixed

draft-ietf-httpbis-semantics-latest.xml
royfielding: fixed

draft-ietf-httpbis-semantics-latest.xml
royfielding: Good point. I changed it to all origins with the same scheme and host.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +851,151 @@ Content-Type: text/plain
...
+   https://example.com:443
+</artwork>
+<t>
+   The notion of an origin for any given URI is central to the semantics of
+   HTTP. Each origin defines its own namespace and controls how identifiers
+   within that namespace are mapped to resources. In turn, how the origin
+   responds to valid requests, consistently over time, determines the
+   semantics that users will associate with a URI, and the usefulness of
+   those semantics is what ultimately transforms these mechanisms into a
+   "resource" for users to reference and access in the future.
```
martinthomson: What do we mean by "users" in this context?  Do you refer to the user behind the user agent?  (In which case, this is fine and you don't need to make any changes.)

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +851,151 @@ Content-Type: text/plain
...
+<t>
+   Although HTTP is independent of the transport protocol, the "http" scheme
+   is specific to associating authority with whomever controls the origin
+   server listening for TCP connections on the indicated port of whatever
+   host is identified within the authority component. This is a very weak
+   sense of authority because it depends on both client-specific name
+   resolution mechanisms and communication that might not be secured from
+   man-in-the-middle attacks. Nevertheless, it is a sufficient minimum for
+   binding "http" identifiers to an origin server for consistent resolution
+   within a trusted environment.
```
martinthomson: I don't like this concept of "trusted environment".  I would rather not defend the "http" scheme at all, simply recognizing the inherent limitations being enough for this document.  Saying how it is should be enough; value judgments can be made by others.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +851,151 @@ Content-Type: text/plain
...
+<t>
+   Note that the presence of a URI does not imply that there is always an
+   HTTP server at the identified origin listening for connections on
+   the identified host and port. Anyone can mint a URI. What the origin
+   determines is who has the right to respond authoritatively to requests that
+   target the identified resource. The delegated nature of registered names
+   and IP addresses creates a federated namespace whether or not an HTTP
+   server is present.
+</t>
+<t>
```
martinthomson: It seems to me that this might be a good point to have a section that is dedicated to discussion of "http" origins.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +851,151 @@ Content-Type: text/plain
...
+   an appropriate origin server.
+</t>
+<t>
+   Access to "http" services is not limited to the default mechanism.
+   For example, the Alt-Svc header field <xref target="RFC7838"/> allows an
+   origin server to identify other services that are also authoritative for
+   that origin. Access to "http" identified resources might also be provided
+   by protocols outside the scope of this document.
+</t>
+<t>
```
martinthomson: Similar to above, a section for "https" origin-specific material here.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -848,6 +851,151 @@ Content-Type: text/plain
...
+   Unlike "http", the "https" scheme associates authority based on the ability
+   of a server to use the private key associated with a certificate that
+   the client considers to be trustworthy for the identified host.
+   If a server presents a certificate that verifiably applies to the host,
+   along with proof that it controls the private key, then a client will
```
martinthomson: You use "the private key" in the singular here, but it is common practice to have many such keys, if only over time.  It might be worth acknowledging that fact.

draft-ietf-httpbis-semantics-latest.xml
martinthomson: > In HTTP/1.1 and earlier, the only URIs for which a client will attribute authority to a server are those for which a TLS connection was specifically established toward the origin's host. Only the characteristics of the connection establishment and certificate are used.  For a host that is a domain name, that name is included in the TLS server_name extension <xref target="RFC6066"/> and &MUST; also appear as either the CN field of the certificate subject or as a dNSName in the subjectAltName field of the certificate (see <xref target="section defining this..."/>)  A host that is an IP address &MUST; appear in the subjectAltName of the certificate.

draft-ietf-httpbis-semantics-latest.xml
MikeBishop: I think "corresponding private key" (i.e. the key that matches the certificate) would be okay.  The server has many certificates for the origin name, each of which has exactly one private key, no?

draft-ietf-httpbis-semantics-latest.xml
royfielding: Thanks.  I've tweaked it a bit to make these requirements on the client and added a ref to RFC6125.

draft-ietf-httpbis-semantics-latest.xml
royfielding: It is important to remind people that HTTP is used in a lot of contexts, including things like non-TLS secured networks and localhost sockets. This is the easy way to phrase it.

draft-ietf-httpbis-semantics-latest.xml
royfielding: yes, it was running long. I am also going to look at moving the whole thing to Routing.

draft-ietf-httpbis-semantics-latest.xml
royfielding: Yep, fixed.

royfielding:
This is now ready to merge, IMO. There is more work to do after merging, specifically taking the old RFC2818 description of initiating HTTP over TLS out and moving just the 1.1-specific bits to Messaging, but I don't want to do that in the same PR.

draft-ietf-httpbis-semantics-latest.xml
martinthomson: Easy doesn't mean right though.  The challenge here is that trusted environment implies that it is possible to trust an environment.  I would say instead:

> In some contexts, clients might choose to accept the risk of attack, or choose to rely on other mechanisms to mitigate those risks.

What I think this needs to capture is that this risk remains even in those places.  I appreciate that this is a value judgment, but I think that it's an important one.

martinthomson:
I agree with Roy.  This is big enough that merging and iterating is the right path.

Issue #161, opened 2018-10-16
Storing responses with Authentication


mnot:
[3.2. Storing Responses to Authenticated Requests](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#caching.authenticated.responses) lists some directives that allow responses with Authentication to be stored, but some of those directives' definitions do not mention this.

mpdude:
I think this should also be added for `public`; see #269.

royfielding:
I think this issue needs to be reopened and the specification text reverted to something more like what was in RFC2616. That spec specifically requires public be present to allow responses to a request with Authorization to be stored by a shared cache, regardless of must-revalidate or proxy-revalidate. This is not implied by those other directives.

PR #269, opened 2019-12-03
Explicitly state that "public" allows responses with Authentication to be stored

mpdude:
In #161, the suggestion was made to explicitly note when a directive allows the response to be cached, even with Authorization being used.

According to section `caching.authenticated.responses`, `public` also has this effect, so I added a note there.

Additionally, in `cache-response-directive.s-maxage`, I tried to make it clearer that `s-maxage` overrides Authorization as well, due to the implication of proxy-revalidate and thus must-revalidate.


mpdude:
Oh, I missed that there is already a reference to `Authorization` in `public`. I'll see if I can get this to be more in line with the other sections.

royfielding:
This PR has been overtaken by #161

Issue #245, opened 2019-09-02
"final" status code/response should be defined in semantics


reschke:
...that will also allow [Caching] to move a reference.

royfielding:
Section 9.2 (Informational status codes) is already describing these as "final response" and "interim response", and I think we use interim elsewhere for this type of requirement. I would prefer that we stick with the interim/final terminology instead of non-final/final.

reschke:
@royfielding - PR adjusted accordingly.

mnot:
Can this be closed?

reschke:
Yes.

PR #275, opened 2020-01-19
In Semantics, define "final" and "interim" status codes, and use that in reference from Caching

reschke:
…at in reference from Caching (#245)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6677,6 +6678,17 @@ Expect: 100-continue
     valid request
    </li>
 </ul>
+<t anchor="final-non-final">
+  <x:anchor-alias value="final"/>
+  <x:anchor-alias value="non-final"/>
+  <iref item="Status Codes" subitem="Final"/>
+  <iref item="Status Codes" subitem="Non-Final"/>
+  <iref item="Status Codes" subitem="Informational"/>
+  A single request can have multiple associated responses: several &OPTIONAL;
```
mnot: "several OPTIONAL" is weird. I would prefer "zero to many"

PR #281, opened 2020-02-02
Be more correct about what was deprecated by RFC 3986 (and also talk about https URIs) (#278)

reschke:
(also adds two paragraph breaks)

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.html
+++ b/draft-ietf-httpbis-semantics-latest.html
@@ -1655,14 +1655,23 @@ Content-Type: text/plain
             <section id="http.userinfo">
                <h4 id="rfc.section.2.5.4"><a href="#rfc.section.2.5.4">2.5.4.</a>&nbsp;<a href="#http.userinfo">Deprecated userinfo</a></h4>
                <div id="rfc.section.2.5.4.p.1">
-                  <p>The URI generic syntax for authority also includes a deprecated userinfo subcomponent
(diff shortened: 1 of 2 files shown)


vanrein:
Thanks!

To be completely formally correct, the term from RFC3986 should not be quoted as "user authentication information in the URI" but as "user authority information in the URI".  Is that correction possible?

reschke:
WFM. @royfielding ?

royfielding:
No, it isn't correct to say that it is user authority information. The purpose of userinfo in a server-based URI format (this://server/) is to supply canned authentication data for use if the server requests such information. It is strictly for use on the client-side and is not directly passed in HTTP unless that (Basic) happens to be the form of challenge requested by the server.

vanrein:
> No, it isn't correct to say that it is user authority information.
> The purpose of userinfo in a server-based URI format (this://server/) is to supply canned authentication data

That would be an override by HTTP of the definition in RFC3986 then?

(Not saying that is illegal; HTTP predates this URI spec.)

vanrein:
The [proposal I am working on](https://tools.ietf.org/html/draft-vanrein-http-unauth-user-05) generalises the constrained use for authentication by allowing the entire URI to be a server-side spec and the client to derive a hint about it as a possible authentication user.

This can end any such difference in interpretation between HTTP/bis and the URI format.

I do know this is a broadening of the user concept in HTTP; I also believe that this is not without a need.

royfielding:
RFC3986 says:
```
   The userinfo subcomponent may consist of a user name and, optionally,
   scheme-specific information about how to gain authorization to access
   the resource.
```

Specifically, the userinfo supplies the information necessary to gain access to the identified resource via a process of authentication. That process is selected and defined by the origin, not by the URI, and the username is not an identifier of authority.

Note that userinfo for http and https URLs has never been allowed, going all the way back to RFC1738. Some applications do use that syntax, internally and in a way that is not subject to the Internet standards process, so we warn about it in general.

PR #282, opened 2020-02-02
In considerations for new header fields, discourage ABNF mixing lists with other constructs (#7)

reschke:


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1894,6 +1894,11 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
     as the comma can only appear inside quoted strings;
     bad example: "Location", as a comma can occur inside a URI).</t>
   </li>
+  <li>
+    <t>When it does use list syntax, this needs to happen at the top-level; that is,
```
mnot: "needs to" is a bit weird here -- maybe "is encouraged to"? Also, describe why it's bad?

royfielding:
That's because mixing a single value or a list doesn't impact anyone other than the spec writers. I suggest we break this long list of mangled bullets into separate subsections that can describe why they should be avoided.

reschke:
Ok, so let's separate the reorg from this particular ticket...

But then:

>That's because mixing a single value or a list doesn't impact anyone other than the spec writers

Isn't that true for all what we say here?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1735,28 +1735,37 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
+      field's definition. A robust value format that distinguishes
+      comma-separated list values from a singular value for the field will
+      allow recipients to more easily discover an improperly repeated field.
+      In contrast, a single value definition that allows an unquoted comma
+      is difficult to distinguish from multiple field lines that have been
+      combined. Hence, "Content-Type" is a good example because a valid comma
+      would only appear inside a quoted string, whereas "Location" is a
+      bad example because a comma might occur inside a valid URI.</t></li>
+    <li><t>When a field definition does use the list syntax, avoid using a
+      list or unquoted comma within the member values, since that makes it
```
mnot: What's the significance of "list or" here?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1735,28 +1735,37 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
 <section title="Considerations for New HTTP Fields" anchor="considerations.for.new.header.fields">
 
 <t>
-   See <xref target="field.names"/> for a general requirements for field names,
+   See <xref target="field.names"/> for general requirements of field names,
```
mnot: "regarding field names" maybe?

reschke:
I think we should close this PR and not merge it.

- it's not anymore about issue #7
- I don't think adding a sub list improves things

Let's open separate tickets for the other changes, and once we're done with that, I'll have another look at what #7 was about.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1735,28 +1735,37 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
+      is difficult to distinguish from multiple field lines that have been
+      combined. Hence, "Content-Type" is a good example because a valid comma
+      would only appear inside a quoted string, whereas "Location" is a
+      bad example because a comma might occur inside a valid URI.</t></li>
+    <li><t>When a field definition does use the list syntax, avoid using a
+      list or unquoted comma within the member values, since that makes it
+      very difficult for a recipient to parse the field unambiguously.
+      A bad example for that is the list of credentials in WWW-Authenticate
+      (<xref target="header.www-authenticate"/>).</t></li>
+    </ul>
```
reschke, line 1768: Not convinced. This suggests that any use of commas inside field values is bad, unless when it happens in quoted strings. Do we really want to go there? What about etags and URIs?

WWW-Authenticate is indeed tricky to parse, but that's because it uses COMMA as *delimiter* in two different layers (separating challenges, but also separating auth-params inside challenges).

Issue #111, opened 2018-06-29
Header terminology


mnot:
From #59:

There's some fuzziness about terminology around the differences between:

* a header's complete value
* a header line
* an individual value in a list-based header field

This could do with some improvement.

mnot:
Another thing to consider is whether we should condone/use  "header" instead of / in addition to "header field"; the former is extremely common use.

mnot:
What if we made an explicit split between the syntactic concept of header fields, and the abstract, post-aggregation concept of the headers that most people are used to working with?

That way, we'd have a clean split; these would be syntactic:

* header field - one line
* header field-name - one line's name
* header field-value - one line's value

... and these would refer to the synthetic value after coalescence:

* header - coalesced header 
* header name - coalesced header name
* header value - coalesced header value

Ideally, we'd adjust "header section" to "header field section" to reinforce that it's syntactic.

reschke:
+1 in general, but I don't believe that calling one thing "header field" and the other "header" is going to work - right now people use it interchangeably. Better find a new attribute, such as "aggregated", "coalesced" or "combined":

mnot:
I suspect that we call it a "header field" and everyone else calls it a "header."

Perhaps if we use only "field" to denote the syntax, and "header" to denote the post-processed thing?

reschke:
That again would change an existing definition and would be IMHO confusing. I still prefer to be more clear by adding an attribute.

mnot:
I agree we need to be clear. However, something like "coalesced header field value" is too verbose for common usage. Ideally, we'd be using two words, not four.

Let's look at it from a slightly different perspective. For practitioners and other specifications, the most common term by far is going to be a term to refer to:

* For headers that don't use the #list convention, its value - while still being able to talk about "extra values that were appended due to additional headers being present"
* For headers that do use the #list convention, a member of the list

If we can come up with a brief, descriptive term for that -- e.g., "header item" -- we can build other things around it, I think.

annevk:
FWIW, Fetch has this as "combined value".

royfielding:
I agree we need to clean this up. I would prefer not to paint that shed in a meeting. Can we just point people to this issue and not spend any time on it? Painting it here is fine.

mcmanus:
in bkk: really discussed header-field vs header nomenclature. Sense of room favored loosening to allow just header but it wasn't unanimous particular wrt email consistency.

mnot:
Discussed by the editors. Proposal:

[...]

PR #283, opened 2020-02-02
Mnot 111

mnot:
Work in progress for #111


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -771,12 +771,12 @@ https://www.example.org
 <section title="Header Field Syntax" anchor="header.field.syntax">
```
royfielding: Field Syntax

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -863,37 +863,37 @@ https://www.example.org
   </tbody>
 </table>
 
-<section title="Header Field Parsing" anchor="field.parsing">
+<section title="Header Field Line Parsing" anchor="field.parsing">
```
royfielding: Field Line Parsing (and I think we can remove "header" from this and the section above)

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -2483,12 +2483,12 @@ Upgrade: websocket
    "IETF (iesg@ietf.org) - Internet Engineering Task Force".
 </t>
 
-<section title="Header Field Registration" anchor="header.field.registration">
+<section title="Field Registration" anchor="header.field.registration">
```
royfielding: should this be "Field Name Registration"?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1093,371 +1093,185 @@ Content-Type: text/plain
 </section>
 </section>
 
-<section title="Header Fields" anchor="header.fields">
+<section title="Header and Trailer Fields" anchor="header.fields">
+   <t>
+      HTTP messages use key/value pairs to convey data about the
```
royfielding: These paragraphs are okay, but I think it would work better to describe them commonly first, as fields within a potential header section or trailer section, and then "In general, we refer to these as HTTP message fields, or just "fields".

Note the "Every message has two ..." is a bit strong. HTTP messages have a (potentially empty) header section and potentially one or more trailer sections.

Issue #24, opened 2017-04-23
Via header: host ABNF could allow ","


reschke:
Reported by quae@daurnimator.com in https://lists.w3.org/Archives/Public/ietf-http-wg/2016OctDec/0527.html:

I think I found a bug in the specification of the Via header as given
in RFC 7230

>From RFC 7230: Via = 1#( received-protocol RWS received-by [ RWS comment ] )
where 1# is a special syntax that means "comma seperated list, at
least one element"

>From RFC 7230: received-by = ( uri-host [ ":" port ] ) / pseudonym
>From RFC 7230: uri-host = <host, see [RFC3986], Section 3.2.2>
>From RFC 3986: host = IP-literal / IPv4address / reg-name
>From RFC 3986: reg-name = *( unreserved / pct-encoded / sub-delims )
>From RFC 3986: sub-delims = "!" / "$" / "&" / "'" / "(" / ")" / "*" /
"+" / "," / ";" / "="

notice "," there in sub-delims; this means that comma is a valid
character in a host.
and hence, that using a comma to terminate a host makes no sense

e.g.
Via: 1.0 fred, 1.1 p.example.net
'fred,' is a valid uri-host
In this case, I think we might be saved by the fact that the rest of
the line doesn't match, so 'fred' ends up being a pseudonym rather
than a uri-host.

However, I believe that there might be corner cases not backed up by
this fallback.

mnot:
@royfielding could you enlighten us as to why 3986 allows such a wide range of characters in reg-name?

mnot:
Discussed in Basel; inclination is to remove `,` from `uri-host`.

reschke:
Restrict the grammar in Via.

reschke:
Proposal:

- simplify "received-by" not to distiguish hosts and pseudonyms, just define the allowed characters
- remove pseudonym ABNF

receiced-by would include...?

1) Most inclusive: VCHAR minus ","
2) Least inclusive: token plus all the characters we need for host names, excluding ","

reschke:
The difference between token and VCHAR is:

    DQUOTE / "(" / ")" / "," / "/" / ":" / ";" / "<" / "=" / ">" / "?" / "@" / "[" / "\" / "]" / "{" / "}"

Which ones do we need to allow?

reschke:
I believe we need to allow the following characters:

[...]

PR #284, opened 2020-02-02
Change Via to use via-host, disallowing comma (#24)

reschke:


reschke:
Roy, could you please review the changed PR?

royfielding:
I am thinking that we should reduce it to unreserved: ALPHA / DIGIT / "-" / "." / "_" / "~"

reschke:
Hmmm, that's even more restrictive than token, which was used for pseudonym.

reschke:
WFM. @mnot ?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -11868,6 +11868,15 @@ Content-Encoding: gzip
    <xref target="RFC4918" x:fmt="of" x:sec="11.2"/>) because of its general
    applicability (<xref target="status.422"/>).
 </t>
+<t>
+   The grammar definition for the Via field's "received-by" was unfortunately
```
mnot: "unfortunately" reads a little weird here. Not objecting strongly, but this para could be massaged a bit.

PR #288, opened 2020-02-03
Change requirements for handling different forms of cache directives

### LINKED ISSUE #128: Quoted cache-control directives
In my testing, many caches do not recognise `Cache-Control: max-age="3600"` as a valid CC directive.

Since this isn't interoperable, we should consider changing the spec so that the interoperable case is documented.

mnot:
Fixes #128


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -2275,6 +2273,15 @@
...
    (<xref target="header.warning"/>)
 </t>
+<t>
+   Some cache directives defined by this specification now have stronger
+   prohibitions against generating the quoted form of their values, since
+   this has been found to create interoperability problems. Consumers of
+   extension cache directives are no longer required to accept both token and
+   quoted-string forms.
+   (<xref target="header.cache-control"/>)
+</t>
```
reschke: ...but they still need to properly parse them for unknown extensions.

Issue #273, opened 2020-01-09
message field registry migration needs to consider status "reserved"


reschke:
We have at least one entry in status "reserved" (for "close"). We need to define how it will show up in the new registry.

reschke:
Proposal 1:

add a status "reserved" (and define the procedure for reserving field names)

Proposal 2:

move "close" to "permanent", but add a comment column in the registry and note what is special about that field name over there.

mnot:
Either is fine with me, but I lean towards (2).

reschke:
Asked mailing list in <https://lists.w3.org/Archives/Public/ietf-http-wg/2020JanMar/thread.html#msg103>.

PR #293, opened 2020-02-03
In instructions for IANA header field registry migration, consider "reserved" as well (#273)

reschke:


mnot:
It might be worth flying this past the list to see if anyone with more IANA lore wants to weigh in.

reschke:
Commit, then ask mailing list?

PR #295, opened 2020-02-03
Give guidance on interoperable field names

### LINKED ISSUE #30: Field-name syntax
Header field-names are defined as tokens. This is an extremely permissive syntax, including characters that will cause confusion and likely break some senders/recipients.

Most of the special characters allowed are not in the registry or seen "in the wild." Some research would be good to substantiate their use, but a starting point might be:

> "-" / "_" / "." / "+" / DIGIT / ALPHA

There are a number of strategies we could take to the transition:

1. Like `OWS` / `BWS`, mark some characters as "do not generate" but "should consume"

2. Disallow registration of header fields with those characters, and discourage their use in unregistered headers

3. If we have more confidence that they're not in use, just ignore headers containing those characters.

mnot:
Fixes #30

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -1278,7 +1278,7 @@ Content-Type: text/plain
    practice some implementations place limits on the characters they accept
    in field-names. To be interoperable, new field names &SHOULD; constrain
    themselves to alphanumeric characters, "-", "_", "." and ".", and &SHOULD;
-   begin with an alphabetic character.
+   begin with an alphanumeric character.
 </t>
 <t>
    Field names ought not be prefixed with "X-"; see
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1273,6 +1273,13 @@ Content-Type: text/plain
    a Description field, it might use "Foo-Desc"; "Description" is too generic,
    and "Foo-Description" is needlessly long.
 </t>
+<t>
+   While the field-name syntax is defined to allow any token character, in
+   practice some implementations place limits on the characters they accept
+   in field-names. To be interoperable, new field names &SHOULD; constrain
+   themselves to alphanumeric characters, "-", "_", "." and ".", and &SHOULD;
+   begin with an alphanumeric character.
+</t>
```
reschke: Maybe even disallow start with numeric?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1273,6 +1273,13 @@ Content-Type: text/plain
    a Description field, it might use "Foo-Desc"; "Description" is too generic,
    and "Foo-Description" is needlessly long.
 </t>
+<t>
+   While the field-name syntax is defined to allow any token character, in
+   practice some implementations place limits on the characters they accept
+   in field-names. To be interoperable, new field names &SHOULD; constrain
+   themselves to alphanumeric characters, "-", "_", "." and ".", and &SHOULD;
```
PiotrSikora: You included `"."` twice.

mnot: Fixed, thanks.

Issue #262, opened 2019-10-29
Allow "Etag" in trailer section


reschke:
See discussion in #16.

mnot:
I think this is uncontroversial, as long as the wording indicates it can appear, but recipients might ignore it there.

Any other headers in our specs that we want to include too? Last-Modified seems logical if we do this...

royfielding:
done

PR #296, opened 2020-02-03
Allow ETag in trailers (#262)

reschke:


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9331,6 +9331,11 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
    &MUST; mark the entity-tag as weak by prefixing its opaque value
    with "W/" (case-sensitive).
 </t>
+<t>
+   The Etag field can appear in the trailer section as well (see
+   (<xref target="trailer.fields"/>). Note however that recipients might
+   ignore it when it is not sent as header field.
```
mnot: as _a_ header field.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9331,6 +9331,11 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
    &MUST; mark the entity-tag as weak by prefixing its opaque value
    with "W/" (case-sensitive).
 </t>
+<t>
+   The Etag field can appear in the trailer section as well (see
+   (<xref target="trailer.fields"/>). Note however that recipients might
```
mnot: seems like a comma or two is missing

reschke: I do see an extra "(" that needs to go, but where do you miss a comma?

mnot: Note, however, that...

PR #299, opened 2020-02-03
Define semantics and acceptable replacement of WS

### LINKED ISSUE #53: Clarify semantics of OWS in header field values
From Julian Reschke:
"We probably should clarify
https://www.greenbytes.de/tech/webdav/rfc7230.html#rule.OWS

with statements about what kind of rewriting is allowed. And, for

https://www.greenbytes.de/tech/webdav/rfc7231.html#header.allow

we need to clarify that "MUST NOT modify" doesn't apply to the *WS rewriting we allow in general."

mnot:
Fixes #53


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -336,12 +336,22 @@
...
+   interpreting it or forwarding the message downstream.
+</t>
 <t>
    The BWS rule is used where the grammar allows optional whitespace only for
    historical reasons. A sender &MUST-NOT; generate BWS in messages.
    A recipient &MUST; parse for such bad whitespace and remove it before
    interpreting the protocol element.
 </t>
+<t>
+   BWS has the same semantics as an empty string. Any content known to be
```
royfielding: "BWS has the same semantics as an empty string" could be confused with an empty quoted-string.  How about "BWS has the same semantics as no whitespace", or more simply that "BWS has no semantics".

PR #302, opened 2020-02-04
A last fine point towards RFC 3986 correct text

vanrein:
A simple PR to address the still-open part of issue https://github.com/httpwg/http-core/pull/281

URIs relay authority information, not authentication.
The next paragraph is different; it mentions a use _as_ authentication.


royfielding:
I understand why you would like that to be the case, but it simply isn't true and that usage is not allowed by HTTP (because userinfo is not defined for http(s) URLs).

Issue #272, opened 2020-01-09
fix grammar for "Vary" header field


reschke:
it uses
   
    "*" / #vary

but "vary" is a token, and token allows "*" anyway. Also, see #7.

reschke:
In particular, "*" is a valid field name. We probably should reserve it in the registry.

annevk:
(CORS has an issue like that as well, including for a `*` method; see the note at https://fetch.spec.whatwg.org/#http-new-header-syntax.)

reschke:
Proposal:

    vary-by = "*" / field-name
    Vary = 1#vary-by

Then state that if "*" appears, no "real" field-name is allowed to appear as well (for backwards compat).

Do we need to discuss cases like

    Vary: *, *

?

royfielding:
I don't see why that is an improvement.

mnot:
I thought we were going to reserve `*` and define its meaning in prose...

mnot:
I.e., it should just be

> Vary = 1#field-name

reschke:
which implies registering "*" as reserved field name, right?

mnot:
Yes. That's #274

mnot:
Discussed; will create a PR to move ABNF and prose to list-based.

reschke:
Do we need to list that as change from 7230?

mnot:
Probably a good idea; will commit to master and link here.

PR #304, opened 2020-02-04
Clarify that any "*" as a member of Vary will disable caching

### LINKED ISSUE #286: Vary *
[Calculating Cache Keys with Vary](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#constructing.responses.from.caches) says:

> A Vary header field-value of "*" always fails to match.

I think this should be:

> A Vary header field value containing a member "*" always fails to match.

mnot:
Fixes #286


mnot:
That's #272

PR #307, opened 2020-02-04
make explicit requirements on origin server to use authority from abs…

### LINKED ISSUE #191: Mismatching absolute URI and Host header
[7230 5.4]() says:

> If the target URI includes an authority component, then a client MUST send a field-value for Host that is identical to that authority component, excluding any userinfo subcomponent and its "@" delimiter (Section 2.7.1).

However, the requirements upon a server receiving a request with a mismatching authority component and Host header are not clearly stated.

Below, it says:

> A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request message that lacks a Host header field and to any request message that contains more than one Host header field or a Host header field with an invalid field-value.

However, a mismatching field-value is not clearly labeled to be invalid. 

The waters are further muddied by [7230 section 5.5]():

> If the request-target is in absolute-form, the effective request URI is the same as the request-target.  Otherwise, the effective request URI is constructed as follows...

... which I've seen people interpret as "the absolute form takes precedence over the Host header."

I think we need to:

- [ ] clarify server behaviour when receiving mismatching absolute form and host header (i.e., tweaking the language around "invalid")
- [ ] modify the beginning of 5.5 to remove the impression outlined above

mnot:
…olute-form when available

Fixes #191

Note that lots of this text needs to be moved to messaging; I think we have another issue for that.


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2326,6 +2326,11 @@ Host: www.example.org
    &MUST; generate a new Host field value based on the received
    request-target rather than forward the received Host field value.
 </t>
+<t>
+   When an origin server receives a request with an absolute-form of
+   request-target, the origin server must ignore the received Host header
```
royfielding: needs to take into account absolute-form URIs that do not have an authority component, such as URNs and mailto and ...

royfielding: note that, in this case, the Host would be sent with empty value.

PR #309, opened 2020-02-04
recommend comma SP when combining field lines

### LINKED ISSUE #148: Exact separator used when combining header fields

mnot:
Fixes #148

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -1208,8 +1208,7 @@ Content-Type: text/plain
    A recipient &MAY; combine multiple field lines with the same field name
    into one field line, without changing the semantics of the message, by
    appending each subsequent field line value to the initial field line value
-   in order, separated by a comma and optional whitespace. For consistency, use a
-   single SP as the OWS.
+   in order, separated by a comma and optional whitespace. For consistency, use comma SP.
 </t>
 <t>
    The order in which field lines with the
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1208,7 +1208,11 @@ Content-Type: text/plain
    A recipient &MAY; combine multiple field lines with the same field name
    into one field line, without changing the semantics of the message, by
    appending each subsequent field line value to the initial field line value
-   in order, separated by a comma. The order in which field lines with the
+   in order, separated by a comma. For consistency, senders ought to combine
+   field lines with comma SP.
+</t>
```
reschke: Is it "senders"?

PR #315, opened 2020-02-06
connection options ought to be registered

### LINKED ISSUE #285: Reserving connection options
~~~
   When defining new connection options, specification authors ought to survey
   existing field names and ensure that the new connection option does
   not share the same name as an already deployed field.
   Defining a new connection option essentially reserves that potential
   field name for carrying additional information related to the
   connection option, since it would be unwise for senders to use
   that field name for anything else.
~~~

We should explicitly tell them to reserve it.

mnot:
Fixes #285


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1671,13 +1671,10 @@ https://www.example.org
...
-   not share the same name as an already deployed field.
-   Defining a new connection option essentially reserves that potential
-   field name for carrying additional information related to the
-   connection option, since it would be unwise for senders to use
-   that field name for anything else.
+   When defining new connection options, specification authors ought to
+   register the connection option as a field name in the Hypertext Transfer
+   Protocol (HTTP) Field Name Registry (<xref
+   target="field.name.registry"/>), to avoid collisions.
 </t>
```
reschke: 1) Is this "ought to" only?

2) "Register as field name" might be to vague; maybe "document it as reserved field name and register that definition"?

mnot: 1. We've used "ought to" in similar statements for other registries. It's not an interoperability requirement on implementations...

2. happy to make the change, but I don't see an improvement.

Issue #259, opened 2019-10-01
request target is h1-specific


mnot:
Semantics and caching contain many uses of "request target" and `request-target`. This is a term that's specific to h1's wire format; it should be specified in a way that is not version-specific.

I think that in almost every case, "effective request URI" is the term we want to use. 

There are a few places where details are discussed (e.g., [Host](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#header.host)); I think these should be moved into h1-messaging.

mnot:
Also some instances of "target URI"

mnot:
Going through Semantics:

[5.1 defines](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.5.1) "target resource" and "target URI" pretty well, and they (mostly target resource) are used consistently and appropriately throughout the spec, with these exceptions:

### 1.

> A server listens on a connection for a request, parses each message received, interprets the message semantics in relation to the identified request target

`s/the identified request target/its target resource/`

### 2.5.3.

> When not being used in absolute form as the request target of an OPTIONS request

Not sure, due to weirdness around OPTIONS.

### 2.5.4.

> A sender must not generate the userinfo subcomponent (and its "@" delimiter) when an "http" or "https" URI reference is generated within a message as a request target or field value.

`s/request target/target URI/`

### 3.3.

> For example, an origin server that publishes very long URI references to its own resources needs to be able to parse and process those same references when received as a request target.

`s/request target/target URI/`

### 5.4.

> For example, the request might have been misdirected, deliberately or accidentally, such that the information within a received request-target or Host header field differs from the host or port upon which the connection has been made.

`s/request-target/target URI/`

It would also be good to revisit why we need an _effective request URI_ that's separate from the concept of a _target URI_... I suspect all instances of the former could be replaced with the latter, and this section's content folded into 5.1 as appropriate.

### 5.6.2.

> If a proxy receives a request-target with a host name that is not a fully qualified domain name, it may add its own domain to the host name it received when forwarding the request. A proxy must not change the host name if the request-target contains a fully qualified domain name.

`s/request-target/target URI/g`

> A proxy must not modify the "absolute-path" and "query" parts of the received request-target when forwarding it to the next inbound server, except as noted above to replace an empty path with "/" or "*".

`s/request-target/target URI/`

### 7.6.3. 

> The CONNECT method requests that the recipient establish a tunnel to the destination origin server identified by the request-target

[...]

PR #316, opened 2020-02-07
map request target to target URI/resource

mnot:
For #259


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1094,28 +1094,28 @@
    caches can use them to keep their contents up to date.
 </t>
 <t>
-   A cache &MUST; invalidate the effective Request URI
-   (<xref target="effective.request.uri"/>) as well as the URI(s) in the
+   A cache &MUST; invalidate the target URI
+   (<xref target="target.resource"/>) as well as the URI(s) in the
```
royfielding: This (existing text) phrase is odd -- shouldn't it say that the cache invalidates the stored responses associated with the target URI, rather than invalidate the URI itself?

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1094,28 +1094,28 @@
...
 <t>
-   A cache &MUST; invalidate the effective request URI 
-   (<xref target="effective.request.uri"/>) when it receives a non-error response 
+   A cache &MUST; invalidate the target URI 
+   (<xref target="target.resource"/>) when it receives a non-error response 
    to a request with a method whose safety is unknown.
 </t>
 <t>
    Here, a "non-error response" is one with a <x:ref>2xx (Successful)</x:ref>
    or <x:ref>3xx (Redirection)</x:ref> status code. "Invalidate" means that
```
royfielding: Oh, so this is a defined term? That's weird. It should at least be defined before it is used, but would prefer to just inline the phrase above.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -170,8 +170,8 @@
 <t>
    Each HTTP message is either a request or a
    response. A server listens on a connection for a request, parses each
-   message received, interprets the message semantics in relation to the
-   identified request target, and responds to that request with one or more
+   message received, interprets the message semantics in relation to
+   its target resource, and responds to that request with one or more
```
royfielding: This would be better as "in relation to the identified target resource" (both to remove ambiguity and to emphasize that the message provides that identifier).

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4743,27 +4718,16 @@ Content-Range: exampleunit 11.2-14.3/25
...
-   the request-target or, if configured to use another proxy, by forwarding
-   the CONNECT request to the next inbound proxy.
-   Any <x:ref>2xx (Successful)</x:ref> response indicates
-   that the sender (and all inbound proxies) will switch to tunnel mode
-   immediately after the blank line that concludes the successful response's
-   header section; data received after that blank line is from the server
-   identified by the request-target.
-   Any response other than a successful response indicates that the tunnel
-   has not yet been formed and that the connection remains governed by HTTP.
+   The recipient proxy can establish a tunnel either by directly connecting
```
royfielding: But this has removed the necessary requirement that CONNECT does not send a URI in the request target. I hate that exception, always have, but we are stuck with it.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4822,16 +4785,16 @@ Proxy-Authorization: basic aGVsbG86d29ybGQ=
    of a server, without implying a resource action.
 </t>
 <t>
-   An OPTIONS request with an asterisk ("*") as the request-target
-   (<xref target="request.target"/>) applies to the server in general rather than to a
+   An OPTIONS request with an asterisk ("*") as the target URI's path component
```
royfielding: No, this change is wrong. It is not a URI and not a path component.

royfielding: We still need to differentiate between request-target and target URI. They are not the same thing. Likewise for all the below cases of "*" and CONNECT.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6421,8 +6384,8 @@ Expect: 100-continue
 </t>
 <t>
    A <x:dfn>protection space</x:dfn> is defined by the canonical root URI (the
```
royfielding: Note to self: this can now be replaced with origin and a link to the new section.

draft-ietf-httpbis-cache-latest.xml
mnot: Reading it, I agree -- will address in a new issue.

draft-ietf-httpbis-cache-latest.xml
mnot: See below.

draft-ietf-httpbis-semantics-latest.xml
mnot: Done in local source

mnot:
OK, I've reverted the changes specific to CONNECT and OPTIONS and opened #340 to tackle those separately.

mnot:
Please review soon, this is likely to rot quickly. Thx.

mnot:
@reschke @royfielding please review.

Issue #99, opened 2018-06-21
Scope of Retry-After


martinthomson:
Several recent discussions have raised the question of the scope of applicability of Retry-After.  It was clear from those discussions that some people interpreted the instruction as applying to the entire server.  Others take it to mean the request, others the resource.  It might also be possible to interpret the code as applying to all requests on a given connection, but that's not really consistent with the overall protocol design.

My understanding is that it can only apply to the request that is responding to, or at most the resource.  Requests are independent of each other and so - while a client might extend the scope by inference, or knowledge about the server - the only scope that the protocol can definitively say it applies to is either the request or the resource.

If it applies to the resource, it might be narrowed in scope to also apply to the method (POST is too expensive, but GET is fine, for example).

mnot:
Since it's already deployed, we can't retroactively define a scope; best we can do is define a new protocol element for those who wish to convey an explicit scope.

martinthomson:
Well, we might reasonably assume that the scope at least includes the request.  It's probably not talking about another request, but yeah, more work needed.

mnot:
I can see pointing out what needs to be defined, and that the core spec doesn't provide that. Do you think we can do anything else in core?

martinthomson:
The scope of "this request" is essentially meaningless.  Because when I make a subsequent request, that's a different request, even if the bits are all the same.  So I think that what we have is:

When a server provides a `Retry-After` header field in a response, it is an indication that making a similar request prior to the elapsed time is unlikely to result in a different answer. Clients are therefore requested to refrain from repeating the request within that time.

A `Retry-After` value can only apply to the specific request that it was sent in response to, but some conditions that lead to a server are necessarily broader in scope than a single request.  A server might include information in the response body to request that the client also refrain from making other requests to the same resource, or even to all resources on the same server.

Clients can apply heuristics to infer the scope that `Retry-After` applies to.  For instance, if a server indicates excessive load with a 503 (Service Unavailable) status code, a client might infer that the condition applies to more than one resource.

ioggstream:
> Several recent discussions have raised the question of the scope of applicability of Retry-After

@martinthomson Can you please link some of them in the issue? 

> best we can do is define a new protocol element for those who wish to convey an explicit scope.

imho adding stuff will make things more confused. 

> A Retry-After value can only apply to the specific request that it was sent in response to

I agree. Gateways may amplify this behavior as they can proxy different services behind the same Origin,
so I can get a 503 from an URL but not from another one.

[...]

PR #317, opened 2020-02-10
The scope of 503 may go beyond the request. See #99.

ioggstream:
## This PR

Clarifies that 503 may apply not only to the given request.
This is because `Retry-After` states:

> Retry-After indicates
>   how long the service is expected to be unavailable to the client.

Instead 503 states:

> Retry-After [suggests an]
> amount of time for the client to wait before retrying the request.

See: #99

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -8492,6 +8492,8 @@ Content-Type: text/plain
    The <x:dfn>503 (Service Unavailable)</x:dfn> status code indicates that the
    server is currently unable to handle the request due to a temporary overload
    or scheduled maintenance, which will likely be alleviated after some delay.
+   The above condition might apply to the issued request, to the resource
+   or to the entire server.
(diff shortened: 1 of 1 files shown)


ioggstream:
I'll prepare an I-D on that, if you think it's worth. Some ideas follows, wdyt?

### An header
- `Backoff-Scope`: header to be used to scope Retry-After or other headers (eg. ratelimit)

### Link relations
This can be header-agile

`Link: </books/>; rel="scope"; header="Retry-After"`
`Link: </books/>; rel="scope"; header="RateLimit-Limit"`

royfielding:
I am not sure what you mean by those proposals, but a new header field like "RateLimit-Scope" (if it's about that) or "Retry-Scope" would make more sense.

ioggstream:
@martinthomson following Roy's suggestion I created a short extension draft https://ioggstream.github.io/draft-polli-retry-scope/draft-polli-retry-scope.html

Feedback welcome.

ioggstream:
I linked the proposal in #99

PR #318, opened 2020-02-10
Suggest scoping retry-after using other message parts.

ioggstream:
## This PR

Suggests passing the scope of Retry-After using other message parts.
This includes the payload body or other to-be-defined headers or parameters.

Fixes: #99

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -8953,7 +8953,6 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
    applies to the target resource, to a specific path or to the whole server
    it MAY convey the scope of Retry-After in the payload body
    or in other message parts.
-
 </t>
 <t>
    The value of this field can be either an HTTP-date or a number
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8949,6 +8949,11 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
    When sent with any <x:ref>3xx (Redirection)</x:ref> response, Retry-After
    indicates the minimum time that the user agent is asked to wait before
    issuing the redirected request.
+   In general, if the server wants to detail whether Retry-After
+   applies to the target resource, to a specific path or to the whole server
+   it MAY convey the scope of Retry-After in the payload body
+   or in other message parts.
+
```
ioggstream:
```suggestion
```

royfielding:
No, this is not an interoperability requirement, nor is it part of HTTP. Yes, it might be a good idea to be able to express such a thing using the protocol, but that would be an extension and should be in a header field defined by a new specification.

PR #330, opened 2020-03-11
Missing fields

MikeBishop:
I didn't update the HTML, because the Makefile relies on Java.  I'm not installing Java just for this; sorry.


mnot:
Thanks!

Issue #331, opened 2020-03-19
Proxy-Authenticate scope


mnot:
P-A is [defined](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#header.proxy-authenticate) as:

> The "Proxy-Authenticate" header field consists of at least one challenge that indicates the authentication scheme(s) and parameters applicable to the proxy for this effective request URI (Section 5.5).

This seems really odd -- P-A is not generally used in a way that's specific to a single target URI.

mnot:
I think it's scoped to the request; i.e., if the client retries the same request with appropriate credentials, it's likely to succeed; we can't say much more than that.

reschke:
FTR, the text we have is inherited from <https://greenbytes.de/tech/webdav/rfc2616.html#header.proxy-authenticate>.

PR #337, opened 2020-03-19
Talk about exceptions to storing and updating headers

### LINKED ISSUE #165: Updating stored headers

mnot:
Fixes #165


mnot:
Updated to use 'fields' as appropriate.

This test relies upon the understanding that `Connection` operates only on headers, not trailers; 2616 [seems to support that](https://tools.ietf.org/html/rfc2616#section-14.10). 

If that's the case, I think we need to make the appropriate changes in Messaging; @royfielding should we do that here, or in a separate issue first?

royfielding:
The reason 

> The first MUST is right, but I believe the last three cases apply to both header and trailer fields. IOW, the names in Connection apply to trailers being forwarded as well.

The reason being that 2616 assumes the trailers would be merged as header fields (or simply dropped) prior to forwarding the message.

mnot:
@royfielding that makes sense, although I suspect even fewer implementations are going to be doing that; might be worth reinforcing in Semantics?

In any case, please review.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -386,6 +386,34 @@
...
+   before storage.</li>
+   <li>Likewise, fields that are known to be "hop-by-hop" are also required
+   to be removed before forwarding, and this &MAY; be implemented by doing so
+   before storage; see <xref target="header.connection"/> for some
+   examples.</li>
+   <li>Header fields that are specific to the connection they occur
+   &MUST-NOT; be stored. Effectively, this is limited to Proxy-Authenticate
+   <xref target="field.proxy-authenticate"/> and Proxy-Authorization <xref
+   target="field.proxy-authorization">, since new fields are
+   discouraged from being tied to a specific connection.</li>
```
mnot, line 409: Note that this is dependent upon the resolution of #331, which may not go this way.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -386,6 +386,34 @@
    prohibited from storing such responses.
 </t>
 
+<section title="Storing Header and Trailer Fields" anchor="storing.fields">
+<t>
+   Caches &MUST; include all received header fields &mdash; particularly
+   unrecognised ones &mdash; when storing a response; this assures that new HTTP
+   header fields can be successfully deployed. However, the following exceptions
```
reschke, line 393: maybe "including" instead of "particularly"?

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1043,8 +1071,14 @@
...
    to replace all instances of the corresponding header fields in the stored
-   response.
+   response, with the following exceptions:
 </t>
+<ul>
+   <li>The exceptions to header field storage in <xref
+   target="storing.fields"/> also apply to header field updates.</li>
+   <li>Caches &MUST-NOT; update the following header fields:
+   Content-Encoding, Content-Length, Content-MD5, Content-Range, ETag.</li>
+</ul>
```
reschke, line 1081: Link the definitions? (I can update the PR)

draft-ietf-httpbis-cache-latest.xml
mnot: Done.

draft-ietf-httpbis-cache-latest.xml
mnot: Please do!

reschke: Done.

mnot:
OK, I think this is good to go with that last change.

@royfielding could you please review?

mnot:
@royfielding see update

PR #338, opened 2020-03-19
Clarify public and private

### LINKED ISSUE #320: Public and Private vs. the world
The language for `Cache-Control: public` and `private` indicates that they override the rest of the caching model unconditionally; as discussed in #319, that is too broad.

Also, `public` is often used even though the message is still cacheable, because developers misunderstand its semantics.

mnot:
Fixes #320

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1621,7 +1621,7 @@
    an Authorization header field (<xref target="caching.authenticated.responses"/>).
 </t>
 <t>
-   Note that it is not necessary to add the public directive to a response if the response
+   Note that it is not necessary to add the public directive to a response that
    is already cacheable according to <xref target="response.cacheability"/>.
 </t>
 <t>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1613,14 +1615,22 @@
...
    an Authorization header field (<xref target="caching.authenticated.responses"/>).
 </t>
 <t>
-   If no explicit freshness information is provided, the response is
-   is heuristically cacheable (<xref target="heuristic.freshness"/>).
+   Note that it is not necessary to add public to a response if the response
+   is explicitly fresh (e.g., due to the presence of the max-age response
+   directive or the Expires header), or if the response is heuristically
+   cacheable (e.g., the 200 and several other response status codes; see
+   <xref target="heuristic.freshness"/>).
```
royfielding: But this paragraph is contradicting the sentence above it (an example where neither max-age nor Expires is sufficient to allow a shared cache to cache the response -- only s-max-age would). Can we combine those examples into one paragraph that shows when public is useful and when it is not?

Also, this paragraph fails to account for extension cache directives, whereas the one above does because of the reference to response.cacheability.

Honestly, I still see no reason to prevent people from marking a response as public even if it is already public by default. Maybe they are working around some other piece of software that has different heuristics?

mnot:
@royfielding see revision.

PR #341, opened 2020-03-20
Move TE: trailers into Semantics

### LINKED ISSUE #18: TE: trailers
The use case for `TE: trailers` is described in RFC7230, section 4.1.2:

> Unless the request includes a TE header field indicating "trailers" is acceptable, as described in Section 4.3, a server SHOULD NOT generate trailer fields that it believes are necessary for the user agent to receive. Without a TE containing "trailers", the server ought to assume that the trailer fields might be silently discarded along the path to the user agent. This requirement allows intermediaries to forward a de-chunked message to an HTTP/1.0 recipient without buffering the entire response.

Section 4.3 adds to that:

> The presence of the keyword "trailers" indicates that the client is willing to accept trailer fields in a chunked transfer coding, as defined in Section 4.1.2, on behalf of itself and any downstream clients. For requests from an intermediary, this implies that either: (a) all downstream clients are willing to accept trailer fields in the forwarded response; or, (b) the intermediary will attempt to buffer the response on behalf of downstream recipients

A few observations:
1. Concrete advice about when to generate `TE: trailers` would be helpful. E.g., should a User-Agent include it on all connections that support trailers? Should an intermediary include it on all connections when the upstream connection also supports trailers, _or_ it commits to buffering the response?
2. A server receiving `TE: trailers` can't infer much from it; it only knows that trailers are going to be processed by the path, but could still be dropped on the floor by the client (in 4.1.2: "recipient MAY process the fields"). Is it really a useful signal?

mnot:
and refine. Fixes #18 

In a perfect world, this would be a separate header field; as it is, we should figure out how to make this reference non-normative.


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1710,9 +1710,15 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
    can be safely merged.
 </t>
 <t>
-   A client can send a <x:ref>TE</x:ref> header field indicating "trailers"
-   is acceptable, as described in <xref target="header.te"/>, to inform the
-   server that it will not discard trailer fields.
+   The presence of the keyword "trailers" in the TE header field (<xref
+   target="header.te"/>) indicates that the client is willing to accept
+   trailer fields, on behalf of itself and any downstream clients. For
+   requests from an intermediary, this implies that either: (a) all
```
reschke: no (b)?

mnot: Ah, yes. In the new world, (b) isn't allowed. Will fix, thx.

mnot:
@reschke @royfielding PTAL.

mnot:
@reschke ?

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1540,19 +1540,6 @@ https://www.example.org
...
-   <xref target="chunked.trailer.section"/>, on behalf of itself and any downstream
-   clients. For requests from an intermediary, this implies that either:
-   (a) all downstream clients are willing to accept trailer fields in the
-   forwarded response; or,
-   (b) the intermediary will attempt to buffer the response on behalf of
-   downstream recipients.
-   Note that HTTP/1.1 does not define any means to limit the size of a
-   chunked response such that an intermediary can be assured of buffering the
-   entire response.
-</t>
```
reschke, line 1555: We seem to lose the note in the last paragraph - is this acceptable?

mnot: Personally I think it's OK to lose it; it's a logical conclusion of the design that someone can infer on their own.

reschke: Ok. Maybe move that statement to "Chunked Encoding" instead?

PR #342, opened 2020-03-20
Describe scope concerns for extensions

### LINKED ISSUE #265: Scope of applicability for HTTP extensions
When defining new header fields or methods or any other extensions to HTTP, it can be the case that those elements make statements where the scope is unclear.  For instance, in #99 we learned that `Retry-After` often states something that applies to an entire server (or server instance), or maybe a subset of resources, or maybe just the request itself.  HTTP generally assumes that the scope of applicability is a request, but it might also be the specific representation or the resource.

New protocol extensions need to consider carefully what the scope of applicability is for new protocol elements and the information they carry.  Adding text about this to the document (to expand #50) would be good.

mnot:
Fixes #265

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -8781,8 +8781,8 @@ Content-Type: text/plain
 <t>
    By default, a status code applies only to the request corresponding to the
    response it occurs within. If a status code applies to a larger scope of
-   applicability -- for example, all requests to the resource in question, or
-   all requests to a server -- this must be explicitly specified. When doing
+   applicability &mdash; for example, all requests to the resource in question, or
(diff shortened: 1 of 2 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1776,6 +1776,14 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
   <li><t>Under what conditions the field can be used; e.g., only in
     responses or requests, in all messages, only on responses to a
     particular request method, etc.</t></li>
+   <li><t>What the scope of applicability for the information conveyed in
+    the header is. By default, headers apply only to the message they are
```
reschke: field?

mnot: Fixed. It's going to take a while to do this consistently...

PR #343, opened 2020-03-20
Adjust multiple Content-Length handling

### LINKED ISSUE #59: Multiple Content-Length values that mismatch
The text that stipulates reject or replace at https://tools.ietf.org/html/rfc7230#section-3.3.2 should probably say that if the values differ than browsers should definitely reject. Only Chrome currently does so consistently per https://github.com/w3c/web-platform-tests/pull/10548, but I'm going to try to get the others on board.

mnot:
... to account for terminology changes, and to clarify that
it doesn't matter whether the values are on different lines.

Fixes #59.


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1128,10 +1128,11 @@ https://www.example.org
...
    If a message is received without <x:ref>Transfer-Encoding</x:ref> and with
-   either multiple <x:ref>Content-Length</x:ref> header fields having
-   differing field values or a single Content-Length header field having an
-   invalid value, then the message framing is invalid and
-   the recipient &MUST; treat it as an unrecoverable error.
+   an invalid <x:ref>Content-Length</x:ref> header field, then the message
+   framing is invalid and the recipient &MUST; treat it as an unrecoverable
+   error, unless the field value can be successfully parsed as a
+   comma-separated list (<xref target="abnf.extension"/>), all values in the
+   list are valid, and all values in the list are the same.
```
annevk: So per https://github.com/web-platform-tests/wpt/pull/10548 `30` and `030` are different values. I guess this technically covers that, but maybe some examples would help?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3613,15 +3613,15 @@ bytes=500-700,601-999
...
-   determining the message body length or forwarding the message.
+   If a message is received that has a Content-Length header field value
+   consisting of the same decimal value as a comma-separated list (<xref
+   target="abnf.extension"/>) -- for example, "Content-Length: 42, 42" --
+   indicating that duplicate Content-Length header fields have been generated
+   or combined by an upstream message processor, then the recipient &MUST;
+   either reject the message as invalid or replace the duplicated field
+   values with a single valid Content-Length field containing that decimal
+   value prior to determining the message body length or forwarding the
+   message.
```
annevk: This already existed, but is replacing really the right language? A networking library should probably not alter the response before passing it on to some API endpoint. At least not for this reason.

draft-ietf-httpbis-messaging-latest.xml
mnot: I'm thinking of adding "... when compared numerically."

draft-ietf-httpbis-semantics-latest.xml
mnot: "forwarding the message" here means to another HTTP hop, not surfacing it to internal code. We don't specify that.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1132,7 +1132,8 @@ https://www.example.org
    framing is invalid and the recipient &MUST; treat it as an unrecoverable
    error, unless the field value can be successfully parsed as a
    comma-separated list (<xref target="abnf.extension"/>), all values in the
-   list are valid, and all values in the list are the same.
+   list are valid, and all values in the list are the same when compared
+   numerically.
```
reschke: Hm, that's a change from RFC 7230. Do we really need to do that? If so, we need to mention it in the Changes section.

mnot: I think we do; will do.

annevk:
Sorry, I don't think we want "when compared numerically" as that's not what implementations do. I was mainly stating that perhaps we should state explicitly that's not what it does.

draft-ietf-httpbis-messaging-latest.xml
mnot: I reverted that, since it's not clear what interoperability really is here.

mnot:
@reschke you wanted to re-review?

PR #344, opened 2020-03-20
Don't rely on the HTTP/1.1 Messaging specification to define "message"

### LINKED ISSUE #311: Define "request" and "response" independent of HTTP/1.1
It looks like "request" and "response" are defined in https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#operation as kinds of "messages", and then "messages" are defined in the same section by reference to https://httpwg.org/http-core/draft-ietf-httpbis-messaging-latest.html#http.message ... which only applies to HTTP/1.1. I *think* these are intended to be version-independent concepts, so they need a version-independent definition.

mnot:
Fixes #311

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -381,8 +381,7 @@
 <iref primary="true" item="connection"/>
 <t>
    HTTP is a stateless request/response protocol that operates by exchanging
-   <x:dfn>messages</x:dfn> (<xref target="http.message"/>) across a reliable
-   transport- or session-layer
+   <x:dfn>messages</x:dfn> across a reliable transport- or session-layer
    "<x:dfn>connection</x:dfn>" (<xref target="connection.management"/>).
    An HTTP "<x:dfn>client</x:dfn>" is a program that establishes a connection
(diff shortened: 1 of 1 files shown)


MikeBishop:
Assuming I'm reading the tags right, you mark this as a "defined term," but just drop the link to the definition.  I don't feel like that helps, since you now have a dangling defined term.

mnot:
@MikeBishop as I said in #311, this is the most minimal approach. To do more, I think we need to rename 4. Header and Trailer Fields to something like "Messages", and expand its introduction accordingly. 

Any thoughts about that?

MikeBishop:
It seems like most of the content you'd want for defining a message is already there in 2.1:

> Nevertheless, a common abstraction is that a message includes some form of envelope/framing, a potential set of named fields up front (a header section), a potential body, and a potential following set of named fields (a trailer section).

Rearranging text a little to put that definition (and make it a definition, rather than "a common abstraction") by the introduction of the term "message" seems like it would work without having to rearchitect Section 4.

mnot:
@MikeBishop `x:dfn` denotes use of a defined term, not a primary definition.

mnot:
@royfielding ?

PR #345, opened 2020-03-20
Move definition of "complete" to semantics

### LINKED ISSUE #334: Defining "incomplete"
Right now, caching refers to h1-messaging to define what an incomplete message is. That's a more general concept, so it belongs in semantics.

Might be worth taking a look at the definition of "partial" too...

mnot:
Fixes #334

Caution against servers using close was already covered in messaging.


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -452,6 +454,16 @@
...
 </t>
+<t>
+   One of the functions of the message framing mechanism is to assure that
+   messages are <x:dfn>complete</x:dfn>. A message is considered complete
+   when all of the octets indicated by its framing are available. Note that,
+   when no explicit framing is used (such as with the "close" connection
+   option in HTTP/1.1; see <xref target="header.connection"/>), a response
+   message that is ended by the transport connection's close is considered
+   complete even though it might be indistinguishable from an incomplete
+   response, unless a transport-level error indicates that it is not complete.
```
royfielding, line 465: This is a different definition and it's incorrect. The close connection option has nothing to do with this.

The prior text is talking about close-terminated messages, which is probably better described as a response message without explicit framing in HTTP/1.x (i.e., a response that ends because the connection closed and there is no metadata or framing to indicate that it closed early).

PR #346, opened 2020-03-22
Move "invalidate" definition up

### LINKED ISSUE #339: Invalidation language is weird
Roy noticed that we use a special meaning of "invalidate" and then define it pretty casually below. This can be improved.

mnot:
Fixes #339


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1099,7 +1099,17 @@
 <t>
    Because unsafe request methods (<xref target="safe.methods"/>) such as PUT, POST or DELETE
    have the potential for changing state on the origin server, intervening
-   caches can use them to keep their contents up to date.
+   caches are required to invalidate stored responses to keep their contents up to date.
+   <x:dfn>Invalidate</x:dfn> means that the cache will either remove all
```
reschke: Not sure about the change from "can use" to "are required".

Are we making a normative change here?

mnot: The requirement is already stated below; this is not normative, just aligning the prose with the requirements.

mnot:
@royfielding please review.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1099,7 +1099,17 @@
...
+   stored responses related to the effective request URI or will mark these
```
royfielding, line 1104: I'd prefer "corresponding to" instead of "related to".

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1099,7 +1099,17 @@
...
+   as "invalid" and in need of a mandatory validation before they can be sent
+   in response to a subsequent request.
+</t>
+<t>
+   Note that this does not guarantee that all appropriate responses are
+   invalidated globally. For example, a state-changing request might invalidate
+   responses in the caches it travels through, but relevant responses still
+   might be stored in other caches that it has not.
```
royfielding, line 1112: I'm not sure if this needs to be said, but if it does then I think it reads better as:

```
   Note that this does not guarantee that all appropriate responses are
   invalidated globally: a state-changing request would only invalidate
   responses in caches that it travels through.
```

draft-ietf-httpbis-cache-latest.xml
mnot: I think the correct thing to say here is something like:

> Invalidate means that the cache will either remove all responses whose target URI matches the given URI, or...

because the actual requirements about what URI to invalidate are given below.

royfielding: okay

PR #348, opened 2020-03-22
Refine requirements on If-Unmodified-Since

### LINKED ISSUE #326: Handling of If-Unmodified-Since with unknown modified dates
I'm working on a new implementation of conditionals, and I'm a little confused about `If-Unmodified-Since`.

What should a server do if a client passed an `If-Unmodified-Since` header, but the target resource does not know its own modification date?

My intuition is that without a `Last-Modified` header, we should assume that the request is *always fresh*, and given that `If-Unmodified-Since` is used to protect against the 'lost update' problem, we should default to returning `412` if the last modification date could not be determined.

I don't think the current draft covers this case, but perhaps I've missed something.

mnot:
Fixes #326


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5902,11 +5902,13 @@ Expect: 100-continue
...
 <t>
    An origin server that receives an If-Unmodified-Since header field &MUST;
-   evaluate the condition prior to performing the method
-   (<xref target="evaluation"/>).
-   The origin server &MUST-NOT; perform the requested method if the selected
-   representation's last modification date is more recent than the date
-   provided in the field value; instead the origin server &MUST; respond with either
+   evaluate the condition (<xref target="evaluation"/>) prior to performing
+   an unsafe method (<xref target="safe.methods"/>).
+</t>
```
reschke: This changes the requirement to apply only to unsafe methods. Is this really intended?

mnot: Yes. Requiring servers to support If-Unmodified-Since on all resources for all request methods isn't viable. Happy to take it to the list if you're concerned, though.

mnot: Looking at this again, I don't think this change is supportable; it works that way in a couple of major servers that I spot checked. So, change backed out.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5911,11 +5911,13 @@ Expect: 100-continue
...
-   (<xref target="evaluation"/>).
-   The origin server &MUST-NOT; perform the requested method if the selected
-   representation's last modification date is more recent than the date
-   provided in the field value; instead the origin server &MUST; respond with either
+   evaluate the condition (<xref target="evaluation"/>) prior to performing
+   the method.
+</t>
+<t>
+   If the origin server has a last modification date available for the selected
+   representation, it &MUST-NOT; perform the requested method if it is more recent than the date
```
royfielding, line 5919: Minor preference over multiple "it" usage:

```
   If the selected representation has a last modification date, the
   origin server MUST NOT perform the requested method if that
   date is more recent than the date provided in the field value.
```

mnot:
After merging this, we should align IMS and IUS with the approach taken in #357

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5932,6 +5934,11 @@ Expect: 100-continue
    field in the response unless it can verify that the request is a duplicate
    of an immediately prior change made by the same user agent.
 </t>
+<t>
+   If the origin server does not have a last modification date available for
+   the selected representation, the If-Modified-Since header field &MUST;
+   be ignored.
+</t>
```
reschke, line 5941: That's a new normative requirement, right? I do not object, but it would need to be called out in changes-from-rfc723*.

royfielding: Perhaps it would be easier to state that the condition evaluates to false if the target resource has no selected representation or if the selected representation has no last modification date available, since both would indicate that the resource has changed.

mnot: It's already implied by the new clause above, I think it's OK to just delete this new requirement.

mnot:
Personally, I think that reduces the utility of the "changes from..." sections -- if we start noting changes this small, they're going to be swamped. I'm not going to refuse to put that in, but I don't think it's the right thing to do. @royfielding thoughts?

reschke:
I consider the "changes from" sections the place where an implementor would need to go to see what might need to be changed in an implementation. (so yes, this is an edge case, but shouldn't we include it in this case?)

PR #350, opened 2020-03-23
Remove obs-fold from semantics

### LINKED ISSUE #116: Should Semantics assume that field values have been coalesced or just unfolded?
It is hard to talk about header fields without also noting they might appear across multiple lines via repeated names or obs-fold lines.  My goal has been to describe the coalescing of fields in Messaging and only deal with complete values in Semantics.  However, I think only obs-fold is specific to HTTP/1.x.

For now, we have the following text (currently in crefs) in the "Field Values" section of Semantics:

> This document assumes that any such obs-fold has been replaced with one or more SP octets prior to interpreting the field value, as described in (xref target="line.folding").

mnot:
Fixes #116


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1454,7 +1452,7 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
    was received as a token or a quoted string.
 </t>
 <t>
-   Historically, HTTP field values could be extended over multiple
+   Historically, HTTP/1.0 and HTTP/1.1 field values could be extended over multiple
    lines by preceding each extra line with at least one space or horizontal
    tab (<x:ref>obs-fold</x:ref>).
    <cref>This document assumes that any such <x:ref>obs-fold</x:ref> has been replaced with one or more
```
royfielding, line 1458: we would have to fix this reference as well. I suggest using a ref to Messaging instead.

mnot:
@reschke https://httpwg.org/specs/rfc7540.html#rfc.section.10.3. See also discussion on #116.

draft-ietf-httpbis-semantics-latest.xml
mnot: I just removed "1.0 and HTTP/1.1" -- there's an appropriate reference to messaging in the next sentence.

mnot:
@reschke why was this closed?

royfielding:
It seems to have been closed by accident instead of merging into master.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1457,9 +1455,9 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
    Historically, HTTP field values could be extended over multiple
    lines by preceding each extra line with at least one space or horizontal
    tab (<x:ref>obs-fold</x:ref>).
-   <cref>This document assumes that any such <x:ref>obs-fold</x:ref> has been replaced with one or more
+   This document assumes that any such <x:ref>obs-fold</x:ref> has been replaced with one or more
    <x:ref>SP</x:ref> octets prior to interpreting the field value,
-   as described in <xref target="line.folding"/>.</cref>
+   as described in <xref target="line.folding"/>.
```
royfielding, line 1460: Note that this is still referencing obs-fold as an ABNF target. It would be better to replace that with "obsolete line folding". I.e.,

```
   Historically, HTTP/1.x field values could be extended over multiple
   lines by preceding each extra line with at least one space or horizontal
   tab (obsolete line folding). This document assumes that obsolete line folding
   is removed by the HTTP/1.x message parsing algorithm (<xref target="line.folding"/>)
   prior to interpreting the field value.
```

reschke:
Sorry for that. I may have misread the merge *from* master.

Issue #335, opened 2020-03-19
Make references to h1-messaging non-normative


mnot:
Both caching and semantics make some references to h1-messaging. They should not be normative.

We can do this by:
 * Removing unnecessary references
 * Turning references into examples / advisory text
 * Clarifying H1 specificity where appropriate 

Includes #332, #333, #311, #259, #183, #182, #118.

mnot:
In semantics:

* `chunked.trailer.section` - 1 ref, is an example
* `connection.management` - 2 refs, one defines "connection", the other is an aside.
*  `differences.between.http.and.mime` - 1 ref in Canonicalization and Text Defaults
* `field.parsing` - 1 ref, talking about leading and trailing WS being discarded
* `h1.effective.request.uri` - 1 ref, illustrative
* `request.target` - see #259, #182 
* `header.connection` - 4 refs:
   - proxy forwarding headers
   - defining hop-by-hop in considerations for new fields
   - requirement for intermediary forwarding messages
   - connection: close on 408
* `header.te` - talking about TE: trailers
* `header.transfer-encoding` - 1 ref, contrasting with Content-Encoding
* `header.upgrade` - 3 refs, definitions of 101, 426 and Via grammar
* `http.message` - 4 refs, definition of message. 
* `media.type.message.http` - 1 ref, requirement from TRACE
* `line.folding` - 2 refs, talking about OWS folding
* `message.body` - 2 refs, talking about message body, parsing reqs.
* `persistent.connections` - 0 refs
* `persistent.tear-down` - 1 ref, Expect requirements
* `request.line` - 1 ref, size limits
* `request.smuggling` - 2 refs, field limits and GET
* `transfer.codings` - 3 refs: proxy transformations, contrasting with content-coding, message/http

mnot:
@royfielding it seems like either `header.upgrade` needs to move into Semantics, or `101` and `426 need to move into Messaging. WDYT?

mnot:
In Caching, there's are two references to `message.body.length`, in Storing Incomplete Responses (talking about what "incomplete" is; see #334), and as a 1.1-specific example.

I think once #334 is resolved, Caching should be good.

royfielding:
> @royfielding it seems like either `header.upgrade` needs to move into Semantics, or `101` and `426 need to move into Messaging. WDYT?

Yes, I think we will be better served by defining all of the header fields in Semantics and stating explicitly when they are only applicable to one version of HTTP (or some subset of versions).

[...]

PR #351, opened 2020-03-23
Move 101 and 426 to Messaging

mnot:
for #335. Splitting this out for separate review.


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -2024,7 +2024,7 @@ https://www.example.org
 </t>
 </section>
 
-<section title="Upgrade" anchor="header.upgrade">
+<section title="Upgrading Connections" anchor="header.upgrade">
```
reschke, line 2027: I would either leave the title as is, or insert a subsection that defines the field itself.

mnot:
I suspect that we should be consistent regarding how we treat Upgrade and CONNECT, because both are extensions that use generic extensibility points, but have version-specific implementation, because they are peering into the actual connection management, not only expressing application semantics.

So maybe that means we define their general meaning in semantics, but their actual implementation in messaging (with the implication that h2 and h3 can do so if they choose to, but absent that, they can't be used there).

Does that make sense?

reschke:
Sounds right to me.

MikeBishop:
That seems reasonable, particularly as HTTP/2 and HTTP/3 already define how CONNECT is used.

reschke:
I rebased and fixed the build. This is the first case where status code definitions would live outside the Semantics document (and that impacts IANA registration tables etc). My conclusion is that the status codes should stay in Semantics.

PR #352, opened 2020-03-23
Define CONNECT and OPTIONS in term of target URIs

### LINKED ISSUE #340: CONNECT, OPTIONS and request-target
As discussed in #259, CONNECT and OPTIONS both have special request-target forms. This needs to be untangled so the terminology is not specific to H1.

mnot:
Fixes #340


mnot:
PTAL.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2123,16 +2123,31 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
+<t>
+   However, there are two special, method-specific forms allowed for the
+   request target in specific circumstances:
+</t>
+<ul>   
+<li>For CONNECT (<xref target="CONNECT"/>), the request target is the host
+name and port number of the tunnel destination, separated by a colon.</li>
+<li>For OPTIONS (<xref target="OPTIONS"/>), the request target can be a
+single asterisk ("*").</li>
+</ul>
```
reschke, line 2147: (indentation)

reschke: also maybe refer to #asterisk-form in messaging?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2123,16 +2123,31 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
+<t>
+   See the respective method definitions for details. These forms &MUST-NOT;
+   be by with other methods.
 </t>
```
reschke, line 2151: missing "used"?

draft-ietf-httpbis-semantics-latest.xml
mnot: Indentation fixed. WRT #asterisk-form, I'd be more inclined to move most of that into Semantics. Thoughts?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -732,7 +732,7 @@ Content-Type: text/plain
    The target of an HTTP request is called a "<x:dfn>resource</x:dfn>". 
    HTTP does not limit the nature of a resource; it merely
    defines an interface that might be used to interact with resources.
-   Each resource is identified by a Uniform Resource Identifier (URI), as
+   Most resources are identified by a Uniform Resource Identifier (URI), as
    described in <xref target="uri"/>.
```
MikeBishop, line 736: Framing it this way suggests that the server itself (`asterisk-form`) and other network endpoints (`authority-form`) are resources without URIs.  This isn't necessarily wrong, but feels like an odd formulation.  Would it be better to say that these oddball methods can address things other than resources?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2130,16 +2130,35 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
...
+      name and port number of the tunnel destination, separated by a colon.
+   </li>
+   <li>
+      For OPTIONS (<xref target="OPTIONS"/>), the request target can be a
+      single asterisk ("*").
+   </li>
+</ul>
+<t>
+   See the respective method definitions for details. These forms &MUST-NOT;
+   be used with other methods.
```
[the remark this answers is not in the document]
royfielding: No, it is forbidden for all methods != CONNECT (these exception are hell to parse when recipients are expected to forward unknown methods).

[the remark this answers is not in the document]
MikeBishop: We might want to send some alternative suggestions to MASQUE, then.

draft-ietf-httpbis-semantics-latest.xml
mnot: Lots of resources don't have URIs; e.g., the state in the browser is a resource, but it usually doesn't have one.

mnot:
@royfielding ?

draft-ietf-httpbis-semantics-latest.xml
royfielding: Well, now, that depends on whose spec is talking.

The word resource implies you can use it at some later time, hence re-source it. That is difficult without an identifier. In HTTP, "resource" is a defined term and every HTTP resource has at least one (usually much more than one) URI, not limited to http(s) URIs, but a message recipient might not know any of those URIs. The server itself and other network endpoints have URIs, but we don't use those URIs in the method-specific HTTP/1.1 requests for historical (foolish) reasons. Hence, the target of an HTTP request (semantics) is a resource, but the request-target (bits on wire) may only transmit a small portion of that resource's identifying data obtained from a URI.

The word resource is also used by WHATWG specs, with an incompatible definition, due to NIH syndrome. Whether or not the state within a browser is a resource, or has a URI (which is almost impossible to avoid within a browser given about: URIs), is another very long and painful story.

mnot: Looking through the OED, that's a very creative interpretation of the etymology of _resource_, but nevertheless interesting (and that was the meaning I was thinking of, not the WHATWG's).

Regardless, could you review the PR?

draft-ietf-httpbis-semantics-latest.xml
reschke: Yes, we probably need that in semantics.

mnot:
Paging @royfielding

mnot:
@royfielding can you please review? This is blocking other things.

PR #353, opened 2020-03-23
Make text body narrower for readability

mnot:
Currently we have about 110+ characters per line;
broadly accepted norms for readability are 55-65.

This takes it to about 85.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ diff00:
           egrep -v -- '^<\!-- ' > diffs/diff_auth_since_00.html
 
 %.html: %.xml $(stylesheet)
-	$(saxon) $< $(xreffer) | $(saxon) - $(stylesheet) xml2rfc-ext-maxwidth=900 xml2rfc-ext-styles="ff-noto ffb-sans-serif fft-sans-serif header-bw" xml2rfc-ext-paragraph-links=yes | awk -f lib/html5doctype.awk > $@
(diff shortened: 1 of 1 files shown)


martinthomson:
I tested this out and the rule on `pre.text, pre.text2` is `width: 69em`.  That is 993.6px for me, which looks really odd. You could almost miss it with a wider page, but it really sticks out now.  I'd recommend going to:

```css
pre.text, pre.text2 {
  width: 100%;
  width: calc(100% - 3.5em - 2px);
}
```
That would allow for the extra margin, border, and padding on the element.

reschke:
The purpose here is that the width reflects what can go inside artwork, not to scale it according to the page width. After all, we're not breaking lines here.

I'm not happy with what I have here as the width computation is excessive; but something like that will need to stay.

martinthomson:
So with the font size you have, 69em is way too much space.  Assuming 72 characters is your goal, you can go down to 44em or 625px (though slightly more is better).

reschke:
@martinthomson - thanks for the link; will have a look.

mnot:
@martinthomson you want to make some modifications?

martinthomson:
It appears as though Julian has [fixed the pre.text rule in the main stylesheet](https://github.com/reschke/xml2rfc/commit/a3149eae511fe851a13caf35346d8f0bb8fcdfa2), which is what would be needed.  So this should be good to go.

Issue #144, opened 2018-10-10
"Connection"


mnot:
Semantics uses "Connection" to mean both "the underlying transport-layer substrate" and "the logical channel that a particular request/response exchange uses" (roughly).

They are the same in H1, but not in H2 and later. Should we consider using a separate term for the latter (e.g., "channel")?

LPardue:
A different term would help when discussing proxying in H2 or HQ, where a channel is carried over a stream.

mnot:
How many downstream specs refer to a HTTP connection? Hopefully none, since it's a stateless protocol.

mnot:
Also, I suspect that "stream" would work very well, since it's semantics in h2 and QUIC is what we're looking for.

royfielding:
stream does not work -- that is a unidirectional term

mnot:
It's not unidirectional in H2, and only sometimes in QUIC (currently).

Looking at Semantics again, it seems like the most minimal change we could make would be to:

* Add text at the start of [2.1 client/server messaging](
https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#operation) explicitly distinguishing between "connection" and "transport connection", with "TCP connection" a subclass of the latter.

* In [1. Introduction](), "This document obsoletes the portions of RFC 7230 that are independent of the HTTP/1.1 messaging syntax and *transport* connection management"

* The sentence in 2.1 starting with "A connection might be used for multiple..." is confusing; I'm not sure what to do here yet.

* in [2.5.1 http URI Scheme](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#http.uri), change 5th para 1st sentence to "Note that the presence of a URI with a given authority component does not imply that there is always an HTTP server listening for *TCP* connections on that host and port."

* Same section, second to last sentence - s/underlying connection protocol/underlying transport protocol/ and s/end-to-end secured connection/end-to-end secured transport/

* In [5.2 Routing Inbound](
https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#routing.inbound), "HTTP requirements regarding *TCP* connection management *in HTTP/1.1* are defined in..."

* In [8.5.1 Challenge and Response](
https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#challenge.and.response), s/underlying connection/underlying transport connection/

* In [8.5.5.2](
[...]

PR #354, opened 2020-03-23
Remove H1-specific connection-related requirements

mnot:
For #144

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -766,7 +766,7 @@ Content-Type: text/plain
 <t>
    Note that the presence of an "http" or "https" URI does not imply that
    there is always an HTTP server at the identified origin listening for
-   TCP connections. Anyone can mint a URI, whether or not a server exists and
+   connections. Anyone can mint a URI, whether or not a server exists and
    whether or not that server currently maps that identifier to a resource.
    The delegated nature of registered names and IP addresses creates a
    federated namespace whether or not an HTTP server is present.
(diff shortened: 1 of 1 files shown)


mnot:
On a case-by-case basis:

* "HTTP requirements regarding connection management" is an informative reference to h1-messaging, so it's already there;
* The requirements for 205 are just a re-statement of existing h1 message framing requirements (and duplicative); 
* The requirement for 408 is a bit odd; it's really just spelling out the implications for being in that state in h1. I don't know why it's a SHOULD.
* 413 has been refactored so it's not h1 specific, but still applicable to it.

reschke:
Ack!

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -767,7 +766,7 @@ Content-Type: text/plain
 <t>
    Note that the presence of an "http" or "https" URI does not imply that
    there is always an HTTP server at the identified origin listening for
-   connections. Anyone can mint a URI, whether or not a server exists and
+   TCP connections. Anyone can mint a URI, whether or not a server exists and
```
royfielding, line 769: Why was TCP added here? It's not desired.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8527,8 +8511,9 @@ Content-Range: bytes 7000-7999/8000
    The <x:dfn>413 (Payload Too Large)</x:dfn> status code indicates
    that the server is refusing to process a request because the request
    payload is larger than the server is willing or able to process.
-   The server &MAY; close the connection to prevent the client from continuing
-   the request.
+   The server &MAY; terminate the request, if the protocol version in use
+   allows it; otherwise, the server &MAY; close the transport
```
royfielding, line 8515: connection, not "transport connection"

PR #360, opened 2020-04-30
Rename anchors from "header." to "field."

reschke:


mnot:
*sigh* this is causing a lot of conflicts with other pull requests. 

Also, in many cases what's being referred to _is_ a header -- i.e., it's only specified to be used in the header section, not the trailer section. I think we should use that terminology consistently, so as they're not confusion (i.e., pointedly calling it a field with no qualification can lead people to believe that it's valid in the trailer section).

reschke:
Hm.

(a) I recall this is what we discussed.
(b) I'm rebasing the PRs right now.
(c) It's valuable to have a predictable anchor name for any header|trailer field. Requiring knowledge about the details defeats the purpose of having that predictable syntax.
(d) The anchor name is something that does not show up in the normative text, but just while navigating the HTML.

royfielding:
It also changes incoming links and is now confusing with the other, previously existing usage of field.blah anchors in the common field syntax sections.

My preference is to revert in favor of the old anchor names, or remove the "field." prefix for anchors to header field definitions (this might require renaming any bare anchor names that conflict).

reschke:
For HTML generated by us, the links do not break, as they are aliased.

The suggestion to remove "field." might work; I'll check that.

mnot:
HTML generated by the RFC Editor and tools.ietf.org, links will break.

reschke:
> For HTML generated by the RFC Editor and tools.ietf.org, links will break.

How. The RFC Editor does not gerate HTML linking to drafts with named anchors. Nor tools.ietf.org.

reschke:
FWIW, we might be able to use the lowercase form of the field name as anchor...

PR #362, opened 2020-05-06
remove request-line and status-line from Semantics

### LINKED ISSUE #361: status-line and request-line in Semantics
These are HTTP/1-specific syntax, not semantics.

_Originally posted by @mnot in https://github.com/httpwg/http-core/issues/259#issuecomment-600949550_

mnot:
Fixes #361

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -5299,7 +5299,7 @@ Proxy-Authorization: basic aGVsbG86d29ybGQ=
    A <x:dfn>100-continue</x:dfn> expectation informs recipients that the
    client is about to send a (presumably large) message body in this request
    and wishes to receive a <x:ref>100 (Continue)</x:ref> interim response if
-   the request-line and header fields are not sufficient to cause an immediate
+   the method, target URI and header fields are not sufficient to cause an immediate
    success, redirect, or error response. This allows the client to wait for an
    indication that it is worthwhile to send the message body before actually
    doing so, which can improve efficiency when the message body is huge or
@@ -5380,21 +5380,21 @@ Expect: 100-continue
    </li>
 </ul>
 <t>
-   An origin server &MUST;, upon receiving an HTTP/1.1 (or later) request-line
-   and a complete header section that contains a 100-continue expectation and
+   An origin server &MUST;, upon receiving an HTTP/1.1 (or later) request that has a method, target URI
+   and complete header section that contains a 100-continue expectation and
    indicates a request message body will follow, either send an immediate
    response with a final status code, if that status can be determined by
-   examining just the request-line and header fields, or send an immediate
+   examining just the method, target URI and header fields, or send an immediate
    <x:ref>100 (Continue)</x:ref> response to encourage the client to send the
    request's message body. The origin server &MUST-NOT; wait for the message
    body before sending the <x:ref>100 (Continue)</x:ref> response.
 </t>
 <t>
-   A proxy &MUST;, upon receiving an HTTP/1.1 (or later) request-line
-   and a complete header section that contains a 100-continue expectation and
+   A proxy &MUST;, upon receiving an HTTP/1.1 (or later) request that has a method, target URI
+   and complete header section that contains a 100-continue expectation and
    indicates a request message body will follow, either send an immediate
    response with a final status code, if that status can be determined by
-   examining just the request-line and header fields, or begin forwarding the
+   examining just the method, target URI and header fields, or begin forwarding the
    request toward the origin server by sending a corresponding request-line
    and header section to the next inbound server. If the proxy believes (from
    configuration or past interaction) that the next inbound server only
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5299,7 +5299,7 @@ Proxy-Authorization: basic aGVsbG86d29ybGQ=
    A <x:dfn>100-continue</x:dfn> expectation informs recipients that the
    client is about to send a (presumably large) message body in this request
    and wishes to receive a <x:ref>100 (Continue)</x:ref> interim response if
-   the request-line and header fields are not sufficient to cause an immediate
+   the method, target URI and header fields are not sufficient to cause an immediate
```
annevk, line 5302: Directly below you use an Oxford comma. You probably want to be consistent (and add the comma 😊). Same below a few times.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4362,7 +4362,7 @@ Content-Range: exampleunit 11.2-14.3/25
     </tr>
     <tr>
       <td>HEAD</td>
-      <td>Same as GET, but only transfer the status line and header section.</td>
+      <td>Same as GET, but only transfer the status code and header section.</td>
       <td><xref target="HEAD" format="counter"/></td>
```
reschke, line 4366: Maybe here we can simplify to "...but do not transfer the response body."?

draft-ietf-httpbis-semantics-latest.xml
mnot: Done.

draft-ietf-httpbis-semantics-latest.xml
mnot: Done, but it raises an interesting question: should HEAD omit the trailer section? I think so, but in theory it could be included w/ h2 and h3.

mnot:
Re "status information" -- I had the same thought, but we already use 'status code' as shorthand elsewhere, so it seemed better to stay with that.

draft-ietf-httpbis-semantics-latest.xml
royfielding: The only stuff sent in a trailer is information generated while sending the body. No server is going to generate and send a body to /dev/null just for the sake of finding trailer info. Apache won't even do that to generate header fields like Content-Encoding.

mnot: Fair enough.

mnot:
@reschke @royfielding please review.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5380,21 +5380,21 @@ Expect: 100-continue
    </li>
 </ul>
 <t>
-   An origin server &MUST;, upon receiving an HTTP/1.1 (or later) request-line
-   and a complete header section that contains a 100-continue expectation and
+   An origin server &MUST;, upon receiving an HTTP/1.1 (or later) request that has a method, target URI
```
annevk, line 5383:
```suggestion
   An origin server &MUST;, upon receiving an HTTP/1.1 (or later) request that has a method, target URI,
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5380,21 +5380,21 @@ Expect: 100-continue
...
-   examining just the request-line and header fields, or send an immediate
+   examining just the method, target URI, and header fields, or send an immediate
    <x:ref>100 (Continue)</x:ref> response to encourage the client to send the
    request's message body. The origin server &MUST-NOT; wait for the message
    body before sending the <x:ref>100 (Continue)</x:ref> response.
 </t>
 <t>
-   A proxy &MUST;, upon receiving an HTTP/1.1 (or later) request-line
-   and a complete header section that contains a 100-continue expectation and
+   A proxy &MUST;, upon receiving an HTTP/1.1 (or later) request that has a method, target URI
```
annevk, line 5393:
```suggestion
   A proxy &MUST;, upon receiving an HTTP/1.1 (or later) request that has a method, target URI,
```

MikeBishop:
Pseudo-header fields are required to come first in H2 and H3, so arguably there's a "pseudo-header" which contains the information which H1 carries in the request-line / status-line.  But it's still part of the header field section, because it gets compressed.

mnot:
Why is it important to distinguish the layout of the syntax in a semantics document, especially when there's such variance between versions?

reschke:
My proposal is not to *distinguish* the syntax, but to have a well-defined name for the set of information that previously was contained in request/status-line. That was already problematic in 1.1 (due to Host:), and with 2 it's even more complex.

mnot:
For requests, it's method + request target; for responses, it's status code + optional status phrase.

Looking at the diff, the amount of overhead in listing these is negligible. Much larger, I think, would be the cognitive overhead of introducing yet another abstract term to readers. I really don't think it's the right thing to do.

reschke:
> For requests, it's method + request target; for responses, it's status code + optional status phrase.

s/request target/target resource/ (or target URI)

mnot:
@royfielding ?

PR #363, opened 2020-05-06
Qualify requirement on applying conditionals

### LINKED ISSUE #358: Evaluation of conditional headers
In Evaluation, we say:

Likewise, a server &MUST; ignore the conditional request header fields
defined by this specification when received with a request method that does
not involve the selection or modification of a
<x:ref>selected representation</x:ref>, such as CONNECT, OPTIONS, or TRACE.
It seems like we need to reference that if we're placing requirements around methods later on.

mnot:
based upon existing requirement in evaluation.

I think this is editorial, because it's just referencing an already-existing requirement there.

Fixes #358.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -5687,7 +5687,7 @@ Expect: 100-continue
 </t>
 <t>
    An origin server that receives an If-Match header field &MUST; evaluate the
-   condition prior to performing the method (<xref target="evaluation"/>).
+   condition prior to performing the method (when applicable; see <xref target="evaluation"/>).
 </t>
 <t>
    To evaluate a received If-Match header field:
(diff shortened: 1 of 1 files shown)


mnot:
@royfielding because a casual reader will not realise that "evaluate the condition" includes tests that might mean they shouldn't proceed.

royfielding:
That sentence does not compute. It isn't for casual readers and it has no relevance to people who don't implement that evaluation. If they don't understand it, how are they supposed to understand what applicable means?

mnot:
"evaluate the condition" is imperative; it doesn't imply that it might not apply in some cases. Coupling it with a MUST and only incorporating the applicability by reference as a parenthetical aside with no obvious connection to the current text is misleading.

This is my preferred way of addressing that. We could also address it in other ways; e.g., instead of 

> An origin server that receives an If-Match header field must evaluate the condition prior to performing the method (Section 8.2.1).

say

> An origin server that receives an If-Match header field must evaluate the condition as per Section 8.2.1 prior to performing the method.

mnot:
See updated text. 

Aside - why is this requirement only a SHOULD for IMS, when it's a MUST for all other conditionals?

royfielding:
> Aside - why is this requirement only a SHOULD for IMS, when it's a MUST for all other conditionals?

Because implementing IMS was optional -- sending a 200 isn't an error case.

PR #367, opened 2020-05-20
Add note about URL resolution to C-L and Referer

### LINKED ISSUE #321: Being explicit about relative URIs
2616's definition of `Referer` said:

> If the field value is a relative URI, it SHOULD be interpreted relative to the Request-URI. 

7231 dropped that, thanks to [this commit](https://trac.ietf.org/trac/httpbis/changeset/2072#file3).

We should probably restore it, given it can cause misconceptions like [this](https://github.com/npm/cli/issues/836#issuecomment-589404272).

mnot:
Text taken from existing statement in Location.

Fixes #321.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -12529,10 +12529,10 @@ Content-Encoding: gzip
   <li>Throughout, replace "effective request URI", "request-target" and similar with "target URI" (<eref target="https://github.com/httpwg/http-core/issues/259"/>)</li>
   <li>In <xref target="considerations.for.new.header.fields"/> and <xref target="considerations.for.new.status.codes"/>, describe how extensions should consider scope of applicability (<eref target="https://github.com/httpwg/http-core/issues/265"/>)</li>
   <li>In <xref target="operation"/>, don't rely on the HTTP/1.1 Messaging specification to define "message" (<eref target="https://github.com/httpwg/http-core/issues/311"/>)</li>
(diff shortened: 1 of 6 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3652,6 +3652,12 @@ bytes=500-700,601-999
 <sourcecode type="abnf7230"><iref primary="true" item="Grammar" subitem="Content-Location"/>
   <x:ref>Content-Location</x:ref> = <x:ref>absolute-URI</x:ref> / <x:ref>partial-URI</x:ref>
 </sourcecode>
+<t>
+   The field value consists of a single URI-reference. When it has the form
```
royfielding, line 3656: I don't think this is correct. Content-Location is not defined as a URI-reference because we don't allow it to contain a fragment. So, I think his needs to say "absolute-URI or partial-URI" or "URI reference excluding fragment".

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7011,6 +7017,12 @@ Expect: 100-continue
 <sourcecode type="abnf7230"><iref primary="true" item="Grammar" subitem="Referer"/>
   <x:ref>Referer</x:ref> = <x:ref>absolute-URI</x:ref> / <x:ref>partial-URI</x:ref>
 </sourcecode>
+<t>
+   The field value consists of a single URI-reference. When it has the form
```
royfielding, line 7021: Likewise, not allowed to contain a fragment due to privacy/security disclosure.

mnot:
Ah, makes sense, of course. Thanks @reschke

reschke:
@royfielding - can you check again?

Issue #193, opened 2019-01-20
Audit: single value header field error handling


mnot:
As discussed in #74, we need to assure that all single-value headers we define have appropriate error handling for when more then one value is received. This could be a variety of things, including "it's not important for this header."

- [ ] Authorization
- [ ] Content-Language
- [x] Content-Length
- [x] Content-Location
- [ ] Content-Range
- [x] Content-Type
- [ ] Date
- [ ] ETag
- [x] Expect
- [x] From
- [ ] Host
- [ ] If-Modified-Since
- [ ] If-Range
- [ ] If-Unmodified-Since
- [ ] Last-Modified
- [ ] Location
- [ ] Max-Forwards
- [ ] Range
- [x] Referer
- [x] Retry-After
- [x] Server
- [x] User-Agent
- [ ] Proxy-Authorization
- [x] Age
- [ ] Expires

annevk:
There's #59 on `Content-Length`. 😊 (I can confirm some clients at least have quite special handling for `Location`, but I haven't written exhaustive tests yet.)

What I'd like to see is that
```
Single-Value-Header: x
Single-Value-Header: y
```
and
```
Single-Value-Header: x, y
```
end up being treated in equivalent manner. Otherwise a naïvely-combining intermediary in front of some clients can cause drastic different results from clients that don't have such an intermediary in front of them.

mnot:
`Expect` is an interesting case. In the past, it allowed multiple values, but now it only allows a single value, `100-continue`. Error handing for it seems to be well-defined if wishy-washy (with a MAY reject).

mnot:
Ticking off Expect due to #203

mnot:
Starting to gather some data about Age [here](https://cache-tests.fyi/#age-parse). It seems like most implementations honour the last value if there are more than one. Safari and Apache look like they consider the comma as making the Age invalid (and therefore not present).

mnot:
Niiiiice... It appears that most implementations will pay attention to the last Age value if they appear on the same header line, but the first header line if there are two.

royfielding:
I am kind of disappointed that they don't treat the comma as a thousands indicator.

annevk:
@mnot that's fairly consistent with other headers. The way parsers in browsers work is that they are per header-line. And it's considered risky to change those such that they effectively work on a combined value. With more tests and encouragement and maybe effectively making it impossible to have this distinction in future wire formats this can be overcome somehow.

mnot:
[...]

PR #369, opened 2020-05-20
 Intent is to make all headers list-based.

mnot:
A start on #193.


royfielding:
Umm, no way. Uh uh. -1.

The ABNF is for generating fields. Singletons are an instruction not to generate more than one. It is impossible, right now, to make that change and expect conforming implementations to magically adapt. Multiple instances of singleton field values are non-interoperable and will remain so for the life of HTTP.

A parser implementation might want a set of robustness rules for multiple field handling, regardless of their generative definition. But the security of those rules are going to depend on the semantics of the field (e.g., Host, Auth, C-L).

PR #373, opened 2020-05-25
Prohibit bare CRs in field values and sections.

### LINKED ISSUE #31: Newline parsing in H1
Popular clients handle LF, CR, and CRLF as newlines. HTTP only defines CRLF.

Previously: https://lists.w3.org/Archives/Public/ietf-http-wg/2014JulSep/0123.html.

mnot:
Fixes #31.

### DIFF
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -3414,6 +3414,10 @@ Upgrade: websocket
    connection management requirements specific to HTTP/1.1.
 </t>
 <t>
+   Prohibited generation of bare CRs outside of payload body.
+   (<xref target="message.parsing"/>)
+</t>
+<t>
    In the ABNF for chunked extensions, re-introduced (bad) whitespace around
    ";" and "=". Whitespace was removed
    in <xref target="RFC7230"/>, but that change was found to break existing
@@ -3435,10 +3439,6 @@ Upgrade: websocket
    Disallowed transfer coding parameters called "q" in order to avoid
    conflicts with the use of ranks in the <x:ref>TE</x:ref> header field.
    (<xref target="transfer.coding.registry"/>)
-</t>
-<t>
-   Prohibited generation of bare CRs outside of payload body.
-   (<xref target="message.parsing"/>)
 </t>
 </section>
 </section>
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -316,6 +316,12 @@
    single LF as a line terminator and ignore any preceding CR.
 </t>
 <t>
+  One or more CR characters not succeeded by LF are "bare CRs", and
+  &MUST-NOT; be generated; recipients &MUST; either reject a field section
+  containing bare CRs, or convert them to SP before processing or forwarding
+  the message.
+</t>
+<t>
    Older HTTP/1.0 user agent implementations might send an extra CRLF
    after a POST request as a workaround for some early server
    applications that failed to read message body content that was
@@ -3416,6 +3422,10 @@ Upgrade: websocket
    conflicts with the use of ranks in the <x:ref>TE</x:ref> header field.
    (<xref target="transfer.coding.registry"/>)
 </t>
+<t>
+   Prohibited the generation of bare CRs in field sections.
+   (<xref target="message.parsing"/>)
+</t>
 </section>
 </section>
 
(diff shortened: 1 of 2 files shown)


mnot:
@reschke the ABNF is not the source of truth; as discussed, it's the recommended generated form. We have prose in [error handling](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#error-handling) that we use to do things like explicitly allow LFs to be separators and folding to be performed. Given that there is confusion, interoperability issues and security issues around bare CRs, a clear requirement seems proportional - especially when the ABNF does not specify error handling.

I could see an argument to downgrade the requirement relating to generation to a note that it's not allowed, and I could see broadening the language in semantics to say that CRLF and LF are also not allowed.

reschke:
> I could see an argument to downgrade the requirement relating to generation to a note that it's not allowed, and I could see broadening the language in semantics to say that CRLF and LF are also not allowed.

For field values (Semantics), we clarify that these are illegal to generate. For recipients, we can say that for error handling, rejecting or converting to SP is common.

Not sure what else needs to be said in Messaging.

mnot:
> For field values (Semantics), we clarify that these are illegal to generate. For recipients, we can say that for error handling, rejecting or converting to SP is common.

I think that the security aspect justifies a new requirement; how about "MUST either reject or remove; suggest replacing with SP"?
 
> Not sure what else needs to be said in Messaging.

I was thinking about outside the field values -- mostly in the request-line and status-line. That said, I agree it's much less of a concern there, and I'm OK with dropping it if you think it's not worth it.

I'll update the PR if you think the above is workable.

reschke:
s/MUST either reject or remove/MUST either reject or replace/

...I guess. Yes, that sounds right to me; although there might be pusback on the MUST (me neutral).

The part for Messaging I believe deserves a separate issue and discussion.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -315,6 +315,12 @@
    fields is the sequence CRLF, a recipient &MAY; recognize a
    single LF as a line terminator and ignore any preceding CR.
 </t>
+<t>
+  One or more CR characters not succeeded by LF are "bare CRs", and
+  &MUST-NOT; be generated; recipients &MUST; either reject a field section
+  containing bare CRs, or convert them to SP before processing or forwarding
+  the message.
```
royfielding, line 322: This should be:

```
   A sender MUST NOT generate a bare CR (a CR character not immediately succeeded
   by LF) within any protocol elements other than the payload body. A recipient of such a
   bare CR MUST either reject that element as invalid or replace each bare CR with SP
   before processing the element or forwarding the message.
```

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -3416,6 +3422,10 @@ Upgrade: websocket
    conflicts with the use of ranks in the <x:ref>TE</x:ref> header field.
    (<xref target="transfer.coding.registry"/>)
 </t>
+<t>
+   Prohibited the generation of bare CRs in field sections.
```
royfielding, line 3426: Prohibited generation of bare CR outside of payload body.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1429,6 +1429,11 @@ deployed and not likely to be registered in a timely manner otherwise.</t>
    A recipient &SHOULD; treat other octets in field content (obs&nbhy;text) as
    opaque data.
 </t>
+<t>
+  Field values containing one or more CR characters ("bare CRs") &MUST-NOT;
+  be generated; recipients &MUST; either reject a field value containing bare
+  CRs, or convert them to SP before processing or forwarding the message.
```
royfielding, line 1435:
```
   A sender MUST NOT generate a bare CR (a CR character not immediately succeeded
   by LF) within a field value. A recipient of such a bare CR MUST either reject that field
   value as invalid or replace each bare CR with SP before processing or forwarding the field.
```

mnot: Applied with tweaks.

mnot:
@reschke I'll make a change note in Messaging.

WRT Semantics, the adjacent text about folding effectively deals with LF and CRLF.

If you want to go to the trouble of separating into two issues and two pull requests, feel free. I'd rather just get it closed.

mnot:
/me looks - @reschke it already is called out. Please re-review.

reschke:
On my agenda for today.

reschke:
opened #377 for field values

reschke:
Splitting the issue was king of intrusive, sorry for that.

The changes for [Messaging] have been merged.

For [Semantics] please see https://github.com/httpwg/http-core/pull/379

PR #376, opened 2020-05-28
Add "Comments" to HTTP field registry, and fix entry for "Close" to be "standard" with comment "(reserved)" (#273) 

reschke:


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1347,56 +1347,79 @@ Content-Type: text/plain
...
+<t>
+   Provisional entries can be removed by the Expert(s) if &mdash; in consultation
+   with the community &mdash; the Expert(s) find that they are not in use. The
+   Experts can change a provisional entry's status to permanent at any time.
+</t>
+<t>
+   Note that names can be registered by third parties (including the
+   Expert(s)), if the Expert(s) determines that an unregistered name is widely
+   deployed and not likely to be registered in a timely manner otherwise.
+</t>
```
mnot, line 1422: This should really be separate :)

reschke: separate from what? (there's no change here to what we have right now)

reschke:
@royfielding - can I proceed with this one?

PR #378, opened 2020-06-03
Prohibit bare CRs in header section

reschke:
changes to -messaging pulled from https://github.com/httpwg/http-core/pull/373


reschke:
I'll go ahead with this one as we already reached consensus in the original PR.

Issue #377, opened 2020-06-03
prohibit vertical whitespace in field values


reschke:
(forked from #31)

PR #379, opened 2020-06-03
prohibit vertical whitespace in field values (#377)

reschke:
this is the part of https://github.com/httpwg/http-core/pull/373 related to -semantics

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -1454,9 +1454,10 @@ Content-Type: text/plain
    opaque data.
 </t>
 <t>
-   Field values containing one or more CR characters ("bare CRs") &MUST-NOT;
-   be generated; recipients &MUST; either reject a field value containing bare
-   CRs, or convert them to SP before processing or forwarding the message.
+   Field values containing control (<x:ref>CTL</x:ref>) characters such as
+   CR or LF are invalid; recipients &MUST; either reject a field value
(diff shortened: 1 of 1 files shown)


reschke:
Continueing the discussion from https://github.com/httpwg/http-core/pull/373#issuecomment-636603232:

> WRT Semantics, the adjacent text about folding effectively deals with LF and CRLF.

Looking at the next text in context:

> Historically, HTTP allowed field content with text in the ISO‑8859‑1 charset [ISO-8859-1], supporting other charsets only through use of [RFC2047] encoding. In practice, most HTTP field values use only a subset of the US-ASCII charset [USASCII]. Newly defined fields SHOULD limit their values to US‑ASCII octets. A recipient SHOULD treat other octets in field content (obs‑text) as opaque data.

> Field values containing one or more CR characters ("bare CRs") MUST NOT be generated; recipients MUST either reject a field value containing bare CRs, or convert them to SP before processing or forwarding the message.

> Leading and trailing whitespace in raw field values is removed upon field parsing (Section 5.1 of [Messaging]). Field definitions where leading or trailing whitespace in values is significant will have to use a container syntax such as quoted-string (Section 4.4.1.2).

reschke:
Proposal:

> Field values containing control (CTL) characters such as CR or LF are invalid; recipients MUST either reject a field value containing control characters, or convert them to SP before processing or forwarding the message.

mnot:
Fixes #377

reschke:
@royfielding ?

Issue #48, opened 2018-03-05
415 and Accept


martinthomson:
This came up in ACME where they wanted to use 415 and Accept.  

RFC 7694 implies this is OK, and makes two observations:

* 415 is for cases where request content type or content coding is not acceptable
* you can use Accept-Encoding with 415

RFC 7694 failed to explicitly allow Accept with 415, though it naturally follows.

We might consider rolling 7694 into any update given its tiny size, and then fix this in the process.

reschke:
Allowing "Allow" sounds right to me.

Whether we want to make this in scope of http11ter we'll need to discuss...

reschke:
Opened #119 to discuss inclusion of RFC 7694.

If we decide not to do this, should we update 7694 instead?

reschke:
Right.

reschke:
@martinthomson - what was the outcome on the ACME side?

martinthomson:
I think that they avoided it:

> If a request does not meet this requirement, then the server MUST return a response with status code 415 (Unsupported Media Type).

With no mention of Accept.  I don't know what Boulder does, but can probably find out if absolutely necessary.

reschke:
@martinthomson - that would be great.

jsha:
In Boulder, if a POST doesn't have `Content-Type: application/json`, we return 415 with no Accept header.

spaceone:
I don't understand the initial issue here fully but it would be nice If HTTP would have something to indicate in a response which response payloads it provides, e.g. using `Accept: application/json; q=1.0, text/html; q=0.9` in a response (e.g. to a `OPTIONS` request)!

And what also would be nice is if the resource could also inform the client which media types it supports in the request (as JSON has no forms to include this information and HTML only allows forms with `application/x-www-form-urlencoded` and `multipart/form-data` there is no standardized way to tell clients the possibilities).

mnot:
Discussed in Basel; ACME apparently doesn't need this any more. Waiting on resolution of #119.

mnot:
@reschke do you wan to take this one?

reschke:
I can give it a try.

PR #382, opened 2020-06-05
Allow "Accept" in response messages (#48)

reschke:


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8665,12 +8670,21 @@ Content-Range: bytes 7000-7999/8000
...
+</t>
+<t>
+   If the problem was caused by an unsupported content coding, the
+   <x:ref>Accept-Encoding</x:ref> response header field
+   (<xref target="field.accept-encoding"/>) ought to be
    used to indicate what (if any) content codings would have been accepted
    in the request. 
 </t>
+<t>
+   On the other hand, it the cause was an unsupported media type, the
```
royfielding, line 8683:
```suggestion
   On the other hand, if the cause was an unsupported media type, the
```

spaceone:
Thank you!

PR #383, opened 2020-06-11
Singleton fields

mnot:
Rough in text defining singleton fields and show what it would look like for Age; more to come.

For #193.


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1460,10 +1462,18 @@ Content-Type: text/plain
...
+   fields."
+</t>
+<t>
+   Because commas are used as a generic delimiter between members, they need
+   to be treated with care if they are allowed as data within a member. This
+   is true for both list-based and singleton fields, as a singleton field
+   might be sent with multiple members erroneously, and a recipient needs to
+   be able to detect this condition. Therefore, members that might contain a
+   comma are typically enclosed in a quoted-string.
 </t>
```
reschke, line 1477: "recipient needs to  be able to detect this condition" - that makes it sound as if this is requirement; unfortunately it is not (see, for instance "Location"). So this may need slightly rephrasing.

royfielding: I still don't see the value in requiring specific behavior on invalid messages unless we have identified an interop failure that needs a consistent resolution. Age does not. I'd either take the max value or a sum, or maybe just mark it stale, but it depends on why I have the cache.

mnot: I changed to 'should be considered stale'; that makes sense. I don't think we can say that they should be added or that the largest value should be used, because it's not clear what the non-conformant upstream implementation is doing.

mnot: rephrased

mnot:
Work list:

- [ ] Authorization
- [ ] Content-Language
- [ ] Content-Length
- [ ] Content-Location
- [ ] Content-Range
- [ ] Content-Type
- [ ] Date
- [ ] ETag
- [ ] Expect
- [ ] From
- [ ] Host
- [ ] If-Modified-Since
- [ ] If-Range
- [ ] If-Unmodified-Since
- [ ] Last-Modified
- [ ] Location
- [ ] Max-Forwards
- [ ] Range
- [ ] Referer
- [ ] Retry-After
- [ ] Server
- [ ] User-Agent
- [ ] Proxy-Authorization
- [x] Age
- [ ] Expires

mnot:
I've taken a fairly conservative approach here; the ticked off items above are either addressed by this PR, are already taken care of, or have a separate issue.

@annevk are there any of the unticked items that are especially thorny for interoperability or security from your standpoint?

annevk:
`Location` is problematic and not fully interoperable. It's also not clear we could combine before parsing there. (Range requests are too, but not the headers as far as I know.)

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1486,6 +1486,8 @@ Content-Type: text/plain
   <x:anchor-alias value="field-value"/>
   <x:anchor-alias value="field-content"/>
   <x:anchor-alias value="field-vchar"/>
+  <x:anchor-alias value="field.singleton"/>
+  <x:anchor-alias value="field.list"/>
```
reschke, line 1490: If would actually make these "singleton.field" and "list.field".

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1524,10 +1526,18 @@ Content-Type: text/plain
    as quoted-string (<xref target="quoted.strings"/>).
 </t>
 <t>
-   Because commas (",") are used as a generic delimiter between members of a
-   list-based field value, they need to be treated with care if they are
-   allowed as data within those members. Typically, list members that might contain a
-   comma are enclosed in a quoted-string.
+   A field's value can contain on or more members, separated by commas (",").
```
reschke, line 1529: This might require rephrasing, as not every occurence of "," inddicates multiple members.

mnot:
@annvk I've added text regarding `Location`:

> A Location field value that has multiple comma-separated members poses difficulties for recovery due to the value's syntax; as a result, recovery in that situation is not interoperable across implementations.

draft-ietf-httpbis-semantics-latest.xml
mnot: The paragraph directly below addresses this.

annevk:
Thanks. I'm still hopeful that will get that resolved some day.

draft-ietf-httpbis-semantics-latest.xml
reschke: I still think that the initial statement is misleading, if not even incorrect. Wil think about alternative text.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2848,8 +2858,8 @@ Host: www.example.org
 </t>
 <t>
    A server &MUST; respond with a <x:ref>400 (Bad Request)</x:ref> status code
-   to any HTTP/1.1 request message that lacks a Host header field and
-   to any request message that contains more than one Host header field
+   to any HTTP/1.1 request message that lacks a Host header field,
+   to any request message that whose Host header field value has more than one member,
```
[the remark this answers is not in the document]
royfielding: If desired, the last sentence should add a parenthetical to invalid:
```
   or a Host header field with an invalid field value (including any value that
   constitutes a list of members).
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6450,7 +6460,8 @@ Expect: 100-continue
 </t>
 <t>
    A recipient &MUST; ignore the If-Unmodified-Since header field if the
-   received field value is not a valid HTTP-date.
+   received field value is not a valid HTTP-date, or if the field value has
+   more than one member.
 </t>
```
royfielding, line 6465: This looks like a bug in the old spec. If-unmodified-since should behave like If-match and cause a 412 when it is invalid (because that doesn't match).

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9645,6 +9656,9 @@ Content-Type: text/plain
     &Note; Some recipients attempt to recover from Location fields
     that are not valid URI references. This specification does not mandate or
     define such processing, but does allow it for the sake of robustness.
+    A Location field value that has multiple comma-separated members
+    poses difficulties for recovery due to the value's syntax; as a result,
+    recovery in that situation is not interoperable across implementations.
```
royfielding, line 9661:
```suggestion
    A Location field value cannot allow a list of members because the comma list separator
    is a valid data character within a URI-reference. If an invalid message is sent with multiple
    Location field instances, a recipient along the path might combine the field instances into
    one value. Recovery of a valid Location field value from that situation is difficult and not
    interoperable across implementations.
```

draft-ietf-httpbis-semantics-latest.xml
mnot: Open a separate issue?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2848,8 +2859,8 @@ Host: www.example.org
 </t>
 <t>
    A server &MUST; respond with a <x:ref>400 (Bad Request)</x:ref> status code
-   to any HTTP/1.1 request message that lacks a Host header field and
-   to any request message that contains more than one Host header field
+   to any HTTP/1.1 request message that lacks a Host header field,
+   to any request message that whose Host header field value has more than one member,
```
mnot, line 2863:
```suggestion
   or a Host header field with an invalid field value (including a value that constitutes a list of members).
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2859,7 +2860,7 @@ Host: www.example.org
 <t>
    A server &MUST; respond with a <x:ref>400 (Bad Request)</x:ref> status code
    to any HTTP/1.1 request message that lacks a Host header field,
-   to any request message that whose Host header field value has more than one member,
+   or a Host header field with an invalid field value (including a value that constitutes a list of members).
    or a Host header field with an invalid field value.
```
royfielding, line 2864:
```suggestion
```

draft-ietf-httpbis-semantics-latest.xml
royfielding: On second thought (and seeing it has always been specified as ignore on invalid), let's just keep it that way.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1605,10 +1607,19 @@ Content-Type: text/plain
...
 </t>
 <t>
-   Because commas (",") are used as a generic delimiter between members of a
-   list-based field value, they need to be treated with care if they are
-   allowed as data within those members. Typically, list members that might contain a
-   comma are enclosed in a quoted-string.
+   A field's value often contains one or more members, separated by commas (",").
+   A field that allows multiple members is referred to as a "list-based" field.
+   A field that only allows a single member is referred to as a "singleton" field.
+</t>
```
reschke, line 1613: Maybe change the first sentence to:

"Commas (",") often are used to separate members in field values."

?

mnot: I don't think that effectively introduces the concept of members of a list.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1243,7 +1243,9 @@
 </sourcecode>
 <t>
    The Age field value is a non-negative integer, representing time in seconds
-   (see <xref target="delta-seconds"/>).
+   (see <xref target="delta-seconds"/>). A cache &SHOULD; consider a response
+   to be stale if an Age field is present and its value is invalid (i.e.,
+   contains a list or something other than a non-negative integer).
 </t>
```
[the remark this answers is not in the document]
mnot: Done.

draft-ietf-httpbis-semantics-latest.xml
reschke: Ah, I didn't realize that this was supposed to introduce a term/concept. That makes it even more important to get it right. Consider:

    If-Match: "a, b"

or

    Location: foo,bar

In this case, do we really call '"a' and 'b"' (or 'foo' and 'bar') "members of a list"?

mnot: I see your point; will revise.

Issue #44, opened 2017-12-12
Extension capabilities


mnot:
We're seeing some _interesting_ use of extensions starting to be discussed, e.g.,:

* Overriding method semantics using a HTTP/2 setting (in the WebSockets draft)
* Referring to pseudo-headers as primary artefacts in HTTP header payloads (in the Signature proposal)

Some guidance about the relationship of version-specific syntax/extensions and "top-level" HTTP semantics would be helpful.

mnot:
I'm not sure what "h2 text in core" we're waiting for here; I think our goal for Semantics is to have as little version-specific text as possible. @mcmanus do you remember any more detail?

royfielding:
I think this is something we need to review for Versioning just prior to WGLC.

MikeBishop:
Looking at [the minutes of that meeting](https://github.com/httpwg/wg-materials/blob/gh-pages/ietf102/minutes.md#extension-capabilities-44), we had "an open issue to put text around the semantic issues around h/2," which I read as describing what an extension to a mapping (or perhaps the mapping itself, including its extensions) can and cannot modify about HTTP semantics.  That was presumably #50.

But then we closed #50 as a duplicate of this issue, so that's this issue again.  There was a PR for #50 which was closed by #342, which is already merged.  Presumably the text we were waiting for was in #342, and if it wasn't, we need to do it here.

mnot:
Perhaps we need to say that a version-specific extension (like SETTINGS or a frame type) can't override or modify the semantics of a version-independent mechanism or extension point (like a method or header) without explicitly being allowed by that protocol element (e.g., CONNECT does this implicitly, we should make it explicit).

mnot:
I think the above would go into a new section, "Extending HTTP", along with references to the various extension points.

That and "Protocol Versioning" (currently in 3. Conformance) might fit best in a new top-level section, "Versioning and Extending HTTP." Probably between Architecture and Conformance?

PR #384, opened 2020-06-11
Rough in discussion of extensibility

mnot:
For #44


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1067,6 +1067,56 @@ Content-Type: text/plain
...
+   expression of them "on the wire" can change, and so the
+   HTTP version number changes when incompatible changes are made to the wire
+   format. Additionally, HTTP allows incremental, backwards-compatible
+   changes to be made to the protocol without changing its version through
+   the use of defined extension points.
+</t>
+
+<section title="Extending HTTP" anchor="extending">
+<t>
+   HTTP defines a number over generic extension points that can be used to
```
mnot, line 1085:
```suggestion
   HTTP defines a number of generic extension points that can be used to
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1067,6 +1067,56 @@ Content-Type: text/plain
...
+<section title="Extending HTTP" anchor="extending">
+<t>
+   HTTP defines a number of generic extension points that can be used to
+   introduce capabilities to the protocol without introducing a new version,
+   including methods (<xref target="method.extensibility"/>), status codes
+   (<xref target="status.code.extensibility"/>), header and trailer fields
+   (<xref target="considerations.for.new.header.fields"/>), and further
+   extensibility points within defined fields (such as Cache-Control in <xref
+   target="cache.control.extensions"/>). Because the semantics of HTTP are
+   not versioned, these extension points are constant; the version of the
```
mnot, line 1092:
```suggestion
   not versioned, these extension points are persistent; the version of the
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1067,6 +1067,56 @@ Content-Type: text/plain
...
+</t>
+<t>
+   Additionally, specific versions of HTTP might have their own extensibility
+   points, such as transfer-codings in HTTP/1.1 (<xref
+   target="field.transfer-encoding"/>) and HTTP/2 (<xref target="RFC7540"/>)
+   SETTINGS or frame types. These extension points are specific to the
+   version of the protocol they occur within.
+</t>
+<t>
+   Version-specific extensions are cannot override or modify the semantics of
```
royfielding, line 1109:
```suggestion
   Version-specific extensions cannot override or modify the semantics of
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5032,6 +5082,11 @@ Content-Range: exampleunit 11.2-14.3/25
    through one or more proxies, which can then be secured using TLS
    (Transport Layer Security, <xref target="RFC8446"/>).
 </t>
+<t>
+   Because CONNECT changes the request/response nature of a HTTP connection,
```
royfielding, line 5086:
```suggestion
   Because CONNECT changes the request/response nature of an HTTP connection,
```

royfielding: Beyond that, I don't think this paragraph says anything useful. What does the reader need to know here?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5032,6 +5082,11 @@ Content-Range: exampleunit 11.2-14.3/25
    through one or more proxies, which can then be secured using TLS
    (Transport Layer Security, <xref target="RFC8446"/>).
 </t>
+<t>
+   Because CONNECT changes the request/response nature of an HTTP connection,
+   specific HTTP versions might have different ways of mapping its semantics
+   into the protocol's wire format.
+</t>
```
royfielding, line 5089: I think you missed my second comment that this paragraph doesn't say anything useful. AIUI, CONNECT was recently extended to allow an origin server to process it, which wasn't about the wire format.

I would prefer that we update the first paragraph to be less specific to 1.1 and then reference the CONNECT extension as an example of an allowed semantic extension because the method had no applicability for an origin server before.

I am also fine with merging this now and fixing CONNECT later.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1067,6 +1067,56 @@ Content-Type: text/plain
...
+   Additionally, specific versions of HTTP might have their own extensibility
+   points, such as transfer-codings in HTTP/1.1 (<xref
+   target="field.transfer-encoding"/>) and HTTP/2 (<xref target="RFC7540"/>)
+   SETTINGS or frame types. These extension points are specific to the
+   version of the protocol they occur within.
+</t>
+<t>
+   Version-specific extensions cannot override or modify the semantics of
+   a version-independent mechanism or extension point (like a method or
+   header) without explicitly being allowed by that protocol element. For
```
reschke, line 1111: s/header/header field/

reschke, line 1111:
```suggestion
   header field) without explicitly being allowed by that protocol element. For
```

draft-ietf-httpbis-semantics-latest.xml
MikeBishop: RFC 7231 already permits origin servers to process CONNECT.

> An origin server that receives a CONNECT request for itself MAY respond with a 2xx (Successful) status code to indicate that a connection is established.  However, most origin servers do not implement CONNECT.

But yes, that extension stretches the bounds of this description.

PR #387, opened 2020-06-23
semantics should only informatively reference messaging

mnot:
Work in progress. `Upgrade` is placed in "Message Routing"; once #384 lands, it should probably move into the protocol versioning and extensibility section.

For #335.


mnot:
OK, I think this is ready for first review, although it's still somewhat rough.

mnot:
> I don't like the changes to Connection. An h2/h3 proxy (if there is such a thing) needs to process those fields on the way through even if they shouldn't exist in h2/h3.

I think that's an argument for moving Connection into Semantics. Make sense?

> Process-wise, I don't think it is true that Semantics can reference Messaging as Informative and still form a cohesive replacement for 723x on the standards track for HTTP/1.1. I would prefer the reference remain normative (meaning readers should read it to understand HTTP) even if we don't use it in a normative requirement.

I don't know what it means to 'form a cohesive replacement for 723x'. What's left in Messaging that readers need to know to understand the semantics of HTTP?

reschke:
> I don't know what it means to 'form a cohesive replacement for 723x'. What's left in Messaging that readers need to know to understand the semantics of HTTP?

Are we talking about what replaces what? In that case isn't it more relevant what we say in "obsoletes" and "updates"?

Right now, we have:

- Caching obsoletes RFC 7234
- Messaging obsoletes RFC 7230
- Semantics obsoletes RFC 7230, 7231, 7232, 7233, 7235

That looks roughly right to me.

That said, I don't have a problem keeping the Messaging reference normative.

Would it make sense to split the changes for the individual header fields into separate sub tasks?

reschke:
> Any plans to merge RFC 7239 into semantics? It seems to me just as relevant and useful these days as documenting Via there.

The difference is that Via has always been here. We can't mrge *everything* into th core specs.

royfielding:
>  don't know what it means to 'form a cohesive replacement for 723x'. What's left in Messaging that readers need to know to understand the semantics of HTTP?

I mean that when the RFC editor prepares a document set to be "HTTP/1.1", they expect the set to be cohesive and normative. They don't like mixing in Informational references as the normative set. In any case, we do refer to messaging many times in a normative way right now, even after these changes, and I see no reason to change the reference.

royfielding:
I picked out the moves of Upgrade and TE into #390 so that they can be reviewed and merged separately.

MikeBishop:
> I mean that when the RFC editor prepares a document set to be "HTTP/1.1", they expect the set to be cohesive and normative. They don't like mixing in Informational references as the normative set

I agree the set should be cohesive, but it doesn't follow that all references within the set have to be normative.  In particular, I don't think there's a problem with saying that Messaging normatively depends on Semantics, while Semantics only informatively references the Messaging document; the normative dependency is simply one-way.  That seems ideal, if not necessarily easy.

PR #397, opened 2020-07-08
Adjust "beginning with... followed by" language

### LINKED ISSUE #372: terminology/definition of where request target and response status live
Section 2.1 has:

> A client sends an HTTP request to a server in the form of a request message, beginning with a method (Section 7) and URI, followed by header fields containing request modifiers, client information, and representation metadata (Section 4), and finally a payload body (if any, Section 6.3.3).
> 
> A server responds to a client's request by sending one or more HTTP response messages, each beginning with a success or error code (Section 9), possibly followed by header fields containing server information, resource metadata, and representation metadata (Section 4), and finally a payload body (if any, Section 6.3.3).

So this is also specific to 1.1.

My proposal would be to define terms for request-line and status-line that are indenpendant of the protocol version.

For 1.1, these would map to what we had before (and might have to include the "Host" header field, right?).

For 2 (and 3, I guess), they would map to the pseudo header fields (all?).

We'd also have to fix the text above, as some ("Host") or all (pseudo header fields) are actually contained in the header field section.

mnot:
Fixes #372

Didn't add change notes as this seems very editorial.


royfielding:
```
<t>
   A client sends an HTTP request to a server in the form of a
   <x:dfn>request</x:dfn> message with a method (<xref
   target="methods"/>) and request target. The request might also
   contain header fields for request modifiers, client information, and
   representation metadata (<xref target="header.and.trailer.fields"/>),
   a payload body (<xref target="payload.body"/>) to be processed
   in accordance with the method, and trailer fields for metadata
   collected while sending the payload.
</t>
<t>
   A server responds to a client's request by sending one or more HTTP
   <x:dfn>response</x:dfn> messages, each including a success or error
   code (<xref target="status.codes"/>). The response might also contain
   header fields for server information, resource metadata, and representation
   metadata (<xref target="header.and.trailer.fields"/>),
   a payload body (<xref target="payload.body"/>) to be interpreted
   in accordance with the status code, and trailer fields for metadata
   collected while sending the payload.
</t>
```

mnot:
@royfielding I'm happy with that if @reschke is.

mnot:
@mnot to incorporate and merge.

PR #398, opened 2020-07-08
State that multiple "*" has no defined semantics

### LINKED ISSUE #305: fix ABNF for If-Match and If-None-Match
...not to special-case "*" as top-level construct.

mnot:
... for If-Match and If-None-Match.

Fixes #305.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -5786,6 +5786,12 @@ Expect: 100-continue
    The If-Match header field can be ignored by caches and intermediaries
    because it is not applicable to a stored response.
 </t>
+<t>
+   Note that an If-Match header field with a list value containing "*" and
+   other values (including other instances of "*") does not have defined
+   semantics, and such field values are not interoperable. This is a consequence
+   of the field being specified as a choice between a single and multiple values.
+</t>
 </section>
 
 <section title="If-None-Match" anchor="field.if-none-match">
@@ -5867,6 +5873,12 @@ Expect: 100-continue
 <t>
    Requirements on cache handling of a received If-None-Match header field
    are defined in <xref target="validation.received"/>.
+</t>
+<t>
+   Note that an If-None-Match header field with a list value containing "*" and
+   other values (including other instances of "*") does not have defined
+   semantics, and such field values are not interoperable. This is a consequence
+   of the field being specified as a choice between a single and multiple values.
 </t>
 </section>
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5786,6 +5786,12 @@ Expect: 100-continue
    The If-Match header field can be ignored by caches and intermediaries
    because it is not applicable to a stored response.
 </t>
+<t>
+   Note that an If-Match header field with a list value containing "*" and
+   other values (including other instances of "*") does not have defined
+   semantics, and such field values are not interoperable. This is a consequence
+   of the field being specified as a choice between a single and multiple values.
```
royfielding, line 5793:
```suggestion
   other values (including other instances of "*") is unlikely to be interoperable.
```

royfielding: Let's not be pedantic. The spec does not prevent senders from sending nonsense.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5868,6 +5874,12 @@ Expect: 100-continue
    Requirements on cache handling of a received If-None-Match header field
    are defined in <xref target="validation.received"/>.
 </t>
+<t>
+   Note that an If-None-Match header field with a list value containing "*" and
+   other values (including other instances of "*") does not have defined
+   semantics, and such field values are not interoperable. This is a consequence
+   of the field being specified as a choice between a single and multiple values.
```
royfielding, line 5881:
```suggestion
   other values (including other instances of "*") is unlikely to be interoperable.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -12602,6 +12614,7 @@ Content-Encoding: gzip
   <li>In <xref target="field.accept"/> and <xref target="status.415"/>, allow "Accept" as response field (<eref target="https://github.com/httpwg/http-core/issues/48"/>)</li>
   <li><xref target="collected.abnf"/> now uses the sender variant of the "#" list expansion (<eref target="https://github.com/httpwg/http-core/issues/192"/>)</li>
   <li>In <xref target="field.name.registry"/>, add optional "Comments" entry (<eref target="https://github.com/httpwg/http-core/issues/273"/>)</li>
+  <li>In <xref target="field.if-match"/> and <xref target="field.if-none-match"/>, state that multiple "*" has no defined semantics (<eref target="https://github.com/httpwg/http-core/issues/305"/>)</li>
```
royfielding, line 12617:
```suggestion
  <li>In <xref target="field.if-match"/> and <xref target="field.if-none-match"/>, state that multiple "*" is unlikely to be interoperable (<eref target="https://github.com/httpwg/http-core/issues/305"/>)</li>
```

PR #406, opened 2020-07-13
Mnot 335b

mnot:
For #335


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -10607,8 +10605,7 @@ Content-Encoding: gzip
 </t>
 <t>
    To promote interoperability, specific recommendations are made for minimum
-   size limits on request-line (<xref target="request.line"/>)
-   and fields (<xref target="header.and.trailer.fields"/>). These are
+   size limits on fields (<xref target="header.and.trailer.fields"/>). These are
```
reschke, line 10608: shouldn't this reference "field.limits"?

mnot, line 10608:
```suggestion
   size limits on fields (<xref target="field.limits"/>). These are
```

PR #412, opened 2020-07-20
Host is 1.1-specific

### LINKED ISSUE #182: Section 5.4 (Host) in Semantics has normative 1.1 requirements
In section 5.4, there are a couple of normative requirements for HTTP/1.1 ("A client MUST send a Host header field in all HTTP/1.1 request messages."; "A client MUST send a Host header field in an HTTP/1.1 request"; "A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request"), and it makes two references to the `absolute-form` production, which is defined in the Messaging draft (and hence is a dangling reference in Semantics). Those should probably live in Messaging, and the Semantics document should be more general.

mnot:
Fixes #182


reschke:
FWIW, does it really help for "Host" to come first if the recipient still has to scan for more occurences to satisfy the subsequent requirement? ("A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request message that lacks a Host header field and to any request message that contains more than one Host header field or a Host header field with an invalid field value.")

royfielding:
> does it really help for "Host" to come first if the recipient still has to scan for more occurences to satisfy the subsequent requirement

Yes, because request information is typically handled in stages with the first stage determining which server/config applies to this request. If none of its hosts match, there's no need to continue. The other requirements can be satisfied later when the entire header section is available.

royfielding:
To put it another way, some servers might choose to terminate a request if the first full block of read memory does not contain Host, since (in the real world practice they care about) that only occurs for denial of service attacks.

Issue #368, opened 2020-05-20
Semantics organisation


mnot:
I find that I struggle with the current organisation of Semantics every time I open it. When you're looking for how a given mechanism (e.g., conneg, partial content) works, it's very confusing to have to bounce between parts of s 6, s 8, s 10 and sometimes s 9.

I think the document would read far more easily and logically (especially in the text-only rendering) if we added four new top-level sections, with the content from the listed sections inserted (likely with some rearrangement):

* Partial Content
  * 8.3 Range
  * 6.3.4 Content-Range
  * 6.3.5 Media Type multipart/bytesranges
  * 10.4.1 Accept-Ranges
* Content Negotiation
  * 6.4 Content Negotation
  * 8.4 Negotiation
  * 6.2.1 Content-Type
  * 6.2.2 Content-Encoding
  * 6.2.3 Content-Language
* Conditional Requests
  * 8.2 Preconditions
  * 10.2 Validators
* Authentication
  * 10.3 Authentication Challenges
  * 8.5 Authentication Credentials

I haven't suggested moving the status codes, as that would be more disruptive, and I think there isn't much value in doing so.

reschke:
I think I agree.

It also helps with fields thare both request and response fields. Right now, Accept-Encoding is in Secion 8, although it also defines the response variant.

royfielding:
The current organization of Semantics is split by request/response, whereas Conditionals, Range, Auth, Conneg, and Caching are better explained in proximity. That is, of course, why we split them for 7231.

I have been thinking of going back to the reference style of 2616, where the major text sections describe what can be done by reference to the detailed sections per element sorted alphabetically.

I can try that after the Apache board meeting today. Did I mention that I am now chairman of the ASF, again?

reschke:
FWIW, if we do a major re-org, we should avoid to have that mixed with other changes (in the same Internet Draft).

royfielding:
Yes, we are long overdue for a new draft set. I think we need to stop waiting for meetings to generate them.

reschke:
Will submit one today.

royfielding:
Unfortunately, any depth-first approach to organizing the document doesn't work because each protocol element is important for two to four different actions one might want to take in HTTP. We keep getting lost in the details.

[...]

PR #414, opened 2020-07-23
Semantics re-org (#368)

royfielding:
I am trying to find a sensible arrangement of sections for Semantics now that it contains so many parts.

My first attempt is to move all detailed field definitions into a major section (between methods and status codes), somewhat like the reference format of RFC2616, and then rearrange the prose descriptions of architecture, messaging, routing, representation, validators, preconditions, range, and auth, so that the whole protocol is described first, using overviews, before the details of each protocol element are defined.

This PR is a draft that will change substantially as things move. I will try to separate section moves from content changes.


mnot:
For #368

royfielding:
git is so horrifically bad at merges that it is easier and cleaner to recreate the org changes manually than keep merging master into here.

Issue #64, opened 2018-05-01
introduce both 1.1 and 2.0


reschke:
The introduction should reference the currently defined wire formats (1.1 and 2, also 1.0 maybe).

We also need to figure out whether and where to mention differences:

- expect/continue
- transfer codings
- status phrase

mnot:
Also, the current introduction to all three documents says:

> HTTP is defined by a series of documents that collectively form the HTTP/1.1 specification

That needs to be adjusted to something like:

> HTTP is defined by [semantics] and [caching], and is mapped to specific wire protocols as versions; currently, [http/1.2] and [http/2] are defined.

MikeBishop:
> > HTTP is defined by [semantics] and [caching], and is mapped to specific wire protocols as versions; currently, **[http/1.2]** and [http/2] are defined.

HTTP/1.2?  Is there something you need to share with the working group, @mnot?  😉

royfielding:
I think (but am not sure) that this has been covered by the changes made in #416 

Maybe @MikeBishop and @martinthomson could have a look and see if the short HTTP/2 and HTTP/3 descriptions need more flowers.

martinthomson:
I think that the number of flowers (->0) is entirely appropriate and I would be happy with those descriptions, with maybe a small tweak to the HTTP/3 text.

mnot:
I left a review; once we resolve comments on the PR, I think this can be closed.

PR #416, opened 2020-07-26
revise introduction for HTTP family of protocols; addresses #64

royfielding:
I got distracted by old text and revised the introduction of semantics.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -12292,14 +12292,13 @@ Content-Encoding: gzip
   <seriesInfo name="RFC" value="8336"/>
 </reference>
 
-<reference anchor="HTTP3"
-           target="https://datatracker.ietf.org/doc/html/draft-ietf-quic-http">
+<reference anchor="HTTP3">
   <front>
     <title>Hypertext Transfer Protocol Version 3 (HTTP/3)</title>
     <author initials="M." surname="Bishop" fullname="Mike Bishop"/>
     <date year="2020" month="June" day="9"/>
   </front>
-  <seriesInfo name="Internet Draft" value="(Work in Progress)"/>
+  <seriesInfo name="Internet-Draft" value="draft-ietf-quic-http-29"/>
 </reference>
 
 <reference anchor="BCP13">
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -131,36 +131,27 @@
 
 <middle>
 <section title="Introduction" anchor="introduction">
+<section title="Purpose" anchor="intro.purpose">
 <t>
-   The Hypertext Transfer Protocol (HTTP) is a stateless application-level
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -174,67 +165,157 @@
...
+   the standards track in 1997 <xref target="RFC2068"/>,
+   1999 <xref target="RFC2616"/>, and 2014
+   (<xref target="RFC7230"/> – <xref target="RFC7235"/>).
+</t>
+<t>
+   HTTP/2 (<xref target="RFC7540"/>) introduced a multiplexed session layer
+   on top of the existing TLS and TCP protocols for exchanging concurrent
+   HTTP messages with efficient header field compression and server push.
+   HTTP/3 (<xref target="HTTP3"/>) has gone even further, guiding the
+   development of QUIC as a secure multiplexing transport over UDP.
```
martinthomson, line 199: I wouldn't have said "guiding" here.  Though it isn't *wrong*, it is probably better to talk about the outcomes of the process, which is that HTTP/3 provides greater independence for concurrent messages by using a QUIC, which is a secure multiplexed transport that uses UDP rather than TCP.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -131,36 +131,27 @@
...
 <section title="Introduction" anchor="introduction">
+<section title="Purpose" anchor="intro.purpose">
 <t>
-   The Hypertext Transfer Protocol (HTTP) is a stateless application-level
-   request/response protocol that uses extensible semantics and
-   self-descriptive messages for flexible interaction with network-based
-   hypertext information systems. HTTP is defined by a series of documents
-   that collectively form the HTTP/1.1 specification:
+   The Hypertext Transfer Protocol (HTTP) is a family of stateless,
+   application-level, request/response protocols that use a generic interface,
```
mnot, line 137: I think "share" would be better than "use" here; that better reflects the relationship between the various "family members."

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -174,67 +165,157 @@
...
+   HTTP/1.1 was designed to refine the protocol's features while retaining
+   compatibility with the existing text-based messaging syntax, improving
+   its interoperability, scalability, and robustness across the Internet.
+   This included length-based payload delimiters for both fixed and dynamic
+   (chunked) content, a consistent framework for content negotiation,
+   opaque validators for conditional requests, cache controls for better
+   cache consistency, range requests for partial updates, and default
+   persistent connections. HTTP/1.1 was introduced in 1995 and published on
+   the standards track in 1997 <xref target="RFC2068"/>,
+   1999 <xref target="RFC2616"/>, and 2014
```
mnot, line 191: "with substantial revisions in..."

royfielding: I tried several variations but this doesn't work with the sentence (and bloats it in a way that isn't helpful).

PR #417, opened 2020-07-27
make section xrefs in tables more concise and consistent (editorial)

royfielding:
did I mention that I hate line breaks in text table crossrefs?

This makes the text format much better, but the html output is worse because the stylesheet isn't padding cells. I am hoping that would be a simple fix.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -256,55 +256,55 @@
     <tr>
       <th>Title</th>
       <th>Reference</th>
-      <th>Changes</th>
+      <th>See</th>
     </tr>
   </thead>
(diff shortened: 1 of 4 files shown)


reschke:
Hm.

I'm not convinced that we should optimize for the text output.

I can revise the table padding, but need to be careful with that.

reschke:
Yes, it needs to come out right with xml2rfc's HTML output.

reschke:
@royfielding - https://github.com/reschke/xml2rfc/commit/6e0313585f451aa21615b90e3cd203cb82ae0a97

royfielding:
I was thinking about this last night and realized that "See" as a column heading might offend someone. I will try changing that to "Ref."

Issue #203, opened 2019-03-25
Expect should be a list header


mnot:
In bis, we locked down `Expect` to a single value, `100-continue`. That's probably a bug; it needs to act like a list header still, even though it only has one allowed value. We might have to define error handling here too.

mcmanus:
ietf104: no conerns raised about adopting a list format

reschke:
I guess we then need to pull in the full old ABNF:

~~~
   Expect       =  "Expect" ":" 1#expectation

   expectation  =  "100-continue" | expectation-extension
   expectation-extension =  token [ "=" ( token | quoted-string )
                            *expect-params ]
   expect-params =  ";" token [ "=" ( token | quoted-string ) ]
~~~

and the simplify it to:

~~~
   Expect       =  "Expect" ":" 1#expectation

   expectation =  token [ "=" ( token | quoted-string ) *expect-params ]
   expect-params =  ";" token [ "=" ( token | quoted-string ) ]
~~~

...and then define dthat the only expection that exists is "100-continue" with no parameters...

reschke:
I wonder whether we can use `parameter` here - node that the old ABNF comes from 2616 which had implied LWS.

annevk:
From other header parsing tests I've done I'd expect the processing model to be "split on comma and strip whitespace" and then see if any resulting token is an ASCII-case-insensitive match for `100-continue`.

reschke:
This introduced a bug - "*parameter" does not work, as "parameter" does not include the ";" delimiter.

mnot:
This can now be closed, correct?

reschke:
Yes.

Issue #33, opened 2017-07-10
MIME type parsing


annevk:
See http://www.rfc-editor.org/errata_search.php?rfc=7231&eid=4031 (rejected due to scope). Clients typically (or all?) allow `text/html;` despite it being invalid.

annevk:
One of the things that I couldn't find addressed anywhere is something like `text/html;charset=gbk;charset=utf-8`. The introduction of https://tools.ietf.org/html/rfc2046 mentions parameters as a set, thereby potentially outlawing duplicate names, but there's nothing normative to that effect anywhere. (Browsers are evenly split on how they handle this case by the way.)

mnot:
See also https://github.com/whatwg/mimesniff/issues/30 e.g., discussion of multiple content-types. This has potential security impact.

annevk:
For MIME types I ended up (re)writing https://mimesniff.spec.whatwg.org/#understanding-mime-types btw and also have a bunch of tests at https://github.com/w3c/web-platform-tests/tree/master/mimesniff/mime-types.

mnot:
Discussed in Basel; inclination is to change the ABNF, and survey similar places to align if necessary.

snuggs:
@mnot pardon my ignorance but what is ABNF? Am I wrong to assume [it is this](https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_form)

Thanks in advance!

mnot:
It is.

mnot:
Note that this is about the trailing semicolon, at least initially.

mnot:
Discussed on editors' call; we should create a new ABNF rure `parameters` in 5.4.1.4. Parameters that allows semicolons without following `parameter`, and use that where appropriate.

reschke:
Just to check (grep in httpbis.abnf):
```
expectation = token [ "=" ( token / quoted-string ) *parameter ]

media-range = ( "*/*" / ( type "/*" ) / ( type "/" subtype ) ) *( OWS ";" OWS parameter )
media-type = type "/" subtype *( OWS ";" OWS parameter )

parameter = parameter-name "=" parameter-value
parameter-name = token
parameter-value = ( token / quoted-string )

transfer-coding = token *( OWS ";" OWS transfer-parameter )
transfer-parameter = token BWS "=" BWS ( token / quoted-string )
```

[...]

PR #420, opened 2020-07-31
fix parameter ABNF for "expect" - broken by previous change (#203)

reschke:
(this is a regression from #408 )


mnot:
LGTM. Do we need a separate issue re: whether to include `;` in parameter (or more likely, a new construct)?

reschke:
> LGTM. Do we need a separate issue re: whether to include `;` in parameter (or more likely, a new construct)?

Just change the title of https://github.com/httpwg/http-core/issues/33 maybe?

PR #421, opened 2020-07-31
Remove 1#

### LINKED ISSUE #210: unnecessary use of 1#element for header field definitions

mnot:
Fixes #210.


mnot:
I don't think any additional prose is necessary; the semantics follow pretty cleanly in each case. E.g., for `Vary`, which "describes what parts of a request message, aside from the method Host header field, and target URI, might influence the origin server's process for selecting...", and an empty list means that there aren't any.

WRT changes - I don't think this is something that requires a specific note.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5885,8 +5885,9 @@ Proxy-Authorization: basic aGVsbG86d29ybGQ=
    order to properly handle this request.
 </t>
 <sourcecode type="abnf7230"><iref primary="true" item="Grammar" subitem="Expect"/>
-  <x:ref>Expect</x:ref> =      #<x:ref>expectation</x:ref>
-  <x:ref>expectation</x:ref> = <x:ref>token</x:ref> [ "=" ( <x:ref>token</x:ref> / <x:ref>quoted-string</x:ref> ) *<x:ref>parameter</x:ref> ]
+  <x:ref>Expect</x:ref> =      1#<x:ref>expectation</x:ref>
```
mnot:
```suggestion
  <x:ref>Expect</x:ref> =      #<x:ref>expectation</x:ref>
```

PR #424, opened 2020-08-03
Specify how to create a redirection request

### LINKED ISSUE #38: 3xx redirects - request formation
see https://github.com/whatwg/fetch/issues/609

mnot:
Fixes #38.


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8611,7 +8611,34 @@ Content-Range: bytes 7000-7999/8000
 <t>
    The <x:dfn>3xx (Redirection)</x:dfn> class of status code indicates that
    further action needs to be taken by the user agent in order to fulfill the
-   request. If a <x:ref>Location</x:ref> header field
+   request. There are several types of redirects:
+</t>
+<ol>
+  <li>
+    Redirects that indicate the resource might be available at a
```
royfielding, line 8618:
```suggestion
    Redirects that indicate this resource might be available at a
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8611,7 +8611,34 @@ Content-Range: bytes 7000-7999/8000
...
+    different URI, as provided by the <x:ref>Location</x:ref> field,
+    as in the status codes <x:ref>301 (Moved Permanently)</x:ref>,
+    <x:ref>302 (Found)</x:ref>, <x:ref>307 (Temporary Redirect)</x:ref>, and
+    <x:ref>308 (Permanent Redirect)</x:ref>.
+  </li>
+  <li>
+    Redirection that offers a choice of matching resources, each capable
+    of representing the original target resource, as in the
```
royfielding, line 8626:
```suggestion
    Redirection that offers a choice among matching resources capable
    of representing this resource, as in the
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8611,7 +8611,34 @@ Content-Range: bytes 7000-7999/8000
...
+    <x:ref>300 (Multiple Choices)</x:ref> status code.
+  </li>
+  <li>
+    Redirection to a different resource, identified by the
+    <x:ref>Location</x:ref> field, that can represent an indirect
+    response to the request, as in the <x:ref>303 (See Other)</x:ref>
+    status code.
+  </li>
+  <li>
+    Redirection to a previously cached result, as in the
```
royfielding, line 8636:
```suggestion
    Redirection to a previously stored result, as in the
```

royfielding: because it is also used by non-cache tools like link checkers.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8620,37 +8647,59 @@ Content-Range: bytes 7000-7999/8000
...
   <li>
     <t>
-      Redirects that indicate the resource might be available at a
-      different URI, as provided by the <x:ref>Location</x:ref> field,
-      as in the status codes <x:ref>301 (Moved Permanently)</x:ref>,
-      <x:ref>302 (Found)</x:ref>, <x:ref>307 (Temporary Redirect)</x:ref>, and
-      <x:ref>308 (Permanent Redirect)</x:ref>.
+      Replace the target URI with the URI in the redirection response's
+      <x:ref>Location</x:ref> header field value, resolving it against the original
+      request's target URI if it is not absolute.
```
royfielding, line 8658:
```suggestion
      Replace the target URI with the URI referenced by the redirection response's
      <x:ref>Location</x:ref> header field value after resolving it relative to the original
      request's target URI.
```

royfielding:
er, the above comment refers to julian's removal of the html generated file, which needs to be removed from the change set (not from master when this PR is applied).

royfielding:
fixed it

mnot:
Change notes added.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -13001,6 +13001,11 @@ transfer-coding = &lt;transfer-coding, see <xref target="Messaging" x:fmt="," x:
    values are present was clarified.
    (<xref target="field.vary"/>)
 </t>
+<t>
+   Parameters in media type, media range, and expectation can be empty via
+   one or more trailing semicolons.
+   (<xref target="parameter"/>)
+</t>
```
reschke, line 13008: Hm.

If I have 

    text/plain ; foo=bar ; ; charset=qux ;

are all of these that we now allow "trailing"?

reschke: ...also this change seems to apply to a different issue...

PR #425, opened 2020-08-03
Define error handling for multiple Content-Type

### LINKED ISSUE #39: Content-Type header parsing
This is related to #33, but subtly different. As explored in https://github.com/whatwg/mimesniff/issues/30 browsers have different code paths for request and response `Content-Type` header parsing. Values such as `*/*, text/html` in a response `Content-Type` header end up being interpreted as `text/html`, presumably for compatibility with deployed content.

This seems like another fallout of intermediaries adding potentially duplicate headers and (early) implementations being poorly tested for erroneous input.

mnot:
Fixes #39.


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4030,6 +4030,13 @@ bytes=500-700,601-999
    match multiple media types that differ only in processing semantics.
    Implementers are encouraged to provide a means to disable such sniffing.
 </t>
+<t>
+   Furthermore, although Content-Type is defined as a singleton header, it is
+   sometimes sent as a list (i.e., with multiple, comma-separated values).
+   Recipients &MAY; attempt to handle this error by using the last syntactically
+   valid member of the list. Note that some implementations might have additional
+   error handling behaviors; e.g., see <xref target="FETCH"/>.
+</t>
```
reschke, line 4039: Does this imply that they MUST NOT handle the error differently? In which case I would object to this as it is a normative change. If it's not, it shouldn't use "MAY".

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4030,6 +4030,13 @@ bytes=500-700,601-999
    match multiple media types that differ only in processing semantics.
    Implementers are encouraged to provide a means to disable such sniffing.
 </t>
+<t>
+   Furthermore, although Content-Type is defined as a singleton header, it is
+   sometimes sent as a list (i.e., with multiple, comma-separated values).
+   Recipients &MAY; attempt to handle this error by using the last syntactically
+   valid member of the list. Note that some implementations might have additional
+   error handling behaviors; e.g., see <xref target="FETCH"/>.
```
royfielding, line 4038:
```suggestion
   Furthermore, although Content-Type is defined as a singleton field, some
   incorrect configurations might cause it to be generated multiple times for
   a response, resulting in a combined field value that appears to be a list.
   Recipients often attempt to handle this error by using the last syntactically
   valid member of the list, but that might bypass upstream security filters if
   they do not share the same error recovery algorithm.
```

I don't see a reason to reference FETCH here.

draft-ietf-httpbis-semantics-latest.xml
royfielding: I also don't think it is good advice. I've never seen an application choose the last value on error.

mnot: MAY changed to can.

@royfielding see issue discussion; this is how all browsers handle it (with [tests](https://wpt.fyi/results/fetch/content-type/response.window.html?label=experimental&label=master&aligned)), so it seems worth mentioning as an option (but not mandating).

mnot:
PTAL at revision; I mentioned Fetch before because of the on-issue discussion, but looking at where they're at, it looks premature to mention it yet.

reschke:
So I updated http://test.greenbytes.de/tech/tc/httpcontenttype/ and included tests for multiple field values. Surprisingly, Safari handles <http://test.greenbytes.de/tech/tc/httpcontenttype/#multitextxmloneline> differently than Chrome and Firefox (so it treats comma-separated different than multiple field instances). @annevk - is this known; does it have a webkit bug #?

PR #426, opened 2020-08-03
Move Connection to Semantics

### LINKED ISSUE #407: Connection
Connection currently is in Messaging, with a few references to it from Semantics.

As per #335, we want to make Semantics as independent from Messaging as possible. 

We've talked about two approaches:

1. I've advocated keeping Connection limited to Messaging, with appropriate (but non-normative) references in Semantics. See 32d7e519ec5 for what this would look like.

2. AIUI @royfielding wants to make a normative requirement on _all_ versions of HTTP to process Connection. This could mean just leaving the text as it is, or moving more of Connection into Semantics (although I imagine that the bits about connection options would probably stay).

Whatever we do will need to be reconciled with [this HTTP/2 text](https://httpwg.org/specs/rfc7540.html#n-connection-specific-header-fields).

mnot:
Fixes #407.


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1781,6 +1674,15 @@ https://www.example.org
        the current response; otherwise,</li>
    <li>The connection will close after the current response.</li>
 </ul>
+<t>
```
reschke, line 1677: need to link to semantics#field.connection here

PR #427, opened 2020-08-03
Add parameters rule

### LINKED ISSUE #33: MIME type parsing

mnot:
Fixes #33.


annevk:
You mean something like `text/html;something=else;;charset=utf-8`? I'm pretty positive the encoding would be picked up.

`text/html;;;;charset=gbk` in https://github.com/web-platform-tests/wpt/blob/master/mimesniff/mime-types/resources/mime-types.json kinda tests for that, but there would be more tests.

reschke:
thanks @annevk

reschke:
I believe this needs to be mentioned as normative change.

Furthermore, the change note only mentions trailing semicolons, but we now allow bare semicolons in between as well.

PR #428, opened 2020-08-04
English improvements

mnot:


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1499,7 +1497,7 @@
    <iref item="no-transform (cache directive)" primary="true" />
 <t>
    The "no-transform" request directive indicates that the client is asking
-   for intermediares (whether or not they implement a cache) to avoid
+   for intermediaries (whether they implement a cache) to avoid
```
reschke, line 1500: this might be incorrect

mnot: https://english.stackexchange.com/questions/3382/whether-or-not-vs-whether

reschke: ok

royfielding: don't believe everything you read on the Internet

OTOH, we can delete the parenthetical.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -778,14 +776,14 @@
 <t>
    The <x:ref>Age</x:ref> header field is used to convey an estimated
    age of the response message when obtained from a cache. The Age field value
-   is the cache's estimate of the number of seconds since the response was
-   generated or validated by the origin server. In essence, the Age value is
+   is the cache's estimate of the number of seconds since origin server generated
```
royfielding, line 779:
```suggestion
   is the cache's estimate of the number of seconds since the origin server generated
```

draft-ietf-httpbis-cache-latest.xml
royfielding:
```suggestion
   for intermediaries to avoid
```

PR #431, opened 2020-08-06
Talk about client half-close

### LINKED ISSUE #22: Clarify rules around half-closed TCP connections
The HTTP RFCs say nothing about the expectations around half-closed TCP connections.

In practice, I haven't seen any HTTP client in the wild send a request, and then send a FIN (`shutdown`) while still waiting for the server's response.

Because we haven't see any clients do that, as of Go 1.8, Go's HTTP server is starting to make assumptions that reading an EOF from the client means that the client is no longer interested in the response. (reading EOF being the closest portable approximation to "the client has gone away").

But in https://github.com/golang/go/issues/18527, a user reports that they have an internal HTTP client which does indeed make a half-closed TCP request.

It would be nice if the HTTP RFCs provided guidance as to whether this is allowed or frowned upon.

I would recommend that the RFC suggest that clients SHOULD NOT half-close their TCP connections while awaiting responses. Because nobody else does, empirically, and relying on reading EOF is a useful signal for servers.

/cc @mnot @benburkert

mnot:
Fixes #22.

### DIFF
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -1199,11 +1199,13 @@ https://www.example.org
    length-delimited messages whenever possible.  The close-delimiting
    feature exists primarily for backwards compatibility with HTTP/1.0.
 </t>
-<t>
-   Note: Request messages are never close-delimited because they are always
-   explicitly framed by length or transfer coding, with the absence of both implying
-   the request ends immediately after the header section.
-</t>
+<aside>
+  <t>
+     &Note; Request messages are never close-delimited because they are always
+     explicitly framed by length or transfer coding, with the absence of both implying
+     the request ends immediately after the header section.
+  </t>
+</aside>
 <t>
    A server &MAY; reject a request that contains a message body but
    not a <x:ref>Content-Length</x:ref> by responding with
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -1904,6 +1904,12 @@ https://www.example.org
    It is unknown whether the reset problem is exclusive to TCP or might also
    be found in other transport connection protocols.
 </t>
+<t>
+   Note that some servers will interpret the client closing its side of a
+   connection as a loss of interest in the response, and therefore close
+   their side without sending any outstanding portions of the response,
+   in order to save resources.
+</t>
 </section>
 
 <section title="TLS Connection Closure" anchor="tls.connection.closure">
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1198,6 +1198,7 @@ https://www.example.org
    by network failure, a server &SHOULD; generate encoding or
    length-delimited messages whenever possible.  The close-delimiting
    feature exists primarily for backwards compatibility with HTTP/1.0.
+   Note that a half-closed TCP connection does not delimit a request.
```
royfielding, line 1201: This note appears out of nowhere.

I think it would be better to change the beginning sentence to add "response" above, as in

```
   Since there is no way to distinguish a complete close-delimited response
   message from a partially received response interrupted
   by network failure, a server &SHOULD; generate encoding or
   length-delimited response messages whenever possible.
   This close-delimiting
   feature exists primarily for backwards compatibility with HTTP/1.0.

```
and then add the note as a paragraph:
```suggestion
   Note: Request messages are never close-delimited because they are always
   explicitly framed by length or transfer coding, with the absence of both implying
   the request ends immediately after the header section. A half-closed TCP
   connection does not delimit a request message, nor does it imply that the client
   is no longer interested in a response. In general, do not rely on transport signals
   to signal edge cases, since HTTP/1.1 is independent of transport.
```

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1903,6 +1904,12 @@ https://www.example.org
    It is unknown whether the reset problem is exclusive to TCP or might also
    be found in other transport connection protocols.
 </t>
+<t>
+   Note that some servers will interpret the client closing its side of a
+   connection as a loss of interest in the response, and therefore close
+   their side without sending any outstanding portions of the response,
+   in order to save resources.
+</t>
```
royfielding, line 1912: I don't think we should add this just because one implementation fails to correctly interpret the protocol. I don't care if it is in Go, they can and should fix the bug.

royfielding (changes requested):

draft-ietf-httpbis-messaging-latest.xml
royfielding: Wait, that's a bit confusing anyway. 

Servers do interpret a full close (client closing read side) as a connection termination, which is fine. What they don't do, normally, is mark the whole connection closed if the client sends FIN (half-close) because that doesn't mean anything other than no more requests to read.

mnot: Reading the issue again I think this is probably the best we can do. I rearranged some; PTAL.

MikeBishop (approved):

royfielding (approved):

PR #432, opened 2020-08-07
remove edge case requirement for If-Match and If-Unmodified-Since

### LINKED ISSUE #166: No validator requirement in 2xx response for conditional update
RFC7232 in https://tools.ietf.org/html/rfc7232#section-3.1 has the following clause:
```
In the latter case, the origin server MUST NOT send a validator header field in the response unless it can verify that the request is a duplicate of an immediately prior change made by the same user agent
```
It doesn't really explain what security or performance considerations are leading to such a requirement, but it calls for a lot of infrastructure and design work to implement it because that logic can be implemented only if the server is stateful. The stateless server cannot comply with such a requirement.

See related mail list thread: https://lists.w3.org/Archives/Public/ietf-http-wg/2018OctDec/0051.html

Either RFC should explain why such logic MUST be implemented or relax "MUST" into "SHOULD" in that clause or remove that clause completely.

royfielding:
evaluating false on a duplicate change request sending 2xx without validator for reasons unknown.

Fixes #166

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -13013,9 +13013,16 @@ transfer-coding = &lt;transfer-coding, see <xref target="Messaging" x:fmt="," x:
 
 <section title="Changes from RFC 7232" anchor="changes.from.rfc.7232">
 <t>
-   Clarify that If-Unmodified-Since doesn't apply to a resource without a
+   Clarified that If-Unmodified-Since doesn't apply to a resource without a
    concept of modification time.
    (<xref target="field.if-unmodified-since"/>)
+</t>
+<t>
+   Removed edge case requirement on If-Match and If-Unmodified-Since that a
+   validator not be sent in a 2xx response when validation fails and the server
+   decides that the same change request has already been applied.
+   (<xref target="field.if-match"/> and
+   <xref target="field.if-unmodified-since"/>)
 </t>
 </section>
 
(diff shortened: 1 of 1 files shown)


royfielding:
Hmm, of course, now I found another related message in my archives that prompted me to remember why the requirement was added. I should probably rewrite the paragraph (without that requirement) and describe the edge case as a note.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6407,18 +6407,29 @@ Expect: 100-continue
...
-   a) the <x:ref>412 (Precondition Failed)</x:ref> status code or
-   b) one of the <x:ref>2xx (Successful)</x:ref> status codes if the origin
-   server has verified that a state change is being requested and the final
-   state is already reflected in the current state of the target resource
+   If-Match condition evaluates to false. Instead, the origin server &MAY;
+   indicate that the conditional request failed by responding with a
+   <x:ref>412 (Precondition Failed)</x:ref> status code. Alternatively,
+   if the request is a state-changing operation that appears to have already
+   been applied to the selected representation, the origin server &MAY; respond
+   with a <x:ref>2xx (Successful)</x:ref> status code
```
reschke, line 6415: Here, we are replacing a "MUST reply with 412 or 2xx" with "MAY reply with 412 or MAY reply with 2xx". Not sure whether this change of wording is good.

royfielding: The original wording is wrong because a server can send any code to the client, for any reason, and it only effects interoperability if the code doesn't reflect the server's action. For example, timeouts, invalid credentials, server error, etc. What an origin server must do is a lot more complicated that an either/or phrase.

royfielding: An alternative would be to write these as "if the server responds with 2xx, it means ..." kind of statements.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6407,18 +6407,29 @@ Expect: 100-continue
...
    (i.e., the change requested by the user agent has already succeeded, but
    the user agent might not be aware of it, perhaps because the prior response
-   was lost or a compatible change was made by some other user agent).
-   In the latter case, the origin server &MUST-NOT; send a validator header
-   field in the response unless it can verify that the request is a duplicate
-   of an immediately prior change made by the same user agent.
+   was lost or an equivalent change was made by some other user agent).
+</t>
+<t>
+   Allowing an origin server to send a success response when a change request
```
mnot, line 6421: successful

PR #439, opened 2020-08-21
Allow for multiple trailer sections and define iterative processing

### LINKED ISSUE #313: Mid-stream trailers
As discussed in #220 and https://github.com/quicwg/base-drafts/issues/3265, mid-stream trailers (i.e., those that arrive before the end of the payload body) are interesting.

A few things to discuss:

- [ ] What are some potential use cases?

- [ ] We recently stopped automatically combining trailers into headers, as that created a lot of issues. What about these -- are they yet again distinct from trailers as well as headers? Or can they be "downgraded" into trailers automatically? In other words, when transitioning from a hop that supports these from one that doesn't, are they just dropped onto the floor?

- [ ] Related - is the expectation that implementations will create separate APIs for these, or would it be that trailers are just available / writable early?

- [ ] What relationship does one of these have to the payload body? E.g., could I write one that "knows" exactly where in the body it occurs, so that it could sign a chunk of the body relative to it? Or could a handling implementation change its ordering in relation to body bytes?

royfielding:
Fixes #313

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -513,14 +513,15 @@
 <iref primary="true" item="complete"/>
 <iref primary="true" item="incomplete"/>
 <t>
-   Each major version of HTTP defines its own syntax for the inclusion of
-   information in messages. Nevertheless, a common abstraction is that a
-   message includes some form of envelope/framing, a potential set of named
-   fields up front (a header section), a potential body, and a
-   potential following set of named fields (a trailer section).
+   Each major version of HTTP defines its own syntax for the communication of
+   messages. Nevertheless, a common abstraction is that each message contains
+   some form of envelope/framing with self-descriptive control data that
+   indicates its semantics and routing, a potential set of named fields
+   up front (a header section), a potential body, and potential fields
+   sent after the body begins (trailer sections).
 </t>
 <t>
-   A client sends an HTTP request to a server in the form of a
+   A client sends requests to a server in the form of a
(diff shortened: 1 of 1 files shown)


mnot (approved):

ioquatix:
Does this mean it's now valid to interleave HTTP/2 DATA frames and HEADER frames?

ioquatix:
```
An HTTP/2 message (request or response) consists of:

   1.  for a response only, zero or more HEADERS frames (each followed
       by zero or more CONTINUATION frames) containing the message
       headers of informational (1xx) HTTP responses (see [RFC7230],
       Section 3.2 and [RFC7231], Section 6.2),

   2.  one HEADERS frame (followed by zero or more CONTINUATION frames)
       containing the message headers (see [RFC7230], Section 3.2),

   3.  zero or more DATA frames containing the payload body (see
       [RFC7230], Section 3.3), and

   4.  optionally, one HEADERS frame, followed by zero or more
       CONTINUATION frames containing the trailer-part, if present (see
       [RFC7230], Section 4.1.2).
```

Does this mean we cannot implement this on HTTP/2?

royfielding:
> Does this mean it's now valid to interleave HTTP/2 DATA frames and HEADER frames?

The feature was removed during WGLC, but no -- it would have required a new frame type.

ioquatix:
Okay, so chunk extensions remains a unique feature of HTTP/1 chunked encoding and there is no plan to expand that further?

reschke:
Yes.

PR #441, opened 2020-08-25
Explain why GET-based queries might transmit sensitive data, suggest …

### LINKED ISSUE #277: Clarification of GET versus POST for information retrieval
Am sure this has been discussed before, and I'm opening a can of worms but here goes (if nothing else would be nice to have this as an issue to be easily referenceable)...

The traditional accepted use case for [GET](https://tools.ietf.org/html/rfc7231#section-4.3.1) is for information retreival and [POST](https://tools.ietf.org/html/rfc7231#section-4.3.3) is for uploading information to a server. This is reflected in the language of the definitions as linked in the previous sentences and does not seem to be materially changed in the [new versions currently in draft](https://tools.ietf.org/html/draft-ietf-httpbis-semantics-06#section-7.3).

However the sections on [Disclosure of Personal Information](https://tools.ietf.org/html/rfc7231#section-9.3) and [Disclosure of Sensitive Information in URIs](https://tools.ietf.org/html/rfc7231#section-9.4) recognise the privacy and security implications of putting sensitive information in the URL as Query Params and state that "Such services ought to use POST-based form submission instead." Again this language seems [unchanged in the current drafts](https://tools.ietf.org/html/draft-ietf-httpbis-semantics-06#section-13.8).

This seems out of sync with the definition of POST provided previously and leads to arguments as to whether information retrieval by sensitive params is an "acceptable" use of POST (particularly for those insisting on a rigid [definition of REST](https://en.wikipedia.org/wiki/Representational_state_transfer#Relationship_between_URI_and_HTTP_methods)). Note I'm not talking about authentication cookies here - but more other params which clarify the resouce required, rather than authenticate access to such a resource. For example "download document id 1234" is often an information retreival request, which is supplemented by cookies to confirm access to that document. In this case the document id may be considered sensitive (e.g. if it's a policy number, or a claim identifier) so should not be sent as a URL param which would rule out GET, unless you want to get into providing the params as HTTP Headers instead which just seems like a messy idea.

Of course you could use a GET with a body but the RFC [explicitly notes why this is a bad idea](https://tools.ietf.org/html/rfc7231#section-4.3.1): "sending a payload body on a GET request might cause some existing implementations to reject the request." and this has been [strengthened in the lastest draft](https://tools.ietf.org/html/draft-ietf-httpbis-semantics-06#section-7.3.1) to: "A client SHOULD NOT generate a body in a GET request" in #202 

The POST definition doesn't explicitly forbid it's use for information retreival, but all the examples suggest data is being sent to the server for storage (perhaps not the first example, but certainly the others) and there are [many](https://stackoverflow.com/questions/52050997/should-i-use-a-post-request-to-send-a-retrieval-request-to-my-server-for-a-large) examples of this being debated.

While this has always been an issue, it is getting more and more important in this more privacy-aware world and with various legislation and security requiremnts and I feel the specs could be more explicit, and reflect what I imagine is a common use case for POST.

Therefore should text and/or the examples be added to clarify the position in regards to retreival via POST requests, and to resolve the apparent discrepency in regards to the [Disclosure of Sensitive Information in URIs](https://tools.ietf.org/html/rfc7231#section-9.4) section?

royfielding:
…POST, link to security consideration, and provide some workarounds

Fixes #277

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -5380,23 +5380,30 @@ Content-Range: exampleunit 11.2-14.3/25
   </rdf:Description>
   <iref primary="true" item="GET method" x:for-anchor=""/>
 <t>
-   The GET method requests transfer of a current <x:ref>selected representation</x:ref> for
-   the <x:ref>target resource</x:ref>. GET is the primary mechanism of
-   information retrieval and the focus of almost all performance
-   optimizations. Hence, when people speak of retrieving some identifiable
-   information via HTTP, they are generally referring to making a GET request.
+   The GET method requests transfer of a current
+   <x:ref>selected representation</x:ref> for the
+   <x:ref>target resource</x:ref>.
 </t>
 <t>
-   The GET method is specifically intended to reflect the quality of
-   "sameness" identified by the request URI as if it were referenced as an
-   ordinary hypertext link.
+   GET is the primary mechanism of information retrieval and the focus of
+   almost all performance optimizations. Hence, when people speak of retrieving
+   some identifiable information via HTTP, they are generally referring to
+   making a GET request. A successful response reflects the quality of
+   "sameness" identified by the target URI.
+   In turn, constructing applications such that they produce a URI
+   for each important resource results in more resources being available for
+   other applications, producing a network effect that promotes further
+   expansion of the Web.
 </t>
 <t>
    It is tempting to think of resource identifiers as remote file system
    pathnames and of representations as being a copy of the contents of such
    files. In fact, that is how many resources are implemented (see
    <xref target="attack.pathname"/> for related security considerations).
-   However, there are no such limitations in practice. The HTTP interface for
+   However, there are no such limitations in practice.
+</t>
+<t>
+   The HTTP interface for
    a resource is just as likely to be implemented as a tree of content
    objects, a programmatic view on various database records, or a gateway to
    other information systems. Even when the URI mapping mechanism is tied to a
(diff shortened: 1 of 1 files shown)


mnot (approved):

reschke (changes requested):
What you say about POST is true not because of the definition of POST, but because of the way HTML forms work. Maybe we should say this to avoid confusion about POST - for instance, POST *can* use constructed URIs as well; it's just not common.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5427,6 +5434,18 @@ Content-Range: exampleunit 11.2-14.3/25
...
+   When information retrieval is performed with a mechanism that constructs a
+   target URI from user-provided information, such as the query fields of a
+   form using GET, potentially sensitive data might be provided that would not
+   be appropriate for disclosure within a URI
+   (see <xref target="sensitive.information.in.uris"/>). In some cases, the
+   data can be filtered or transformed such that it would not reveal such
+   information. In others, particularly when there is no benefit from caching
+   a response, using the POST method (<xref target="POST"/>) instead of GET
+   will transmit such information in the request body rather than construct a
+   new URI.
```
royfielding, line 5447:
```suggestion
   a response, using the POST method (<xref target="POST"/>) instead of GET
   will usually transmit such information in the request body rather than construct a
   new URI.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -11247,15 +11266,24 @@ Content-Encoding: gzip
...
-   sensitive data because that data will be placed in the target URI. Many
-   existing servers, proxies, and user agents log or display the target URI
-   in places where it might be visible to third parties. Such services ought
-   to use POST-based form submission instead.
+   When an application uses client-side mechanisms to construct a target URI
+   out of user-provided information, such as the query fields of a form using
+   GET, potentially sensitive data might be provided that would not be
+   appropriate for disclosure within a URI. POST is often preferred in such
+   cases because it doesn't construct a URI; instead, POST transmits the
+   potentially sensitive data in the request body. However, this hinders
```
royfielding, line 11280:
```suggestion
   cases because it usually doesn't construct a URI; instead, POST of a form transmits the
   potentially sensitive data in the request body. However, this hinders
```

reschke (approved):

PR #446, opened 2020-09-03
Reorganizing Semantics work in progress

royfielding:
Addresses #368


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -172,14 +172,15 @@
    HTTP has been the primary information transfer protocol for the World Wide
    Web since its introduction in 1990. It began as a trivial mechanism for
    low-latency requests, with a single method (GET) to request transfer of a
-   presumed hypertext document identified by a given pathname (HTTP/0.9).
+   presumed hypertext document identified by a given pathname. This original
+   protocol is now referred to as HTTP/0.9.
```
reschke, line 176: Maybe reference <https://www.w3.org/Protocols/HTTP/AsImplemented.html>?

royfielding: I am pretty sure that we tried that years ago and the RFC editor refused it.

royfielding:
So, this is pretty close. The sections are mostly organized (except the internals of Message Composition) and I haven't changed any significant content. The next step would be to consolidate redundant sections, move some of the header field descriptive content up to the associated features, and fill in remaining gaps.

Minor uglies like the IANA section additions (after generated tables) could be improved as well.

draft-ietf-httpbis-semantics-latest.xml
martinthomson: I'd try again.  They took a less stable URL to a WHATWG spec recently.

royfielding:
Done. Maybe medium-rare.

PR #453, opened 2020-09-24
is if -> as if

martinthomson:


royfielding:
Thanks, I fixed it directly in the main branch because of the reorg.

Issue #411, opened 2020-07-20
Review use of 'field'


mnot:
I want to take another look at how we use `field` vs `header field` and `trailer field`. There are a few cases where `header field` has been aggressively trimmed down to `field` which could lead readers to believe that it's valid in trailers too.

mnot:
In Messaging `field.parsing`, it says:

> Messages are parsed using a generic algorithm, independent of the
   individual field names. The contents within a given field line value are
   not parsed until a later stage of message interpretation (usually after the
   message's entire header section has been processed).

Should that be 'after the entire field section has been processed'?

mnot:
Just below:

> A server &MUST; reject any received request message that contains
   whitespace between a header field name and colon with a response status code of
   <x:ref>400 (Bad Request)</x:ref>.

Is that specific to headers?

mnot:
And then:

>    The field line value does not include any leading or trailing whitespace: OWS
   occurring before the first non-whitespace octet of the field line value or after
   the last non-whitespace octet of the field line value ought to be excluded by
   parsers when extracting the field line value from a header field line.

'header' doesn't seem appropriate there.

mnot:
In `line.folding`, 

>    Historically, HTTP header field line values could be extended over multiple

This isn't harmful (since it's just illustrative), but maybe misleading; should just be 'field line values'

mnot:
In `field.transfer-encoding`, 

> Additional information about the encoding parameters can be
   provided by other header fields not defined by this specification.

Is that an intentional restriction?

mnot:
In `message.body.length`:

>    Any response to a HEAD request and any response with a
   <x:ref>1xx (Informational)</x:ref>, <x:ref>204 (No Content)</x:ref>, or
   <x:ref>304 (Not Modified)</x:ref> status code is always
   terminated by the first empty line after the header fields, regardless of
   the header fields present in the message, and thus cannot contain a
   message body.

... or trailer sections?

reschke:
Am 08.10.2020 um 07:57 schrieb Mark Nottingham:
> Just below:
>
>     A server &MUST; reject any received request message that contains
>     whitespace between a header field name and colon with a response
>     status code of
>     <x:ref>400 (Bad Request)</x:ref>.
>
> Is that specific to headers?

For trailers, a MUST would not work, as it would require buffering the
payload before sending a response.

mnot:
Good point, I remember that now :)

mnot:
I think we can close.

PR #464, opened 2020-10-08
Go over 'field' usage

mnot:
principles I tried to adhere to:

- `field` refers to something that could be either a trailer or header; otherwise it should consistently be `header field` or `trailer field`.
- In limited contexts where it's super-clear, we can omit `header` or `trailer`; e.g., "When the Foo header field has a field value of bar..."
- It's OK to say `Foo field value` without specifying `header` or `trailer`, as long as it's clear from context what it is

In caching, I used the introducing text for each field definition to specify whether it was limited to requests or responses (as apppropriate). This is only on first use there, to keep things reasonably sized. We do that sporadically elsewhere already; unless folks push back, I'll do it for messaging and semantics in followup commits.

Change notes TBD, obviously.

For #411.


mnot:
Ping @reschke and @royfielding - don't want this to rot too much.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -318,7 +318,7 @@
    after message parsing has delineated the individual field lines.
 </t>
 <t>
-   Although the line terminator for the start-line and header
+   Although the line terminator for the start-line and
    fields is the sequence CRLF, a recipient &MAY; recognize a
```
reschke, line 322: Here, "header" is correct, no?

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -900,7 +900,7 @@ https://www.example.org
    Messages are parsed using a generic algorithm, independent of the
    individual field names. The contents within a given field line value are
    not parsed until a later stage of message interpretation (usually after the
-   message's entire header section has been processed).
+   message's entire field section has been processed).
 </t>
```
reschke, line 904: maybe "header field section"?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6075,7 +6075,7 @@ Expect: 100-continue
   <x:ref>Authentication-Info</x:ref> = #<x:ref>auth-param</x:ref>
 </sourcecode>
 <t>
-   The Authentication-Info header field can be used in any HTTP response,
+   The Authentication-Info field can be used in any HTTP response,
    independently of request method and status code. Its semantics are defined
```
reschke, line 6079: Can this be a trailer?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8403,7 +8403,7 @@ Content-Range: exampleunit 11.2-14.3/25
 </t>
 <t>
    A 204 response is terminated by the first empty line after the header
-   fields because it cannot contain a message body.
+   fields because it cannot contain a message body or trailers.
 </t>
```
reschke, line 8407: that seems to be a 11ism

reschke (changes requested):

draft-ietf-httpbis-messaging-latest.xml
mnot: I think that's correct. Could you please make suggestions using the +- icon above ^^ (cmd-g shortcut, apparently)?

draft-ietf-httpbis-semantics-latest.xml
mnot: I was hoping to avoid this; there's a much deeper question here as to whether a response without a body (1xx, 204 and 304) can have trailers, regardless of HTTP version. Should we address it?

draft-ietf-httpbis-semantics-latest.xml
mnot: > Authentication-Info can be sent as a trailer field (Section 5.6) when the authentication scheme explicitly allows this.

draft-ietf-httpbis-messaging-latest.xml
mnot: Yes. Please make a commit suggestion.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -318,7 +318,7 @@
    after message parsing has delineated the individual field lines.
 </t>
 <t>
-   Although the line terminator for the start-line and header
+   Although the line terminator for the start-line and
```
reschke, line 321:
```suggestion
   Although the line terminator for the start-line and header
```

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -900,7 +900,7 @@ https://www.example.org
    Messages are parsed using a generic algorithm, independent of the
    individual field names. The contents within a given field line value are
    not parsed until a later stage of message interpretation (usually after the
-   message's entire header section has been processed).
+   message's entire field section has been processed).
```
reschke, line 903:
```suggestion
   message's entire header field section has been processed).
```

draft-ietf-httpbis-semantics-latest.xml
reschke: I guess to. It is possible in HTTP/2, right?

draft-ietf-httpbis-semantics-latest.xml
reschke: ok

draft-ietf-httpbis-semantics-latest.xml
mnot: In the sense that a GET can have a body, possibly. Not sure what implementations do, though.

reschke: The actual point I was trying to make is that the whole statement ("...terminated by first empty line...") is specific to HTTP/1.1.

draft-ietf-httpbis-messaging-latest.xml
mnot: Ah - no, this isn't specific to header fields, it's also true for trailer fields. I.e., this text is not about the delimiter between the start-line and header fields, it's about the *line terminator* that the start-line and fields share.

draft-ietf-httpbis-messaging-latest.xml
mnot: I think this should be true for bidders and trailers as well.

draft-ietf-httpbis-semantics-latest.xml
mnot: Ah, good point, will push a correction.

draft-ietf-httpbis-messaging-latest.xml
reschke: ok.

draft-ietf-httpbis-messaging-latest.xml
reschke: Right.

draft-ietf-httpbis-semantics-latest.xml
wtarreau: For me there's a big issue there that needs to be addressed, I agree. 7540 is silent on this point. It doesn't deny it and technically it's easy to do it. But does it have any sense at all if we consider that historically the trailers were anything post-body ?  But I can understand why some implementations would like to systematically put a status in trailers, not only to ease processing, but also because it's not always easy to guess the result before sending headers. I think grpc mandates something like this for the grpc-status trailers, though I'm not sure what it suggests for bodyless responses.

wtarreau: I think the root of the question revolves around "what should an agent do when it receives a HEADERS frame indicating there's no body (204/304/response to HEAD), DATA frames carrying a body, and a HEADERS frame containing trailers. Should it just silently ignore the body but still parse the trailers ? Should it stop after headers ? Should it error ? But in H1 we didn't error, we only had to possibly close the connection after headers in such a case (if it was ever detected).

wtarreau (commented):
Mark, if you change this one, I think you'd also need to update the wording of HEAD and CONNECT for consistency, around line 1837.

PR #466, opened 2020-10-09
Tighten up Vary language

### LINKED ISSUE #459: Selecting from multiple responses
> If multiple selected responses are available (potentially including responses without a Vary header field), the cache will need to choose one to use. When a selecting header field has a known mechanism for doing so (e.g., qvalues on Accept and similar request header fields), that mechanism may be used to select preferred responses; of the remainder, the most recent response (as determined by the Date header field) is used, as per Section 4.
>
> Note that in practice, some resources might send the Vary header field on responses inconsistently. When a cache has multiple responses for a target URI, and one or more omits the Vary header field, it should use the most recent non-empty value available to select an appropriate response for the request.

This language is a bit ambiguous regarding whether the received response is included in those considered, etc.

mnot:
Fixes #459

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -597,7 +597,7 @@
    Note that in practice, some resources might send the Vary header field on
    responses inconsistently. When a cache has multiple responses for a
    target URI, and one or more omits the Vary header field, it &SHOULD; use the
-   most recently received non-empty Vary field value available to select an appropriate response
+   most recent (see <xref target="age.calculations"/>) non-empty Vary field value available to select an appropriate response
    for the request.
 </t>
 <t>
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -597,7 +597,7 @@
    Note that in practice, some resources might send the Vary header field on
    responses inconsistently. When a cache has multiple responses for a
    target URI, and one or more omits the Vary header field, it &SHOULD; use the
-   most recent (see <xref target="age.calculations"/>) non-empty Vary field value available to select an appropriate response
+   most recent (see <xref target="age.calculations"/>) valid Vary field value available to select an appropriate response
    for the request.
 </t>
 <t>
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -594,9 +594,11 @@
    per <xref target="constructing.responses.from.caches"/>.
 </t>
 <t>
-   Note that in practice, some resources might send the Vary header field on
-   responses inconsistently. When a cache has multiple responses for a
-   target URI, and one or more omits the Vary header field, it &SHOULD; use the
+   Some resources mistakenly omit the Vary header field from their default
+   response (i.e., the one sent when no more preferable response is available), 
+   selecting it for requests to that resource even when 
+   more preferable responses are available. When a cache has multiple responses for a
+   target URI and one or more omits the Vary header field, it &SHOULD; use the
    most recent (see <xref target="age.calculations"/>) valid Vary field value available to select an appropriate response
    for the request.
 </t>
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -589,7 +589,7 @@
    responses without a Vary header field), the cache will need to choose one to use.
    When a selecting header field has a known mechanism for doing so (e.g., qvalues on
    <x:ref>Accept</x:ref> and similar request header fields), that mechanism &MAY; be
-   used to select preferred responses; of the remainder, the most recent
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -589,15 +589,15 @@
...
+   used to select a preferred response. If such a mechanism is not available, or leads to equally preferred responses, the most recent
    response (as determined by the <x:ref>Date</x:ref> header field) is used, as
    per <xref target="constructing.responses.from.caches"/>.
 </t>
 <t>
    Note that in practice, some resources might send the Vary header field on
    responses inconsistently. When a cache has multiple responses for a
    target URI, and one or more omits the Vary header field, it &SHOULD; use the
-   most recent non-empty value available to select an appropriate response
+   most recently received non-empty Vary field value available to select an appropriate response
```
royfielding, line 600: I don't think this change is correct -- "recent" includes adjustment by Age, whereas most recently received means something else.

I also don't think the original requirement is correct, but it may reflect an old argument I lost long ago. Basically, the notion that an old Vary field is more relevant than a newer response without Vary is obviously wrong as soon as a resource stops varying.

royfielding (changes requested):

draft-ietf-httpbis-cache-latest.xml
mnot:
```suggestion
   most recent (see <xref target="age.calculations"/>) non-empty Vary field value available to select an appropriate response
```

mnot: @royfielding See ^^

reschke: FWIW: the change about recent *non-empty* was added in https://github.com/httpwg/http-core/commit/b9789cb51370caa6c9c73d9eb70806691b3285e1#diff-c2c4977337243f1b71ca2e3fd5b2e00dc95ebf32f8f63594e3b85eb21df30230 - I'm not sure whether this change was correct, but if we stick with it we may want to mention it in "changes from 7234".

reschke: FWIW, the addition of

> 
"Note that in practice, some resources might send the Vary header field on responses inconsistently. When a cache has multiple responses for a given target URI, and one or more omits the Vary header field, it SHOULD use the most recent non-empty value available to select an appropriate response for the request."

happenen in draft 06. Isn't this a normative change that needs to be listed under "changes from RFC 7234"?

royfielding: I think it is flat out wrong and should be reverted. An empty Vary is just as valid as a full Vary and that shouldn't affect which valid stored response is chosen.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -589,15 +589,15 @@
...
+   used to select a preferred response. If such a mechanism is not available, or leads to equally preferred responses, the most recent
    response (as determined by the <x:ref>Date</x:ref> header field) is used, as
    per <xref target="constructing.responses.from.caches"/>.
 </t>
 <t>
    Note that in practice, some resources might send the Vary header field on
    responses inconsistently. When a cache has multiple responses for a
    target URI, and one or more omits the Vary header field, it &SHOULD; use the
-   most recent non-empty value available to select an appropriate response
+   most recent (see <xref target="age.calculations"/>) non-empty Vary field value available to select an appropriate response
```
mnot, line 600:
```suggestion
   most recent (see <xref target="age.calculations"/>) valid Vary field value available to select an appropriate response
```

mnot:
Need to talk about the on-the-wire scenarios; representation w/ vary vs without.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -589,15 +589,15 @@
...
    <x:ref>Accept</x:ref> and similar request header fields), that mechanism &MAY; be
-   used to select preferred responses; of the remainder, the most recent
+   used to select a preferred response. If such a mechanism is not available, or leads to equally preferred responses, the most recent
    response (as determined by the <x:ref>Date</x:ref> header field) is used, as
    per <xref target="constructing.responses.from.caches"/>.
 </t>
 <t>
    Note that in practice, some resources might send the Vary header field on
    responses inconsistently. When a cache has multiple responses for a
    target URI, and one or more omits the Vary header field, it &SHOULD; use the
```
mnot, line 599:
```suggestion
   Note that in practice, some resources might send the Vary header field on
   responses inconsistently; often, when the resource mistakenly omits it
   from the default response, causing it to be selected for all requests to that
   resource, even when other more preferable responses are available.   
   When a cache has multiple responses for a
   target URI, and one or more omits the Vary header field, it &SHOULD; use the
```

mnot:
When making the edit above, I started to wonder what the best strategy was to recommend:

* Use the most recent valid Vary field value (current text)
* Just give the one without a Vary field value the lowest priority

Thoughts @reschke @royfielding ?

royfielding (approved):

reschke (approved):

PR #468, opened 2020-10-09
Validation Failures (Mnot 462)

### LINKED ISSUE #462: Validation failures
[4.3.3 says](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#validation.response):

> However, if a cache receives a 5xx (Server Error) response while attempting to validate a response, it can either forward this response to the requesting client, or act as if the server failed to respond. In the latter case, the cache may send a previously stored response (see Section 4.2.4).

I think that MAY should be prose, and should also list retrying the request as an option.

mnot:
Fixes #462.


reschke (approved):

royfielding (approved):

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1038,7 +1038,7 @@
      forward this response to the requesting client, or act as if the
      server failed to respond. In the latter case, the cache &MAY; send a
```
reschke, line 1039:
```suggestion
     server failed to respond. In the latter case, the cache can send a
```

mnot, line 1039:
```suggestion
     server failed to respond. In the latter case, the cache can send a
```

PR #469, opened 2020-10-09
Constrain field combination to a section

### LINKED ISSUE #454: Combining field values across sections
The practice of combining list-form fields into a single value is widespread, and the document addresses this pretty clearly.  However it never really says that this combination needs to be limited to a single field section.  It's pretty clear that trailers need to be kept discrete, but there is nothing so direct.

It's probably worth adding "within a field section" somewhere.

mnot:
Fixes #454.


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1451,9 +1451,11 @@ Content-Type: text/plain
 <section title="Header Fields" anchor="header.fields">
 <t>
    <iref item="field"/>
+   <iref item="section"/>
    HTTP messages use key/value pairs to convey data about the
    message, its payload, the target resource, or the connection.
-   They are called "HTTP fields" or just "<x:dfn>fields</x:dfn>".
+   They are called "HTTP fields" or just "<x:dfn>fields</x:dfn>" that
+   occur in groups called "<x:dfn>field sections</x:dfn>" or just "sections".
```
martinthomson, line 1458:
```suggestion
   They are called "HTTP fields" or just "<x:dfn>fields</x:dfn>".
   Fields occur in groups called "<x:dfn>field sections</x:dfn>" or just "sections".
```

martinthomson (approved):
Thanks, this works nicely.  Shame that you had to add a new term, but we already got that in QUIC.

reschke (approved):

royfielding (approved):

PR #479, opened 2020-10-22
Correct rules considered case-insensitive in dates

### LINKED ISSUE #473: Date case sensitivity
Caching 4.2:

> Although all date formats are specified to be case-sensitive, a cache recipient should match day, week, and time-zone names case-insensitively.

What does "week" refer to in this text? Also, why is this separate from the text on HTTP-dates? Do we really mean that caches should parse them in a different fashion than other components?

mnot:
And mention in Semantics.

Fixes #473.

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -681,7 +681,7 @@
 </t>
 <ul>
    <li>Although all date formats are specified to be case-sensitive, 
-      a cache recipient &SHOULD; match the day-name, day-name1, month and GMT rules
+      a cache recipient &SHOULD; match the field value
       case-insensitively.</li>
    <li>If a cache recipient's internal implementation of time has less
       resolution than the value of an HTTP-date, the recipient &MUST;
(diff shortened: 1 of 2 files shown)


reschke (commented):
will add some links...

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -681,7 +681,7 @@
 </t>
 <ul>
    <li>Although all date formats are specified to be case-sensitive,
```
reschke, line 683: AFAIU, we now list all parts of the ABNF that are not digits or delimiters. So why not simply match the whole string case-insensitively?

reschke (changes requested):

draft-ietf-httpbis-cache-latest.xml
mnot: Fair point.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -681,7 +681,7 @@
 </t>
 <ul>
    <li>Although all date formats are specified to be case-sensitive, 
-      a cache recipient &SHOULD; match day, week, and time-zone names
+      a cache recipient &SHOULD; match the day-name, day-name1, month and GMT rules
```
reschke, line 684:
```suggestion
      a cache recipient &SHOULD; match the field value
```

reschke (approved):

PR #480, opened 2020-10-22
Define request collapsing

### LINKED ISSUE #475: Request collapsing
We should really mention this, as it's common and can cause misunderstanding. I also am likely to need to reference it in #474.

mnot:
Fixes #475

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -532,7 +532,7 @@
    see <xref target="invalidation" />.
 </t>
 <t>
-   A response that is stored or can be stored can be used to satisfy multiple
+   A response that is stored or storable can be used to satisfy multiple
    requests, provided that it is allowed to reuse that response for the requests
    in question. This enables caches to "collapse" multiple requests into a single forward
    request upon a cache miss, thereby reducing load on the origin server and network.
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -532,6 +532,15 @@
    see <xref target="invalidation" />.
 </t>
 <t>
+   A response that is stored or can be stored can be used to satisfy multiple
+   requests, provided that it is allowed to reuse that response for the requests
+   in question. This enables caches to "collapse" multiple requests into a single forward
+   request upon a cache miss, thereby reducing load on the origin server and network.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -531,6 +531,15 @@
    Also, note that unsafe requests might invalidate already-stored responses;
    see <xref target="invalidation" />.
 </t>
+<t>
+   A response that is stored or can be stored can be used to satisfy multiple
+   requests, provided that it is allowed to reuse that response for the requests
```
reschke, line 536: Maybe "A response that is stored or storable can be used to satisfy multiple..." ?

reschke (changes requested):

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -531,6 +531,15 @@
    Also, note that unsafe requests might invalidate already-stored responses;
    see <xref target="invalidation" />.
 </t>
+<t>
+   A response that is stored or can be stored can be used to satisfy multiple
```
mnot, line 535:
```suggestion
   A response that is stored or storable can be used to satisfy multiple
```

reschke (approved):

reschke (commented):
Maybe add a minimal explanation of "collapse"?

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -531,6 +531,15 @@
    Also, note that unsafe requests might invalidate already-stored responses;
    see <xref target="invalidation" />.
 </t>
+<t>
+   A response that is stored or storable can be used to satisfy multiple
+   requests, provided that it is allowed to reuse that response for the requests
+   in question. This enables caches to "collapse" multiple requests into a single forward
```
mnot, line 537: add a parenthetical more fully describing what's happening

royfielding (approved):

draft-ietf-httpbis-cache-latest.xml
mnot: Looking at this again, "multiple requests into a single forward request" says all that needs to be, I think. I will change "requests" to "incoming requests" to clarify.

mnot:
```suggestion
   in question. This enables caches to "collapse" multiple incoming requests into a single forward
```

PR #482, opened 2020-10-22
Clarify duplicate and invalid freshness info

### LINKED ISSUE #460: Handling duplicate directives (part II)
4.2.1 says:

> When there is more than one value present for a given directive (e.g., two Expires header fields, multiple Cache-Control: max-age directives), the directive's value is considered invalid. Caches are encouraged to consider responses that have invalid freshness information to be stale.

What about multiple `no-store`?

mnot:
Fixes #460


reschke (changes requested):
Do we need to note that as normative change?

reschke (approved):

royfielding (approved):

PR #485, opened 2020-10-22
Clarify Age error handling

### LINKED ISSUE #471: Revisiting Age error handling
My [tests for Age parsing](https://cache-tests.fyi/#age-parse) seem to indicate we might want to reconsider how [errors are handled](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#field.age).

Of implementations that honour `Age` (two don't):

- none will consider `0, 0` invalid
- none will consider `0` on two separate field lines invalid
- none will consider a duplicated value on two lines to make the header invalid
- if there are two `Age` header lines, all will consider only the first (and one, Safari, appears to use the min of them)
- none considers a non-numeric value to be invalid
- only Fastly and Firefox consider a non-negative integer to be invalid

... where invalid currently is required to be considered stale.

mnot:
Fixes #471

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1243,9 +1243,17 @@
 </sourcecode>
 <t>
    The Age field value is a non-negative integer, representing time in seconds
-   (see <xref target="delta-seconds"/>). A cache &SHOULD; consider a response
-   to be stale if an Age field is present and its value is invalid (i.e.,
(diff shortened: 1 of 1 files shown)


reschke (changes requested):
This needs to be listed as normative change (IMHO).

mnot:
@reschke It already is, this just changes how it's done.

reschke (approved):

PR #487, opened 2020-10-22
Talk about field backwards compatibility

### LINKED ISSUE #448: Defining new preconditions
Some guidance about how and if new preconditions can be defined would be helpful.

See [mailing list thread](https://www.w3.org/mid/7c1e8964-1a4b-42e6-8af1-fc7a4e43ce14@www.fastmail.com).

mnot:
Fixes #448


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9717,9 +9717,33 @@ Content-Type: text/plain
...
+   by recipients without previous arrangement. For example, a new request
+   precondition cannot be assumed to be supported by servers; a server that
+   has not yet implemented it will ignore the field and process the request
+   as usual. Therefore, a function that uses new fields needs to either
+   degrade gracefully when handled by a recipient who does not understand
+   them, or use negotiation mechanism to establish whether the recipient
+   supports the fields in question. Typically, this will be out of band; for
+   example, using a particular media type might imply support for associated
+   header fields, or interacting with a defined well-known URI <xref
+   target="RFC8615"/> might do so.
```
reschke, line 9744: Other examples: response header fields (Accept-Ranges, Accept-CH), and OPTIONS.

mnot: Could you say a bit more?

reschke: Gotcha. So I happen to disagree with "typically out of band" because there are two IHMO widely used ways which are in band (OPTIONS and response header fields).

mnot: Fair enough, will just list both in-band and out-of-band options...

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9719,9 +9719,33 @@ Content-Type: text/plain
...
```
royfielding, line 9746:
```suggestion
   However, defining a field doesn't guarantee its deployment or recognition
   by recipients. Most fields are designed with the expectation that a recipient
   can safely ignore (but forward downstream) any field not recognized.
   In other cases, the sender's ability to understand a given field might be
   indicated by its prior communication, perhaps in the protocol version
   or fields that it sent in prior messages, or its use of a specific media type.
   Likewise, direct inspection of support might be possible through an
   OPTIONS request or by interacting with a defined well-known URI
   <xref target="RFC8615"/> if such inspection is defined along with
   the field being introduced.
```

royfielding (changes requested):

mnot:
I believe that we said this could be merged when Julian's issue was addressed, which Roy's suggestion did, so merging.

PR #490, opened 2020-10-23
Refine how headers are excepted from update on 304

### LINKED ISSUE #488: Updating stored headers II
Following #165, I updated tests to match the [new requirements](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#freshening.responses) and the [results are not encouraging](https://cache-tests.fyi/?suite=update304).

Ignoring implementations that obviously don't support updating stored headers (nuster, Fastly, nginx), the new requirement:

> Caches MAY update the following header fields: Content-Encoding, Content-Length, Content-MD5 (Section 14.15 of [RFC2616]), Content-Range, ETag

has the effect of making browsers more conferment, but other implementations less conformant (because they do update those headers).

I think the most rational approach to address this would be to soften and contextualise the requirement, something like:

> In some cases, caches (especially in user agents) store processed representations of the received response, rather than the response itself, and updating header fields that affect that processing can result in inconsistent behaviour and security issues. Caches in this situation MAY omit these header fields from updating stored representations on an exceptional basis, but SHOULD limit such omission to those fields necessary to assure integrity of the stored representation. 
>
> For example, a browser cache might store a response after removing content-codings; updating that response with a different `Content-Encoding` header field would be problematic. Likewise, a browser cache might store a post-parse tree representation of HTML, rather than the bytes received in the response; updating the `Content-Type` header field would not be workable in this case.
>
> Furthermore, some fields are automatically processed and removed by the HTTP implementation; for example, the `Content-Range` header field. Implementations MAY automatically omit such header fields from updates, even when the processing does not actually occur.
>
> Finally, the `Content-Length` header field MUST NOT be updated, as this can cause security issues (see [ref]).
>
> Note that the `Content-*` prefix is not a signal that a header is omitted from update; it is only a convention for content-related fields.

Note that this doesn't mention `ETag` or `Last-Modified`, because if they have a different value, the response won't be selected for update.

mnot:
Fixes #488.


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1080,13 +1080,38 @@
...
-   target="storing.fields"/> also apply to header field updates.</li>
-   <li>Caches &MUST-NOT; update the following header fields:
-   <x:ref>Content-Encoding</x:ref>, <x:ref>Content-Length</x:ref>,
-   Content-MD5 (<xref target="RFC2616" x:fmt="of" x:sec="14.15"/>),
-   <x:ref>Content-Range</x:ref>, <x:ref>ETag</x:ref>.</li>
+   <li>Header fields excepted from storage in <xref target="storing.fields"/>,</li>
+   <li>Header fields that the cache's stored representation of the response depends upon, as described below,</li>
+   <li>Header fields that are automatically processed and removed by the recipient, as described below, and</li>
+   <li>The Content-Length header field.</li>
 </ul>
```
reschke, line 1087: I'm a bit nervous that we seem to loose a concrete list of field names in a normative requirement.

reschke (changes requested):

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1080,13 +1080,38 @@
...
+<t>
+   In some cases, caches (especially in user agents) store processed
+   representations of the received response, rather than the response itself,
+   and updating header fields that affect that processing can result in
+   inconsistent behaviour and security issues. Caches in this situation &MAY;
```
royfielding, line 1092:
```suggestion
   inconsistent behavior and security issues. Caches in this situation &MAY;
```

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1080,13 +1080,38 @@
...
+   necessary to assure integrity of the stored representation.
+</t>
+<t>
+   For example, a browser cache might store a response after removing
+   content-codings; updating that response with a different Content-Encoding
+   header field would be problematic. Likewise, a browser cache might store a
+   post-parse tree representation of HTML, rather than the bytes received in
+   the response; updating the Content-Type header field would not be workable
+   in this case.
+</t>
```
royfielding, line 1104: I feel this is a bit confusing, since the browser is usually not acting as a cache when it stores data transformed from a response. Why is this part relevant to HTTP?

A better way to think of it is the original response exists with its data portion transformed and stored by reference. The data is no longer with the response; HTTP does not care because this is a user agent. However, the metadata (and anything else the browser chose to store about its original request/response) is stored like a cached response and can be validated/updated like a cached response. A browser can use that stored metadata and request context when choosing to update (or not) its stored response, whether or not the received representation was transformed, provided that validation uses the original (pre-transformed) response fields.

royfielding: Hence, when a successful validation response (e.g., 304) is received with new field information, the browser needs to interpret that information with respect to the original request and its original metadata.

royfielding (changes requested):

draft-ietf-httpbis-cache-latest.xml
royfielding: IOW, all cache validations are based on field metadata; no cache operates by sending the data stream back to the server to see if it is still valid. Hence, how the response data is processed or stored after the response is received is not relevant to HTTP so long as the recipient keeps the original metadata for later validation requests.

draft-ietf-httpbis-cache-latest.xml
mnot: Discussed on call.

draft-ietf-httpbis-cache-latest.xml
mnot:
```suggestion
<t>
   For example, a browser might store a response's body after removing
   content-codings, thereby making its metadata inaccurate. 
   Updating that stored metadata with a different Content-Encoding
   header field would be problematic. Likewise, a browser might store a
   post-parse tree representation of HTML, rather than the body received in
   the response; updating the Content-Type header field would not be workable
   in this case, because any assumptions about the format made in parsing would
   now be invalid.
</t>
```

@royfielding PTAL

royfielding (approved):

Issue #496, opened 2020-10-23
5.1. Protocol Version


mnot:
This seems out of place, although I understand that versioning is very message-based.

I'd suggest putting it in 15. Extending HTTP?

royfielding:
It's not about extending HTTP. The version number literally exists as a field of the message, or at least inherited by the context of the connection, and must be retained by recipients because of version-specific requirements. That's why we need to define both cases as part of the message abstraction.

mnot:
Fair enough. Can we open the section with a sentence or two that says that it's a field of the message, to give that context? The current text is pretty abstract.

PR #498, opened 2020-10-23
ref https://www.w3.org/Protocols/HTTP/AsImplemented.html (#496)

reschke:


reschke:
@ylafon - is Oct 1999 correct? (taken from Last-Modified)

ylafon:
Last edit is 1999 but was just a markup fix, original document is dated 1996-10-03.
HTH,

PR #518, opened 2020-11-20
   Relax 60-second for LM strong compare

### LINKED ISSUE #510: The 60 second strong Last-Modified window
[7.9.2.2. Comparison](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#lastmod.comparison): 

>A Last-Modified time, when used as a validator in a request, is implicitly weak unless it is possible to deduce that it is strong, using the following rules:
>
...
> * The validator is about to be used by a client in an If-Modified-Since, If-Unmodified-Since, or If-Range header field, because the client has a cache entry for the associated representation, and
> * That cache entry includes a Date value, which gives the time when the origin server sent the original response, and
> * The presented Last-Modified time is at least 60 seconds before the Date value.

I'm hearing that this 60-second window is problematic for videos, where clients often request a partial video segment within a minute of it being published (thanks to e.g., CMAF). Since those requests use `If-Range`, they require a strong validator.

Should we revisit this, given that it is an arbitrarily chosen value by our own admission? I agree that some caution is necessary here, but a one-size-fits-all approach doesn't seem viable.

mnot:
Fixes #510


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3964,19 +3970,25 @@ Upgrade: websocket
...
-   arbitrary 60-second limit guards against the possibility that the Date and
+   have a <x:ref>Date</x:ref> value equal to its Last-Modified time. Previous
+   revisions of this specification imposed an
+   arbitrary 60-second limit to guard against the possibility that the Date and
    Last-Modified values are generated from different clocks or at somewhat
-   different times during the preparation of the response. An
-   implementation &MAY; use a value larger than 60 seconds, if it is
-   believed that 60 seconds is too short.
+   different times during the preparation of the response. This specification
+   has relaxed that to allow reasonable discretion.
```
royfielding, line 3991: I would move the last two sentences to the changes from 7232 section to make that section a little more clear, rather than leave them here. It isn't going to help new implementations.

royfielding: Also, check the build failing. It might be due to nested list needing different tags in rfc xml.

royfielding (changes requested):

draft-ietf-httpbis-semantics-latest.xml
royfielding: IIRC, the inner <ul> has to be outside the <li>, maybe.

royfielding (approved):

PR #519, opened 2020-11-20
Move example req/res into the intro

### LINKED ISSUE #492: 3.6. Example Request and Response
seems very out of place. 

I could almost see it as content in 3, as a kind of introduction to the terminology. There are of course other places it could go, but the current location doesn't make much sense.

mnot:
Not perfect, but better

Fixes #492

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -273,7 +273,7 @@
                                 &lt;   response
 </artwork></figure>
 <t>
-   The following example illustrates a typical message exchange for a
+   The following example illustrates a typical HTTP/1.1 message exchange for a
    GET request (<xref target="GET"/>) on the URI "http://www.example.com/hello.txt":
 </t>
 <t>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -260,6 +260,50 @@
...
+   simplest case, this might be accomplished via a single bidirectional
+   connection (===) between the user agent (UA) and the origin server (O).
+</t>
+<figure><artwork type="drawing">
+         request   &gt;
+    <x:highlight>UA</x:highlight> ======================================= <x:highlight>O</x:highlight>
+                                &lt;   response
+</artwork></figure>
+<t>
+   The following example illustrates a typical message exchange for a
```
martinthomson, line 276: Worth qualifying?

```suggestion
   The following example illustrates a typical HTTP/1.1 message exchange for a
```

reschke (approved):

royfielding:
I really don't see how this helps anything. The reason we define all those terms in section 2 is so that we can use them like we do in this example.

If the example doesn't fit in section 2, then we should move it to Messaging.

mnot:
Discussed on call; either remove the section heading but leave in situ, or move message examples (not flow diagram) to new section at bottom of terminology.

PR #526, opened 2020-11-20
Factor out Updating Stored Headers and use in Combining Responses

### LINKED ISSUE #458: Combining Partial Content
3.4 says:

> When combining the new response with one or more stored responses, a cache must use the header fields provided in the new response, aside from Content-Range, to replace all instances of the corresponding header fields in the stored response.

As per other discussions, this needs to be more nuanced; either a reference to 4.3.4, or new text.

mnot:
Now that we have three different places referring to the text on updating stored headers, it deserves its own section.

In the process, I noticed that the current text forgot to mention that CC: private and no-cache can have arguments that affect header storage.

Fixes #458.

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -421,6 +421,55 @@
 </t>
 </section>
 
+<section title="Updating Stored Header Fields" anchor="update">
+<t>
+   Caches are required to update a stored response's header fields from another
+   (typically newer) response in several situations; for example, see <xref
+   target="combining.responses"/>, <xref target="freshening.responses"/> and
+   <xref target="head.effects"/>.
+</t>
+<t>
+   When doing so, the cache &MUST; use the header fields in the provided
+   response to replace all instances of the corresponding header fields in
+   the stored response, with the following exceptions:
+</t>
+<ul>
+   <li>Header fields excepted from storage in <xref target="storing.fields"/>,</li>
+   <li>Header fields that the cache's stored representation of the response depends upon, as described below,</li>
+   <li>Header fields that are automatically processed and removed by the recipient, as described below, and</li>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -408,6 +408,10 @@
    before forwarding the message, and this &MAY; be implemented by doing so
    before storage; see <xref target="field.connection"/> for some
    examples.</li>
+   <li>The no-cache (<xref target="cache-response-directive.no-cache"/>) and
+   private (<xref target="cache-response-directive.private"/>) cache
+   directives can have arguments that prevent storage of header fields by all
```
reschke, line 413:
```suggestion
   directives can have arguments that prevent storage of fields by all
```

(that's actually a general question; does caching only apply to *header* fields???)

mnot: It's just header fields in this case; see https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#storing.fields and the directive definitions.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -421,6 +425,55 @@
...
+</t>
+<t>
+   Furthermore, some fields are automatically processed and removed by the
+   HTTP implementation; for example, the Content-Range header field.
+   Implementations &MAY; automatically omit such header fields from updates,
+   even when the processing does not actually occur.
+</t>
+<t>
+   Note that the Content-* prefix is not a signal that a header field is omitted
+   from update; it is only a convention for naming content-related fields.
```
mnot, line 473:
```suggestion
   from update; it is a convention for MIME header fields, not HTTP.
```

royfielding (approved):

PR #533, opened 2020-11-25
Try moving it to the bottom of terminology.

### LINKED ISSUE #492: 3.6. Example Request and Response

mnot:
Fixes #492.


royfielding (approved):
I moved the ascii art diagram back to the origin server section (where it came from in 7230) since it leads into the other diagrams.

reschke (approved):

PR #536, opened 2020-11-26
move transfer-coding ABNF to Semantics and clean it up

reschke:


royfielding:
I think we discussed this last week and Julian has pending changes to be made before review.

royfielding (approved):
Fine to merge, but shouldn't we also move the extensibility and IANA registry sections?

reschke:
This just moves the ABNF, not the definition of the transfer codings.

Issue #532, opened 2020-11-23
"Messages": tune intro


reschke:
We currently say:

"The request might also contain header fields (Section 5.4) for request modifiers, client information, and representation metadata, a payload body (Section 5.5.4) to be processed in accordance with the method, and trailer fields (Section 5.6) for metadata collected while sending the payload."

I think it would be better to say:

"The request might also contain header fields (Section 5.4) for request modifiers, client information, and representation metadata, a payload (Section 5.5) to be processed in accordance with the method, and trailer fields (Section 5.6) for metadata collected while sending the payload."

so s/payload body (Section 5.5.4)/payload (Section 5.5)/

royfielding:
but the "payload" is both some of the metadata and the transfer-decoded body.

reschke:
...but it really reads strange to reference 5.4, 5.5.4, and 5.6...

royfielding:
This is going to be fixed as part of #553

PR #546, opened 2020-12-02
define the Message Abstraction as an ADT rather than as a message

### LINKED ISSUE #540: Semantics: "5.3. Header Fields" -> "5.3. Header and Trailer Fields"?

royfielding:
This fixes #540 and a lot more in the process.

- move Fields to its own section before the message abstraction;
- define the Message Abstraction as an ADT so that it makes sense regardless of the message syntax;
- define and forward reference the control data;
- move protocol version definition up and normative requirements to control data;
- clearly define the header and trailer lookup tables as abstractions and the sections as where to look for those fields;
- define the payload and clean up old mentions of message body and payload body;
- merge paragraphs that had duplicate definitions in trailers and payload.

Normally I would separate the commit into moves first and then changes, but I have run out of time and would rather have reviews of the end result. We can do that in a separate PR if needed.


royfielding:
also fixes #532

reschke (commented):
This is indeed hard to review.

Should I try to split this into separate PRs?

royfielding:
Yeah, the diff is hopeless. I will start breaking it out into separate pieces.

royfielding:
Replaced by #551 + #554 + #555 + #557

Issue #550, opened 2020-12-02
move protocol version definition up and normative requirements to control data


royfielding:
As @mnot has mentioned, the protocol version section sticks out in the message abstraction. I found it works better if split into two sections and definitions moved earlier in the document. Then the protocol version requirements on messaging can be called out in a section on control data, which I forgot to fill out.

PR #551, opened 2020-12-02
move protocol version section up to conformance

royfielding:
and only define the versioning there; leave message version requirements in control data.

(no content changes)


royfielding:
fixes #550

Issue #294, opened 2020-02-03
Description and use of protocol layers


mnot:
Regarding representation, message, payload, selected representation, etc. -- we could do a better job of conveying these concepts, nearly everyone finds them confusing.

See also #163.

mnot:
Waiting for #368, want to see what that looks like first.

mnot:
I think #672 will cover this sufficiently.

PR #557, opened 2020-12-03
improve message abstraction to reflect semantics after message is parsed

### LINKED ISSUE #556: improve message abstraction to reflect semantics after message is parsed
I found it easier to describe the different message sections if we talk about the abstract data structures produced as a result of parsing, as opposed to the specific parsing mechanisms of 1.1.

royfielding:
Fixes #556

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -2166,7 +2166,7 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
 </t>
 <t>
    The "<x:dfn>header section</x:dfn>" of a message consists of a sequence of
-   of header fields. Each header field might modify or extend message
+   of header field lines. Each header field might modify or extend message
    semantics, describe the sender, define the payload, or provide additional
    context.
 </t>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -13039,6 +13039,16 @@ transfer-coding = &lt;transfer-coding, see <xref target="Messaging" x:fmt="," x:
    (<xref target="parameter"/>)
 </t>
 <t>
+   An abstract data type for HTTP messages has been introduced to define the
+   components of a message and their semantics as an abstraction across
+   multiple HTTP versions, rather than in terms of the specific syntax form of
+   HTTP/1.1 in <xref target="Messaging"/>, and reflect the contents after the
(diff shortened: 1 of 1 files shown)


royfielding:
This is also related to #294

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2105,22 +2158,24 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
...
 </t>
 <t>
-   <iref item="header section"/>
-   Fields that are sent/received before the message body are referred to as
-   "header fields" (or just "headers", colloquially) and are located within the
-   "<x:dfn>header section</x:dfn>" of a message. We refer to some named fields
-   specifically as a "header field" when they are only allowed to be sent in
-   the header section.
+   The "<x:dfn>header section</x:dfn>" of a message consists of a sequence of
+   of header fields. Each header field might modify or extend message
```
mnot, line 2169:
```suggestion
   of header field lines. Each header field might modify or extend message
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2014,24 +2023,48 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
...
+   obtained while sending the payload.
+</t>
 <t>
-   Each major version of HTTP defines its own syntax for the communication of
-   messages. However, they share a common data abstraction.
+   Framing and control data is sent first, followed by a header section
+   containing fields for the headers table. When a message includes a payload,
+   the payload data is sent after the header section and potentially
+   interleaved with zero or more trailer sections containing fields for the
+   trailers table.
```
mnot, line 2050: I think we should explicitly note that trailers are constrained in some protocol versions.

mnot (changes requested):
Left a couple of notes, otherwise looks good.

draft-ietf-httpbis-semantics-latest.xml
royfielding: We do that in the section on Trailers, to the extent they are constrained. HTTP/2 and 3 do not explicitly constrain trailers, they just don't have a frame to carry them yet.

royfielding: https://github.com/httpwg/http-core/pull/557/commits/3a96888c38b67b28e045f75fc111ac38fb7e5fb7

mnot (approved):

reschke (approved):

PR #575, opened 2020-12-08
add :authority pseudo-header to Host discussion and make section applicable to both

### LINKED ISSUE #511: Semantics talking about Host header field

royfielding:
fixes #511


mnot (approved):

reschke (approved):

kazuho (commented):
Thank you for the changes!

PR #576, opened 2020-12-08
clarify the payload in 200 (OK)

### LINKED ISSUE #572: "15.3.1. 200 OK" representation data
"HEAD: the same representation as GET, but without the representation data;"

Should that be "payload data"?

royfielding:
by changing DL to a table and tweaking description of HEAD. Fixes #572

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -8149,7 +8149,7 @@ Content-Range: exampleunit 11.2-14.3/25
     </tr>
     <tr>
       <td>OPTIONS</td>
-      <td>a representation of the communications options</td>
+      <td>a representation of communication options for the target resource</td>
     </tr>
     <tr>
       <td>TRACE</td>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -8116,8 +8116,8 @@ Content-Range: exampleunit 11.2-14.3/25
 <table align="left">
   <thead>
     <tr>
-      <th>method</th>
-      <th>meaning of the payload</th>
+      <th>request method</th>
+      <th>meaning of the response payload</th>
     </tr>
   </thead>
   <tbody>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -8125,36 +8125,35 @@ Content-Range: exampleunit 11.2-14.3/25
   <thead>
     <tr>
       <th>request method</th>
-      <th>meaning of the response payload</th>
+      <th>response payload is a representation of</th>
     </tr>
   </thead>
   <tbody>
     <tr>
       <td>GET</td>
-      <td>a representation of the <x:ref>target resource</x:ref></td>
+      <td>the <x:ref>target resource</x:ref></td>
     </tr>
     <tr>
       <td>HEAD</td>
-      <td>the same representation as GET, but without transferring the
-          representation data</td>
+      <td>the <x:ref>target resource</x:ref>, like GET, but without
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8113,33 +8113,43 @@ Content-Range: exampleunit 11.2-14.3/25
...
-  <dd>
-    a representation of the request message as received by the
-    end server.
-  </dd>
-</dl>
+<table align="left">
+  <thead>
+    <tr>
+      <th>method</th>
+      <th>meaning of the payload</th>
```
mnot, line 8120:
```suggestion
      <th>request method</th>
      <th>meaning of the response payload</th>
```

Just so we're super-clear.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8113,33 +8113,43 @@ Content-Range: exampleunit 11.2-14.3/25
...
+      <td>the same representation as GET, but without transferring the
+          representation data</td>
+    </tr>
+    <tr>
+      <td>POST</td>
+      <td>a representation of the status of, or results obtained from,
+          the action</td>
+    </tr>
+    <tr>
+      <td>PUT, DELETE</td>
```
mnot, line 8139: Since it's a table, shouldn't these be in separate rows to aid scanning?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8113,33 +8113,43 @@ Content-Range: exampleunit 11.2-14.3/25
...
+      <td>a representation of the status of the action</td>
+    </tr>
+    <tr>
+      <td>OPTIONS</td>
+      <td>a representation of the communications options</td>
+    </tr>
+    <tr>
+      <td>TRACE</td>
+      <td>a representation of the request message as received by the
+          end server</td>
```
mnot, line 8149: "end server" seems a bit odd here, and I don't think it's a term we use anywhere else. Perhaps "server performing the trace"?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8113,33 +8113,43 @@ Content-Range: exampleunit 11.2-14.3/25
...
+      <td>a representation of the status of, or results obtained from,
+          the action</td>
+    </tr>
+    <tr>
+      <td>PUT, DELETE</td>
+      <td>a representation of the status of the action</td>
+    </tr>
+    <tr>
+      <td>OPTIONS</td>
+      <td>a representation of the communications options</td>
```
mnot, line 8144:
```suggestion
      <td>a representation of the communications options of the target resource</td>
```

mnot (changes requested):
Some nits, otherwise sure.

draft-ietf-httpbis-semantics-latest.xml
reschke: Ha, we can use "rowspan" here :-)

royfielding: I tried it but the rowspan visualization doesn't work for both html and text. It just looks like a bug in the table.

I prefer the existing text to the alternative of repeating the content.

draft-ietf-httpbis-semantics-latest.xml
royfielding:
```suggestion
          server returning the trace</td>
```

draft-ietf-httpbis-semantics-latest.xml
royfielding:
```suggestion
      <td>a representation of communication options for the target resource</td>
```

mnot (approved):

reschke (approved):

PR #596, opened 2020-12-09
move method ABNF one paragraph down (fixes #560)

### LINKED ISSUE #560: 9.1 (Methods) Overview - "method" ABNF
...the one line

    method = token

looks a bit misplaced.

Proposal:

- move it (not sure where), or
- have a preamble stating that this is the syntax definition

reschke:


royfielding (changes requested):
I'd rather move it one paragraph down, if you don't mind. The paragraph below that is explaining why the token is case-sensitive.

royfielding (approved):

PR #601, opened 2020-12-10
Potentially allow Range handling on methods other than GET (fixes #581)

### LINKED ISSUE #581: Can methods other than GET support Range?
14.2:

"A server MUST ignore a Range header field received with a request method other than GET."

16.1.2:

"Likewise, if the new method might have some use for partial response semantics (Section 14.2), it ought to document this, too."

reschke:


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7638,8 +7638,13 @@ bytes=500-700,601-999
...
    support efficient recovery from partially failed transfers and partial
-   retrieval of large representations. A server &MUST; ignore a Range header
-   field received with a request method other than GET.
+   retrieval of large representations.
+</t>
+<t>
+   A server &MUST; ignore a Range header field received with a request method
+   which is unrecognized or for which range handling is not defined (of the
+   methods defined in this specification, only <x:ref>GET</x:ref> defines
+   range handling).
```
royfielding, line 7647:
```suggestion
<t>
   A server &MUST; ignore a Range header field received with a request method
   which is unrecognized or for which range handling is not defined. Of the
   methods defined by this specification, only <x:ref>GET</x:ref> defines
   range handling.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -13125,6 +13130,10 @@ transfer-coding = &lt;transfer-coding, see <xref target="Messaging" x:fmt="," x:
    one range. Moving the extension grammar into range specifiers also allows
    protocol specific to byte ranges to be specified separately.
 </t>
+<t>
+   Is is now possible to define Range handling on methods other than GET. 
+   (<xref target="field.range"/>)
```
royfielding, line 13135:
```suggestion
   Is is now possible to define Range handling on extension methods.
   (<xref target="field.range"/>)
```

royfielding (changes requested):

royfielding (approved):

PR #615, opened 2020-12-17
remove accept-params from Accept field ABNF (fixes #568)

### LINKED ISSUE #568: accept-ext in Accept header field used in practice?
...would be good to have tests...

reschke:


royfielding (changes requested):
Fine except that the second note contains two normative requirements. So, that cannot be a note, but it can be a normal paragraph (probably before the first note).

reschke:
Good point.

royfielding (approved):

mnot:
For #568.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6404,16 +6401,21 @@ Expect: 100-continue
...
-   indicating a relative weight (<xref target="quality.values"/>), and then zero or more extension
-   parameters. The "q" parameter is necessary if any extensions (accept-ext) are present,
-   since it acts as a separator between the two parameter sets.
+   Each media-range might be followed by optional applicable media type
+   parameters (e.g., <x:ref>charset</x:ref>), followed by an optional "q"
+   parameter for indicating a relative weight (<xref target="quality.values"/>).
+</t>
+<t>
+   This specification does not define "accept parameters" following the 
+   special "q" parameter, as was done in earlier editions. Senders using
```
mnot, line 6410: Editions doesn't seem quite right - what language do we use in other places? Revisions?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6404,16 +6401,21 @@ Expect: 100-continue
...
+   weights &SHOULD; still always send "q" last (after all media-range
+   parameters). Recipients &SHOULD; process any parameter named "q" as
+   weight, no matter whether it appears last or not.
 </t>
 <aside>
   <t>
-    &Note; Use of the "q" parameter name to separate media type
-    parameters from Accept extension parameters is due to historical
+    &Note; Use of the "q" parameter name to control conntent negotiation
```
mnot, line 6417:
```suggestion
    &Note; Use of the "q" parameter name to control content negotiation
```

mnot (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6404,16 +6401,21 @@ Expect: 100-continue
...
+   Each media-range might be followed by optional applicable media type
+   parameters (e.g., <x:ref>charset</x:ref>), followed by an optional "q"
+   parameter for indicating a relative weight (<xref target="quality.values"/>).
+</t>
+<t>
+   This specification does not define "accept parameters" following the 
+   special "q" parameter, as was done in earlier editions. Senders using
+   weights &SHOULD; still always send "q" last (after all media-range
+   parameters). Recipients &SHOULD; process any parameter named "q" as
+   weight, no matter whether it appears last or not.
```
royfielding, line 6413:
```suggestion
   weights &SHOULD; send "q" last (after all media-range
   parameters). Recipients &SHOULD; process any parameter named "q" as
   weight, regardless of parameter ordering.
```

royfielding (changes requested):

draft-ietf-httpbis-semantics-latest.xml
royfielding: More like
> Previous specifications of Accept allowed additional extension parameters to appear after the weight parameter. The accept extension grammar has been removed because they had a complicated definition, were not being used in practice, and more easily deployed through new header fields.

royfielding (approved):

PR #616, opened 2020-12-18
note that RFC 1945 also describes HTTP/0.9 (fixes #614)

### LINKED ISSUE #614: HTTP/0.9: ref RFC 1945 as well
...turns out that RFC 1945 defines HTTP/0.9 as well (see https://greenbytes.de/tech/webdav/rfc1945.html#http.version). Maybe we should add this reference, or replace the existing one.

reschke:


royfielding (changes requested):
why do we need to say this? I'd agree if it were to replace the 0.9 reference somehow, but I don't think it needs to be said otherwise.

reschke:
Im ok with replacing the 0.9 reference (although having the link here acknowledges where 0.9 came from, right).

The idea was to mention 0.9 in the context of RFC 1945 as well as that is an IETF document which actually describes what 0.9 *is*.

royfielding (approved):

PR #620, opened 2020-12-31
Ignore content-range in requests

### LINKED ISSUE #618: can Content-Range appear in a request?

mnot:
Fixes #618.

Note outstanding question on the issue regarding PUT.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -7778,6 +7778,11 @@ bytes=500-700,601-999
    A proxy that receives such a message &SHOULD; forward it downstream.
 </t>
 <t>
+   A server &MUST; ignore a Content-Range header field received in a request
+   with a method for which Content-Range support is not defined. No request
+   method in this specification is defined to support Content-Range.
+</t>
+<t>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7777,6 +7777,11 @@ bytes=500-700,601-999
    recipient &MUST-NOT; attempt to recombine it with a stored representation.
    A proxy that receives such a message &SHOULD; forward it downstream.
 </t>
+<t>
+   A server &MUST; ignore a Content-Range header field received in a request
+   with a method for which Content-Range support is not defined. No request
+   method in this specification is defined to support Content-Range.
```
royfielding, line 7783:
```suggestion
   method in this specification is defined to support Content-Range in a request.
```

royfielding (approved):

mnot:
@royfielding any thoughts about whether we should mention PUT?

reschke:
I don't think we need to.

reschke (approved):

royfielding:
I think PUT is already covered in PUT.

Issue #628, opened 2021-01-04
Including


mnot:
In the abstract:

> The Hypertext Transfer Protocol (HTTP) is a stateless application-level protocol for distributed, collaborative, hypertext information systems. This document defines the semantics shared by all versions of HTTP, including its architecture, terminology, core protocol elements, and extensibility mechanisms, along with the "http" and "https" Uniform Resource Identifier (URI) schemes.

"semantics... including its architecture, terminology..." seems incorrect; suggest "semantics... and its architecture..."

PR #630, opened 2021-01-04
Tweak abstract

mnot:
For #628

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -104,7 +104,7 @@
    The Hypertext Transfer Protocol (HTTP) is a stateless application-level
    protocol for distributed, collaborative, hypertext information systems.
    This document defines the semantics shared by all versions of HTTP,
-   including its architecture, terminology, core protocol elements, and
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -104,7 +104,7 @@
    The Hypertext Transfer Protocol (HTTP) is a stateless application-level
    protocol for distributed, collaborative, hypertext information systems.
    This document defines the semantics shared by all versions of HTTP,
```
martinthomson, line 106: How about?

> This document describes the overall architecture of HTTP, establishes common terminology, and defines aspects of the protocol that are shared by all versions. In this definition are core protocol elements, extensibility mechanisms, and the "http" and "https" Uniform Resource Identifier (URI) schemes.

martinthomson (commented):
I always find possessive for "HTTP" a little hard to handle.

reschke (commented):
I like Martin's suggestion.

royfielding (approved):

reschke (approved):

PR #635, opened 2021-01-04
More payload -> payload data

mnot:
I don't like this terminology, it's too easily confused and isn't distinguished from 'representation' and 'message' sufficiently.


royfielding (changes requested):
Most of these changes are wrong. None are necessary. The payload and payload data are not separate conceptual entities, so the only time we need to distinguish the two literally is for a HEAD response. I'll go through the individual changes later.

royfielding:
see also #654

royfielding:
Changed to use content instead.

PR #638, opened 2021-01-04
Cache can't be used when acting as a tunnel

mnot:
... full stop.


reschke (approved):

mnot:
Also 500f337 (committed instead of PR'd, sorry; if we don't do this that will need to be backed out)

royfielding (approved):

PR #653, opened 2021-01-06
provide a definition for partial PUT

### LINKED ISSUE #618: can Content-Range appear in a request?

royfielding:
Fixes #618 the other way

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -4770,16 +4770,9 @@ Content-Encoding: gzip
    between the related resources.
 </t>
 <t>
-   An origin server that allows PUT on a given target resource &MUST; send a
-   <x:ref>400 (Bad Request)</x:ref> response to a PUT request that contains a
-   <x:ref>Content-Range</x:ref> header field (<xref target="field.content-range"/>), since
-   the payload is likely to be partial content that has been mistakenly PUT as
-   a full representation.
-   Partial content updates are possible by targeting a separately identified
-   resource with state that overlaps a portion of the larger resource, or by
-   using a different method that has been specifically defined for partial
-   updates (for example, the PATCH method defined in
-   <xref target="RFC5789"/>).
+   Some origin servers support use of the <x:ref>Content-Range</x:ref> header
+   field (<xref target="field.content-range"/>) as a request modifier to
+   perform a partial PUT, as defined in <xref target="partial-PUT"/>.
 </t>
 <t>
    Responses to the PUT method are not cacheable. If a successful PUT request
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7868,6 +7862,66 @@ Content-Range: bytes 734-1233/1234
...
+   If an origin server supports <x:ref>Content-Range</x:ref> on PUT but the
+   range given is inconsistent with the target resource's current
+   representation and cannot be interpreted as a complete replacement,
+   the origin server &SHOULD; respond with either the
+   <x:ref>412 (Precondition Failed)</x:ref> status code (if an applicable
+   precondition is false) or the <x:ref>409 (Conflict)</x:ref> status code
+   (if not).
+</t>
+<t>
+   Unfortunately, this partial PUT feature is not backards compatible with the
```
reschke, line 7911:
```suggestion
   Unfortunately, this partial PUT feature is not backwards compatible with the
```

reschke (commented):
Maybe I'm misreading this, but the details do not seem to allow what would be the natural seconf request in a sequence of partial PUTs for a resumable upload: *appending* to an existing representation.

royfielding:
> Maybe I'm misreading this, but the details do not seem to allow what would be the natural second request in a sequence of partial PUTs for a resumable upload: _appending_ to an existing representation.

Good point, though I didn't think appending was an option. I have reduced the description to just the basics.

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7868,6 +7863,45 @@ Content-Range: bytes 734-1233/1234
...
+   target resource that does not support partial PUT requests.
+</t>
+<t>
+   If an origin server supports <x:ref>Content-Range</x:ref> on PUT but the
+   range provided is inconsistent or invalid for that resource (as defined by
+   the origin server), the origin server &SHOULD; respond with either the
+   <x:ref>412 (Precondition Failed)</x:ref> status code (if an applicable
+   precondition is false) or the <x:ref>409 (Conflict)</x:ref> status code
+   (if not).
+</t>
```
royfielding, line 7890: Remove this bit as well because it assumes too much.
```suggestion
<t>
   If an origin server supports <x:ref>Content-Range</x:ref> on PUT but the
   range provided is inconsistent or invalid for that resource (as defined by
   the origin server), the origin server &SHOULD; respond with either the
   <x:ref>412 (Precondition Failed)</x:ref> status code (if an applicable
   precondition is false) or the <x:ref>409 (Conflict)</x:ref> status code
   (if not).
</t>
```

royfielding:
This is now effectively the same as what @mnot suggested but as a separate section referred to by both PUT and Content-Range.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7868,6 +7863,37 @@ Content-Range: bytes 734-1233/1234
 </ul>
 </section>
 
+<section title="Partial PUT" anchor="partial-PUT">
+  <x:anchor-alias value="partial PUT"/>
+  <iref primary="false" item="Fields" subitem="Content-Range" x:for-anchor=""/><iref primary="false" item="Header Fields" subitem="Content-Range" x:for-anchor=""/><iref primary="false" item="Content-Range header field" x:for-anchor=""/>
```
reschke, line 7868:
```suggestion
  <iref primary="false" item="Fields" subitem="Content-Range"/><iref primary="false" item="Header Fields" subitem="Content-Range"/><iref primary="false" item="Content-Range header field"/>
```

reschke (approved):

mnot (approved):

PR #666, opened 2021-01-08
Clarify how multiple validators are handled

### LINKED ISSUE #659: Freshening responses with more than one validator
[4.3.4 Freshening Stored Responses upon Validation](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#freshening.responses) is silent about what to do when there's more than one validator in the message. Freshening responses with HEAD (just below) specifies that any validators must match.

(use of 'match' there is a bit fuzzy, should probably link to where it's defined)

mnot:
Fixes #659

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1099,7 +1099,7 @@
 <ul>
    <li>
      If the new response contains one or more <x:dfn>strong validators</x:dfn> (see
-     <xref target="weak.and.strong.validators"/>), then those strong validators
+     <xref target="weak.and.strong.validators"/>), then each of those strong validators
      identify the selected representation for update. All the stored
      responses with one of those same strong validators are identified for update. If
      none of the stored responses contain at least one of the same strong validators, then the
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1108,7 +1108,7 @@
    <li>
      If the new response contains no strong validators but does contain
      one or more <x:dfn>weak validators</x:dfn>, and those
-     validator correspond to one of the cache's stored responses, then the most
+     validators correspond to one of the cache's stored responses, then the most
      recent of those matching stored responses is identified for update.
    </li>
    <li>
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1101,7 +1101,7 @@
      If the new response contains one or more <x:dfn>strong validators</x:dfn> (see
      <xref target="weak.and.strong.validators"/>), then those strong validators
      identify the selected representation for update. All the stored
-     responses with the same strong validators are identified for update. If
+     responses with one of those same strong validators are identified for update. If
      none of the stored responses contain at least one of the same strong validators, then the
      cache &MUST-NOT; use the new response to update any stored responses.
    </li>
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1098,16 +1098,17 @@
 </t>
 <ul>
    <li>
-     If the new response contains a <x:dfn>strong validator</x:dfn> (see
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1098,16 +1098,17 @@
 </t>
 <ul>
    <li>
-     If the new response contains a <x:dfn>strong validator</x:dfn> (see
-     <xref target="weak.and.strong.validators"/>), then that strong validator
+     If the new response contains one or more <x:dfn>strong validators</x:dfn> (see
+     <xref target="weak.and.strong.validators"/>), then those strong validators
      identifies the selected representation for update. All the stored
```
reschke, line 1103:
```suggestion
     identify the selected representation for update. All the stored
```

reschke, line 1103:
```suggestion
     identify the selected representation for update. All the stored
```

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1098,16 +1098,17 @@
...
+     responses with the same strong validators are identified for update. If
+     none of the stored responses contain the same strong validators, then the
      cache &MUST-NOT; use the new response to update any stored responses.
    </li>
    <li>
-     If the new response contains a <x:dfn>weak validator</x:dfn> and that
-     validator corresponds to one of the cache's stored responses, then the most
+     If the new response contains no strong validators but does contain
+     one or more <x:dfn>weak validators</x:dfn>, and those
+     validator correspond to one of the cache's stored responses, then the most
```
reschke, line 1111:
```suggestion
     validators correspond to one of the cache's stored responses, then the most
```

reschke:
There's still a difference when the response mixes strong and weak validators, right? If both types are present, the weak ones would get ignored. Is that intentional?

royfielding:
> There's still a difference when the response mixes strong and weak validators, right? If both types are present, the weak ones would get ignored. Is that intentional?

Yes, that is intentional.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1098,16 +1098,17 @@
...
    <li>
-     If the new response contains a <x:dfn>strong validator</x:dfn> (see
-     <xref target="weak.and.strong.validators"/>), then that strong validator
-     identifies the selected representation for update. All the stored
-     responses with the same strong validator are identified for update. If
-     none of the stored responses contain the same strong validator, then the
+     If the new response contains one or more <x:dfn>strong validators</x:dfn> (see
+     <xref target="weak.and.strong.validators"/>), then those strong validators
+     identify the selected representation for update. All the stored
+     responses with the same strong validators are identified for update. If
```
royfielding, line 1104:
```suggestion
     responses with one of those same strong validators are identified for update. If
```

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1098,16 +1098,17 @@
 </t>
 <ul>
    <li>
-     If the new response contains a <x:dfn>strong validator</x:dfn> (see
-     <xref target="weak.and.strong.validators"/>), then that strong validator
-     identifies the selected representation for update. All the stored
-     responses with the same strong validator are identified for update. If
-     none of the stored responses contain the same strong validator, then the
+     If the new response contains one or more <x:dfn>strong validators</x:dfn> (see
+     <xref target="weak.and.strong.validators"/>), then those strong validators
```
royfielding, line 1102:
```suggestion
     <xref target="weak.and.strong.validators"/>), then each of those strong validators
```

royfielding (changes requested):
This is a little hard to understand because the original text assumes we are talking about a single strong etag and a single weak LM time.

In theory, we should either require that one match or require that they all match. In practice, the entity-tag is definitive and all the other validators are simple string match if and only if no entity-tag is present.

I don't know if we want to monkey patch this, or just throw out the existing text and point to Semantics.

royfielding (approved):

reschke (approved):

PR #667, opened 2021-01-08
Even HTTP/1.0 caches now support Age and CC

### LINKED ISSUE #660: Very Old Warnings
> Note: Some HTTP/1.0 caches might not implement Cache-Control.

Is this warning still relevant? If we want to keep it, I think we need to qualify it; e.g., "A few old HTTP/1.0 caches that are unlikely to be used on the open Internet..."

Similar text in cc: no-cache.

mnot:
... it's just deployments of very very old caches that would be of concern here.

Given how unlikely this is in 2020, how broken they would be for other reasons, and the
prevalence of HTTPS, it seems safe to modify these statements.

Fixes #660.

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -2515,6 +2515,7 @@
 <section title="Since draft-ietf-httpbis-cache-13" anchor="changes.since.13">
 <ul x:when-empty="None yet.">
   <li>In <xref target="cache-response-directive.must-revalidate"/>, clarify requirements around generating an error response (<eref target="https://github.com/httpwg/http-core/issues/608"/>)</li>
+  <li>In <xref target="age.calculations"/>, <xref target="field.cache-control"/>, and <xref target="cache-response-directive.no-cache"/>, remove notes about very old HTTP/1.0 behaviours (<eref target="https://github.com/httpwg/http-core/issues/660"/>)</li>
 </ul>
 </section>
 </section>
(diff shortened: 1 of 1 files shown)


reschke (commented):
...add change tracking...

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1237,10 +1234,9 @@
 </t>
 <t>
    The presence of an Age header field implies that the response was not
-   generated or validated by the origin server for this request. However,
-   lack of an Age header field does not imply the origin was contacted, since
-   the response might have been received from an HTTP/1.0 cache that does not
-   implement Age.
+   generated or validated by the origin server for this request. Note that
+   in rare circumstances, the response might have been received from an HTTP/1.0
+   cache that does not implement Age.
```
royfielding, line 1239:
```suggestion
   generated or validated by the origin server for this request.
```

royfielding: We need to either keep the original or remove the sentence entirely -- the entire point of that sentence is to say that lack of Age doesn't imply the reverse (that the response was generated or validated by the origin server for this request). I'd be fine with deleting or keeping the original sentence, or just removing the explanatory suffix ", since the response might ..."

royfielding (changes requested):

draft-ietf-httpbis-cache-latest.xml
royfielding: I mean, the original sentence:

> However, lack of an Age header field does not imply the origin was contacted, since the response might have been received from an HTTP/1.0 cache that does not implement Age.

can be deleted or replaced with

> However, lack of an Age header field does not imply the origin was contacted, since the response might have been received from a non-conforming cache.

or simply

> However, lack of an Age header field does not imply the origin was contacted.

mnot: See latest comment; I went with the last suggestion. Thanks.

royfielding (approved):

reschke (approved):

PR #670, opened 2021-01-11
Specify IMS and IUS like INM and If-Match

### LINKED ISSUE #665: Duplicative specification of precondition handling
[Caching 4.3.2](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#validation.received) re-specifies precondition processing from [Semantics 13](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#preconditions). 

E.g.,

> A request containing an If-None-Match header field (Section 13.1.2 of [Semantics]) indicates that the client wants to validate one or more of its own stored responses in comparison to whichever stored response is selected by the cache. If the field value is "*", or if the field value is a list of entity-tags and at least one of them matches the entity-tag of the selected stored response, a cache recipient should generate a 304 (Not Modified) response (using the metadata of the selected stored response) instead of sending that stored response.

> If an If-None-Match header field is not present, a request containing an If-Modified-Since header field (Section 13.1.3 of [Semantics]) indicates that the client wants to validate one or more of its own stored responses by modification date. A cache recipient should generate a 304 (Not Modified) response (using the metadata of the selected stored response) if one of the following cases is true:

[...]

Do we want to try to fold these into Semantics so that it isn't duplicated? That would require some small adjustments to Semantics AFAICT (e.g., to make it clear what the selected representation is in each case, etc.).

mnot:
Fixes #665

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -13376,7 +13376,7 @@ Content-Type: text/plain
   <li>In <xref target="length-requirements"/>, rephrase the actual recipient parsing requirements (<eref target="https://github.com/httpwg/http-core/issues/634"/>)</li>
   <li>In <xref target="considerations.for.new.methods"/>, mention request target forms in considerations for new methods (<eref target="https://github.com/httpwg/http-core/issues/636"/>)</li>
   <li>Changed to using "content" instead of "payload" or "payload data" to avoid confusion with the payload of version-specific messaging frames (<eref target="https://github.com/httpwg/http-core/issues/654"/>)</li>
-  <li>In <xref target="field.if-modified-since"/> and <xref target="field.if-unmodified-since"/>, specify evaluation in a way similar to other conditional header fields (<eref target="https://github.com/httpwg/http-core/issues/665"/>)</li>
+  <li>In <xref target="field.if-modified-since"/>, <xref target="field.if-unmodified-since"/>, and <xref target="field.if-range"/>, specify evaluation in a way similar to other conditional header fields (<eref target="https://github.com/httpwg/http-core/issues/665"/>)</li>
   <li>In <xref target="field.date"/>, specify that recipients can replace an invalid Date header field value with the time received (<eref target="https://github.com/httpwg/http-core/issues/669"/>)</li>
 </ul>
 </section>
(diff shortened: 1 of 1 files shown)


reschke (approved):
I don't think this change is needed, but if you prefer it this way...

royfielding (changes requested):
I prefer the original, particularly given that these are requirements.

mnot:
They're still requirements, and without the changes, we don't explicitly say how to evaluate the conditionals; elsewhere (e.g., [precedence](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#precedence) we refer to them as evaluating to either true or false, hence these changes.

reschke:
I see. Maybe a less intrusive change would be to say:

"... The origin server SHOULD NOT perform the requested method if **the condition evaluates to false:** the selected representation's last modification date is earlier than or equal to the date provided in the field value; instead, the origin server SHOULD generate a 304 (Not Modified) response, including only those metadata that are useful for identifying or updating a previously cached response."

royfielding (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7172,14 +7188,25 @@ Expect: 100-continue
...
 </t>
 <t>
-   If the selected representation has a last modification date,
-   the origin server &MUST-NOT; perform the requested method if that date is
-   more recent than the date provided in the field value.
+   To evaluate a received If-Unmodified-Since header field:
+</t>
+<ol>
+   <li>
+      If the selected representation's last modification date is earlier or
```
reschke, line 7200:
```suggestion
      If the selected representation's last modification date is earlier than or
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7106,18 +7107,32 @@ Expect: 100-continue
...
-   response.
+   evaluate the condition as per <xref target="evaluation"/> prior to
+   performing the method.
+</t>
+<t>
+   To evaluate a received If-Modified-Since header field:
+</t>
+<ol>
+   <li>
+      If the selected representation's last modification date is earlier or
```
reschke, line 7123:
```suggestion
      If the selected representation's last modification date is earlier than or
```

reschke (approved):
(but see comments)

PR #671, opened 2021-01-11
Invalid Date handling

### LINKED ISSUE #669: When Date doesn't parse
[Date](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#field.date) specifies what to do when it isn't present (use the received time), but not what to do when it doesn't parse.

The intuitive thing to do is align, and say that the received time should be used in this situation as well -- but probably not require implementations to replace it if forwarded.

I might write some tests...

mnot:
Fixes #669

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1027,26 +1027,14 @@
    containing an <x:ref>If-Modified-Since</x:ref> header field
    (<xref target="field.if-modified-since"/>) indicates that the client wants to validate
    one or more of its own stored responses by modification date.
-   A cache recipient &SHOULD; generate a <x:ref>304 (Not Modified)</x:ref>
-   response (using the metadata of the selected stored response) if one of the
-   following cases is true:
 </t>
-<ol>
-  <li>
-    the selected stored response has a <x:ref>Last-Modified</x:ref>
-    field value that is earlier than or equal to the conditional timestamp;
-  </li>
-  <li>
-    no <x:ref>Last-Modified</x:ref> header field is present in the selected stored
-    response, but it has a <x:ref>Date</x:ref> field value that is earlier than
-    or equal to the conditional timestamp; or,
-  </li>
-  <li>
-    neither <x:ref>Last-Modified</x:ref> nor <x:ref>Date</x:ref> is present
-    in the selected stored response, but the cache recorded it as having been
-    received at a time earlier than or equal to the conditional timestamp.
-  </li>
-</ol>
+<t>
+   If a request contains the <x:ref>If-Modified-Since</x:ref> header field and
+   the <x:ref>Last-Modified</x:ref> header field is not present in a selected
+   stored response, a cache &SHOULD; use the <x:ref>Date</x:ref> field value
+   to evaluate the conditional; if it is not present, the time that the
+   response was received &SHOULD; be used.
+</t>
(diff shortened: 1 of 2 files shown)


reschke (commented):
It's a bit unfortunate that this change removes text that hasn't anything to do with Date handling. Maybe separate the changes, or rephrase the issue subject?

mnot:
Those changes are editorial.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1027,26 +1022,14 @@
...
-    in the selected stored response, but the cache recorded it as having been
-    received at a time earlier than or equal to the conditional timestamp.
-  </li>
-</ol>
+<t>
+   If a request contains the <x:ref>If-Modified-Since</x:ref> header field and
+   the <x:ref>Last-Modified</x:ref> header field is not present in a selected
+   stored response, a cache &SHOULD; use the <x:ref>Date</x:ref> field value
+   to evaluate the conditional; if it is not present, the time that the
+   response was received &SHOULD; be used.
```
royfielding, line 1031:
```suggestion
   If a request contains an <x:ref>If-Modified-Since</x:ref> header field and
   the <x:ref>Last-Modified</x:ref> header field is not present in a selected
   stored response, a cache &SHOULD; use the stored response's
   <x:ref>Date</x:ref> field value (or, if no Date field is present, the time
   that the stored response was received) to evaluate the conditional.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5510,6 +5510,11 @@ Expect: 100-continue
    corresponding Date header field to the message's header section if it is
    cached or forwarded downstream.
 </t>
+<t>
+   A recipient with a clock that receives a response with a Date header field
+   that is invalid &MAY; record the time it was received and replace the Date header
+   field in the message's header section.
```
royfielding, line 5516:
```suggestion
   A recipient with a clock that receives a response with an invalid Date
   header field value &MAY; replace that value with the time that
   response was received.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -13315,6 +13320,7 @@ Content-Type: text/plain
   <li>In <xref target="length-requirements"/>, rephrase the actual recipient parsing requirements (<eref target="https://github.com/httpwg/http-core/issues/634"/>)</li>
   <li>In <xref target="considerations.for.new.methods"/>, mention request target forms in considerations for new methods (<eref target="https://github.com/httpwg/http-core/issues/636"/>)</li>
   <li>Changed to using "content" instead of "payload" or "payload data" to avoid confusion with the payload of version-specific messaging frames (<eref target="https://github.com/httpwg/http-core/issues/654"/>)</li>
+  <li>In <xref target="field.date"/>, clarify how invalid Date headers fields should be handled (<eref target="https://github.com/httpwg/http-core/issues/669"/>)</li>
```
royfielding, line 13323:
```suggestion
  <li>In <xref target="field.date"/>, specify that recipients can replace an invalid Date header field value with the time received (<eref target="https://github.com/httpwg/http-core/issues/669"/>)</li>
```

royfielding (changes requested):

royfielding (approved):

reschke (approved):

PR #695, opened 2021-01-28
Try to better define "strong" encryption

martinthomson:
In the definition of the https scheme, the phrase "strong encryption" is used.
This implies some sort of judgment.  I think that the key is that both
client and server need to agree on the adequacy of the protections.  Noting
that this might be contextual is somewhat obvious at that point, but I thought
it might be worth pointing out that what is generally acceptable might change
over time.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -1106,7 +1106,7 @@ Content-Type: text/plain
    A client &MUST; ensure that its HTTP requests for an "https" resource are
    secured, prior to being communicated, and that it only accepts secured
    responses to those requests. Note that the definition of what cryptographic
-   mechanisms are acceptable to client and server are negotiated in TLS and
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1105,7 +1105,9 @@ Content-Type: text/plain
 <t>
    A client &MUST; ensure that its HTTP requests for an "https" resource are
    secured, prior to being communicated, and that it only accepts secured
-   responses to those requests.
+   responses to those requests. Note that the definition of what cryptographic
+   mechanisms are acceptable to client and server are negotiated in TLS and
+   can change over time.
```
royfielding, line 1110:
```suggestion
   responses to those requests. Note that the definition of what cryptographic
   mechanisms are acceptable to client and server are usually negotiated and
   can change over time.
```

royfielding (approved):

PR #701, opened 2021-01-29
TE: trailers is in a request

martinthomson:
I know that this is pretty obvious and implied, but maybe it's worth saying.


reschke:
maybe just s/TE header field/TE request header field/?

Issue #700, opened 2021-01-28
Messages aren't always self-descriptive


martinthomson:
Section 6 claims that [messages are self-descriptive](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.6.p.5).  It uses the word "intended", which is an intention that is broadly achieved, with only one exception that I'm aware of at the syntactic level.

Having recently implemented message/http, I discovered that some response messages cannot be processed without knowledge of the corresponding request.  This is pretty annoying and might be worth noting as a possible exception here.  As noted in [Section 6.3 of -messaging](https://httpwg.org/http-core/draft-ietf-httpbis-messaging-latest.html#message.body.length) the response to a HEAD request might include Content-Length, but the content of the response is always empty.

At a semantic level, I don't think that any response can be considered without knowledge of the request, so I don't think that self-descriptiveness can possibly apply there.  Besides, this section is clearly intended to provide an abstract definition of the syntax and not semantics.

royfielding:
See also #706

PR #706, opened 2021-01-29
How a HEAD response terminates is version-specific

martinthomson:
Though I think that this text is correct, an HTTP/3 response only ends when
the stream is closed, not at the end of the header section.  I think that it's
not necessary to say this bit anyway.


royfielding (commented):
We need something else here, which you alluded to in a different issue. The client MUST understand that the response will not contain content, since otherwise it looks like the beginning of a GET response.

royfielding (approved):

royfielding:
The related parts were done in #700 so this bit can be applied now as editorial.

PR #707, opened 2021-01-29
Origin servers might support CONNECT

martinthomson:
The original text here sort of implied that it was enough for an origin server
to just send 2xx.  That assumed that the request was for the origin server
itself and elided a bunch of detail.  This is shorter and clearer, I think.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -4865,9 +4865,8 @@ Content-Encoding: gzip
 </t>
 <t>
    CONNECT is intended only for use in requests to a proxy.
-   An origin server that receives a CONNECT request for itself &MAY;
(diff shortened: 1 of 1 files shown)


royfielding (commented):
This is actually a change to the protocol in order to support the use of CONNECT as a means of enabling websockets via a gateway, right? We should be specific and reference the RFC as an example.

martinthomson:
I didn't intend anything here other than a clarification of what was written.  I'm mostly guessing what was intended as I don't know what the original text truly intended to say.

(I think that it was always possible to use CONNECT with a target of the origin server to do ...stuff.  But I didn't want to highlight that here; it's a little bit of an odd thing to decide to do and so I wouldn't want to encourage it.)

royfielding (approved):

PR #717, opened 2021-01-31
Clarify status code range (fixes #684)

### LINKED ISSUE #684: status-code parsing
I looked into status-code a tiny bit and noticed that browsers do not have fully conforming parsers (quelle surprise, I know). I only looked at H/1.

This might not warrant any changes, but if people are writing tests this might be good to take into account, especially for newer versions of HTTP.

https://github.com/whatwg/fetch/issues/1142

reschke:


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8108,6 +8109,9 @@ Content-Range: exampleunit 11.2-14.3/25
     valid request
    </li>
 </ul>
+<t>
+   Values outside the range 100..599 are invalid.
```
royfielding, line 8113:
```suggestion
   Values outside the range 100..599 are invalid. Implementations often use three-digit
   integer values outside of that range (e.g., 600..999) for internal communication of
   non-HTTP status (e.g., library errors).
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -13383,6 +13387,7 @@ Content-Type: text/plain
 
 <section title="Since draft-ietf-httpbis-semantics-14" anchor="changes.since.14">
 <ul x:when-empty="None yet.">
+  <li>In <xref target="status.codes"/>, clarify that values outside the range 100..599 are invalid (<eref target="https://github.com/httpwg/http-core/issues/684"/>)</li>
```
royfielding, line 13390:
```suggestion
  <li>In <xref target="status.codes"/>, clarify that status code values outside the range 100..599 are invalid (<eref target="https://github.com/httpwg/http-core/issues/684"/>)</li>
```

royfielding (changes requested):
I think we need to say why they are invalid.

mnot (approved):

mnot:
Do we want to give any guidance about what an implementation should do when receiving an invalid code? I ask because if we don't, implementations are likely to behave in different ways.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8064,8 +8064,9 @@ Content-Range: exampleunit 11.2-14.3/25
 <section title="Status Codes" anchor="status.codes">
   <iref item="Status Code"/>
 <t>
-   The (response) status code is a three-digit integer code giving the result of the
-   attempt to understand and satisfy the request.
+   The (response) status code is a three-digit integer code between 100 and
+   599 (inclusive), giving the result of the attempt to understand and satisfy
+   the request.
```
royfielding, line 8069: On second read, this is getting clunky and burying the definition. I'd prefer:
```suggestion
   The status code of a response is a three-digit integer code that describes the
   result of the request and the semantics of the response, including whether or
   not the request was successful and what content is enclosed (if any).
   All valid status codes for HTTP are within the range of 100 to 599, inclusive.
```

royfielding (changes requested):

royfielding:
> Do we want to give any guidance about what an implementation should do when receiving an invalid code? I ask because if we don't, implementations are likely to behave in different ways.

I think that is application-dependent, since behavior will depend on how the user interacts with the response and whether the application is directly using HTTP or using a library that happens to use HTTP.

Perhaps

> A client that receives an invalid status code SHOULD use the response like a 500 status code.

mnot:
I think adding that would be helpful -- it gives people room for other behaviours, but suggests a sensible default handling.

annevk:
I suspect

> A client that receives an invalid status code SHOULD use the response like a 500 status code.

might be open to misinterpretation as I could see people not considering 299 to be valid, for instance. (It's in the valid range, but I suppose it needs to be registered to be truly valid?)

reschke:
> could see people not considering 299 to be valid, for instance

We really need to say that 100...599 are valid, not that all valid codes are in that range.

annevk:
> A valid status code is a status code in the range 100 to 599, inclusive.

Would make that clear, I think.

reschke:
Not really. It only says that valid codes are in 100..599, not that *all* values in that range are valid.

annevk:
I read it as defining valid status code being a status code in that range.

reschke:
You are right. I misread it.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8148,6 +8150,13 @@ Content-Range: exampleunit 11.2-14.3/25
     valid request
    </li>
 </ul>
+<t>
+   Values outside the range 100..599 are invalid. Implementations often use three-digit
+   integer values outside of that range (i.e., 600..999) for internal communication of
+   non-HTTP status (e.g., library errors). A client that receives a response with
+   an invalid status code &SHOULD; process it as if it had a <x:ref>5xx (Server Error)</x:ref>
```
royfielding, line 8157:
```suggestion
   an invalid status code &SHOULD; process the response as if it had a <x:ref>5xx (Server Error)</x:ref>
```

royfielding (approved):

PR #718, opened 2021-01-31
Values of newly defined fields should only use VCHAR/SP/HTAB (fixes #696)

### LINKED ISSUE #696: US-ASCII
In [Section 5.5](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.5.5.p.3):

> Newly defined fields SHOULD limit their values to US-ASCII octets. 

This is probably overly broad considering that the ABNF says `VCHAR / SP / HTAB` (I question the wisdom of including HTAB, but now is not the time to worry about that).  Could this be changed to "printable or visible US-ASCII octets, plus SP and HTAB"?  

(Incidentally, why doesn't the `field-content` rule use the ABNF `WSP` rule?)

reschke:


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1601,7 +1601,8 @@ Content-Type: text/plain
    through use of <xref target="RFC2047"/> encoding.
    In practice, most HTTP field values use only a subset of the
    US-ASCII charset <xref target="USASCII"/>. Newly defined
-   fields &SHOULD; limit their values to US-ASCII octets.
+   fields &SHOULD; limit their values to visible US-ASCII octets
+   (VCHAR), plus SP and HTAB.
```
royfielding, line 1605:
```suggestion
   fields &SHOULD; limit their values to visible US-ASCII octets
   (VCHAR), SP, and HTAB.
```

royfielding (commented):
oxford comma that

royfielding (approved):

PR #720, opened 2021-01-31
Clarify that selection between multiple acceptable codings is only relevant when some of the acceptable ones are redundant (fixes #714)

### LINKED ISSUE #714: Multiple content codings are possible
However, [this text](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.12.5.3.p.9) from item 4 implies otherwise.

> If multiple content-codings are acceptable, then the acceptable content-coding with the highest non-zero qvalue is preferred.

I think that this only applies if the server wishes to choose between codings, as opposed to apply them all.  Compression codings are usually mutually exclusive, but other codings are not necessarily.  Splitting this point into a paragraph might be the way to resolve this:

> A message can include multiple content codings [I think no hyphen is correct], but some codings, such as those that compress content, can be redundant.  When selecting between multiple redundant content codings, the acceptable content coding with the highest non-zero qvalue is preferred.

reschke:
…levant when some of the acceptable ones are redundant (fixes #714)

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -13389,7 +13389,7 @@ Content-Type: text/plain
 <section title="Since draft-ietf-httpbis-semantics-14" anchor="changes.since.14">
 <ul x:when-empty="None yet.">
   <li>Avoid the term "whitelist" (<eref target="https://github.com/httpwg/http-core/issues/688"/>)</li>
-  <li>In <xref target="field.accept-encoding"/>, clarify that selection <em>between</em> multiple acceptable codings is only relevant when some of the acceptable ones are redundant (<eref target="https://github.com/httpwg/http-core/issues/714"/>)</li>
+  <li>In <xref target="field.accept-encoding"/>, clarify that selection <em>between</em> multiple acceptable codings is only relevant when they have the same purpose (<eref target="https://github.com/httpwg/http-core/issues/714"/>)</li>
 </ul>
 </section>
 </section>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6623,9 +6623,13 @@ Expect: 100-continue
...
        it is accompanied by a qvalue of 0. (As defined in <xref target="quality.values"/>, a
        qvalue of 0 means "not acceptable".)</li>
-   <li>If multiple content-codings are acceptable, then the acceptable
-       content-coding with the highest non-zero qvalue is preferred.</li>
 </ol>
+<t>
+   A message can include multiple content codings, but some codings, such as
+   those that compress content, can be redundant. When selecting between
+   multiple redundant content codings, the acceptable content coding with
+   the highest non-zero qvalue is preferred.
```
royfielding, line 6631:
```suggestion
   A representation could be encoded with multiple content codings. However, most
   content codings are alternative ways to accomplish the same purpose
   (e.g., data compression). When selecting between multiple content codings that
   have the same purpose, the acceptable content coding with the highest
   non-zero qvalue is preferred.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -13384,6 +13388,7 @@ Content-Type: text/plain
 <section title="Since draft-ietf-httpbis-semantics-14" anchor="changes.since.14">
 <ul x:when-empty="None yet.">
   <li>Avoid the term "whitelist" (<eref target="https://github.com/httpwg/http-core/issues/688"/>)</li>
+  <li>In <xref target="field.accept-encoding"/>, clarify that selection <em>between</em> multiple acceptable codings is only relevant when some of the acceptable ones are redundant (<eref target="https://github.com/httpwg/http-core/issues/714"/>)</li>
```
royfielding, line 13391:
```suggestion
  <li>In <xref target="field.accept-encoding"/>, clarify that selection <em>between</em> multiple acceptable codings is only relevant when they have the same purpose (<eref target="https://github.com/httpwg/http-core/issues/714"/>)</li>
```

royfielding (changes requested):

PR #724, opened 2021-02-01
Network congestion or server load

martinthomson:
The section was talking about server load as well, but it doesn't mention that
being a cost of closing connections.


royfielding (changes requested):
This is specifically talking about dropping idle connections to reduce current server load, so that addition doesn't make sense on its own. It might make sense if it also explained why/when dropping idle connections can result in increased load (like when they are not really idle). I'm not too keen on the original paragraph.

martinthomson:
I thought that the assumption was OK-ish: the expectation that clients will retry.  That said, I agree that this entire thing could be better.  I was looking for a minimal fix.

royfielding (approved):

Issue #725, opened 2021-02-01
TLS does not require clean close for resumption


martinthomson:
A clean close of the original connection was a requirement for use of resumption in previous versions of TLS, but TLS 1.3 does not have that same requirement.

I just checked, which was tedious.  Then I found this in RFC 5246:
> Note that as of TLS 1.1, failure to properly close a connection no longer requires that a session not be resumed.

This calls into question the entirety of [Section 9.8](https://httpwg.org/http-core/draft-ietf-httpbis-messaging-latest.html#tls.connection.closure).  I think that the only requirement regarding the TLS close_notify alert needs to apply only when framing responses that don't have `Transfer-Encoding: chunked` or `Content-Length`.  That might make this section much simpler.

Also, I note that this section uses the term "reuse", which is not consistent with "resume" or "resumption" in TLS.  If there is any need for mention of TLS resumption, then the TLS term should be used.

royfielding:
Fixed by #734

PR #734, opened 2021-02-02
Fixes for TLS connection closure

martinthomson:
1. A response is incomplete if the TLS connection close was incomplete.

2. Remove text about TLS reuse/resumption, which was very inconsistent.

On the first, I removed a case of "partial closure" rather than "incomplete
closure".  I also linked the incomplete message handling section with the TLS
connection closure section.

### DIFF
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -1485,7 +1485,9 @@ https://www.example.org
    transfer coding nor Content-Length is terminated by closure of the
    connection and, thus, is considered complete regardless of the number of
    message body octets received, provided that the header section was received
-   intact.
+   intact.  When using TLS, an incomplete close cannot be used to terminate a
+   response this way; such a response is incomplete (see
+   <xref target="tls.connection.closure"/>).
 </t>
 </section>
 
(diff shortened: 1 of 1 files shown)


martinthomson:
Apparently I'm not allow to link to #725 after the fact.  Sorry for not having done so in the opening comment.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1836,43 +1838,41 @@ https://www.example.org
...
+   An incomplete close does not call into question the security of the data
+   already received, but it could indicate that subsequent data might have been
+   truncated. As TLS is not directly aware of HTTP message framing, it is
+   necessary to examine the HTTP data itself to determine whether messages were
+   complete. Handing of incomplete messages is defined in
+   <xref target="incomplete.messages"/>.
 </t>
 <t>
-   When encountering a premature close, a client &SHOULD; treat as completed
+   When encountering a incomplete close, a client &SHOULD; treat as completed
```
royfielding, line 1855:
```suggestion
   When encountering an incomplete close, a client &SHOULD; treat as completed
```

royfielding:
Fixes #725 

This is okay, but the paragraph before your first change should be rewritten instead. I can fix that later.

PR #747, opened 2021-02-06
Move statement about relation between Content-Length and Transfer-Encoding into Messaging (fixes #704)

### LINKED ISSUE #704: Content-Length and Transfer-Encoding coexistence
Does [the following statement](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.8.6.p.5) better belong in -messaging?

> A sender MUST NOT send a Content-Length header field in any message that contains a Transfer-Encoding header field.

reschke:


royfielding (approved):
But I don't know what this accomplishes given the paragraph after this one also refers to Transfer-Encoding.

reschke:
Good point.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3584,10 +3584,6 @@ Upgrade: websocket
 <artwork type="example">
   Content-Length: 3495
 </artwork>
-<t>
-   A sender &MUST-NOT; send a Content-Length header field in any message that
-   contains a <x:ref>Transfer-Encoding</x:ref> header field.
-</t>
 <t>
    A user agent &SHOULD; send a Content-Length in a request message when no
    <x:ref>Transfer-Encoding</x:ref> is sent and the request method defines
```
martinthomson, line 3589: The fix for this might be as follows:

>  A user agent `&SHOULD;` send a Content-Length in a request message when the request method defines a meaning for enclosed content.  However, in HTTP/1.1 Content-Length cannot be included when `<x:ref>`Transfer-Encoding`</x:ref>` is also present.

martinthomson (approved):
Thanks.  Fairly simple fix here.  I've suggested a tweak for Roy's concern (which I can't make a suggestion before because of GitHub limitations.

mnot (approved):
This makes sense on its own (with Martin's adjustment). Still not thrilled about having transfer-codings in semantics.

reschke:
I guess all the options are ugly.

Should we leave the text wehre ut was and just apply the proposed change in https://github.com/httpwg/http-core/pull/747#discussion_r571732892 ?

PR #756, opened 2021-02-10
Define 'cache key'

### LINKED ISSUE #728: Define "cache key"
This might be obvious, but the -caching draft never concretely says that a cache key is the information that a cache uses to select a single cached response.  It just [says](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#rfc.section.2.p.2): "The cache key is comprised of, [...]"

mnot:
Fixes #728.

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -264,7 +264,7 @@
 </t>
 <iref item="cache key" />
 <t>
-   The <x:dfn>cache key</x:dfn> is the information used to select a response and
+   The <x:dfn>cache key</x:dfn> is the information a cache uses to select a response and
    is comprised of, at a minimum, the request method and target
    URI used to retrieve the stored response; the method determines under which
    circumstances that response can be used to satisfy a subsequent request. However, many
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -264,7 +264,8 @@
 </t>
 <iref item="cache key" />
 <t>
-   The <x:dfn>cache key</x:dfn> is comprised of, at a minimum, the request method and target
+   The <x:dfn>cache key</x:dfn> is the information used to select a response and
```
martinthomson, line 267:
```suggestion
   The <x:dfn>cache key</x:dfn> is the information a cache uses to select a response and
```
Maybe avoid the passive voice in this case.

martinthomson (approved):
LGTM

draft-ietf-httpbis-cache-latest.xml
mnot: But it's my friend.... 

I'm not thrilled about the repeated "cache... cache" but yes this is better.

reschke (approved):

royfielding (approved):

PR #757, opened 2021-02-10
Broaden cache poisoning

### LINKED ISSUE #730: Cache poisoning doesn't just affect shared caches
A private cache that can be manipulated so that it mistakenly creates a cache entry might not have the reach of a shared cache, but it can still be a problem.  Cache poisoning occurs when a cache can be written to by an entity other than the one that the cache would normally recognize as being authorized to write to it.

Take the websocket poisoning where a cache misread a websocket stream as HTTP queries and established cached records under arbitrary origins.  This would allow for poisoning of entries normally not controlled by the host that was using the websocket.  Even though that might have been a single user affected by the poisoning, the effect was serious enough and motivated the inclusion of masking in websocket.

mnot:
Fixes #730

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1891,11 +1891,11 @@
 
 <section title="Cache Poisoning" anchor="cache.poisoning">
 <t>
-   Various attacks might be amplified by being stored in a shared cache. Such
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Simple :)

reschke (approved):

royfielding (approved):

PR #760, opened 2021-02-10
Be more explicit about Content-Length errors

### LINKED ISSUE #748: Semantics: 8.6 Content-length: too lax about multiple values
Regarding the case of multi-valued header, we have this:

> If a message is received that has a Content-Length header field value consisting of the same decimal value as a comma-separated list (Section 5.6.1) - for example, "Content-Length: 42, 42" - indicating that duplicate Content-Length header fields have been generated or combined by an upstream message processor, then the recipient MUST either reject the message as invalid or replace the duplicated field values with a single valid Content-Length field containing that decimal value.

I'm not seeing anything there which forbids the use of "Content-length: 5,10" anymore. I'd rather emphasize the need for extremely strict parsing of Content-length and keep the duplicate case as the only acceptable exception to this rule, approximately like this:

"The Content-Length header field plays a crucial role in message delimitation in HTTP/1.1, and a different error handling between agents was shown to have important consequences in terms of security.  As such, it is extremely important that agents do not accept Content-Length header field values that do not strictly comply with the ABNF, that they take care of rejecting values that they are not able to accurately represent internally, and that multiple occurrences of the header field are always checked for, and rejected as invalid. As an exception, if the header field is received multiple times with the same value, or is received as a comma-separated list of identical values, the recipient MAY replace all of them with a single valid Content-Length field containing that decimal value instead of rejecting the message."

mnot:
Fixes #748

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -3634,11 +3634,12 @@ Upgrade: websocket
    (<xref target="attack.protocol.element.length"/>).
 </t>
 <t>
-   Because it is used for message delimitation in HTTP/1.1, the Content-Length
-   header field value can affect the security of the protocol. This is true even
-   when the current connection is not using HTTP/1.1, because HTTP is an
-   intermediated protocol, and messages might be transmitted using other versions
-   of the protocol.
+   Because Content-Length is used for message delimitation in HTTP/1.1,
+   its field value can impact how the message is parsed by downstream
+   recipients even when the immediate connection is not using HTTP/1.1.
+   If the message is forwarded by a downstream intermediary, a Content-Length
+   field value that is inconsistent with the received message framing might
+   cause a security failure due to request smuggling or response splitting.
 </t>
 <t>
    As a result, implementations &MUST-NOT; forward a message with a
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -3642,11 +3642,11 @@ Upgrade: websocket
    cause a security failure due to request smuggling or response splitting.
 </t>
 <t>
-   As a result, implementations &MUST-NOT; forward a message with a
+   As a result, a sender &MUST-NOT; forward a message with a
    Content-Length header field value that is known to be incorrect.
 </t>
 <t>
-   Likewise, implementations &MUST-NOT; forward a message with a Content-Length
+   Likewise, a sender &MUST-NOT; forward a message with a Content-Length
    header field value that does not match the ABNF above, with one exception:
    If a message is received that has a Content-Length header field value
    consisting of the same decimal value as a comma-separated list (<xref
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -3648,7 +3648,7 @@ Upgrade: websocket
 </t>
 <t>
    Likewise, implementations &MUST-NOT; forward a message with a Content-Length
-   header field value that does not match the ABNF above, with one exception.
+   header field value that does not match the ABNF above, with one exception:
    If a message is received that has a Content-Length header field value
    consisting of the same decimal value as a comma-separated list (<xref
    target="abnf.extension"/>) &mdash; for example, "Content-Length: 42, 42" &mdash;
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -3642,9 +3642,7 @@ Upgrade: websocket
 </t>
 <t>
    As a result, implementations &MUST-NOT; forward a message with a
-   Content-Length header field value that is known to be incorrect, unless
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3634,6 +3634,21 @@ Upgrade: websocket
    (<xref target="attack.protocol.element.length"/>).
 </t>
 <t>
+   Because it is used for message delimitation in HTTP/1.1, the Content-Length
+   header field value can affect the security of the protocol. This is true even
+   when the current connection is not using HTTP/1.1, because HTTP is an
+   intermediated protocol, and messages might be transmitted using other versions
+   protocol.
```
reschke, line 3641:
```suggestion
   of the protocol.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3634,6 +3634,21 @@ Upgrade: websocket
...
+</t>
+<t>
+   As a result, implementations &MUST-NOT; forward a message with a
+   Content-Length header field value that is known to be incorrect, unless
+   its value is being used for message delimitation. Specific versions of HTTP
+   place additional requirements on implementations to enforce this general one.
+</t>
+<t>
+   Likewise, implementations &MUST-NOT; forward a message with a Content-Length
+   header field value that does not match the ABNF above, with one exception.
```
reschke, line 3651:
```suggestion
   header field value that does not match the ABNF above, with one exception:
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3634,6 +3634,21 @@ Upgrade: websocket
...
+   Because it is used for message delimitation in HTTP/1.1, the Content-Length
+   header field value can affect the security of the protocol. This is true even
+   when the current connection is not using HTTP/1.1, because HTTP is an
+   intermediated protocol, and messages might be transmitted using other versions
+   protocol.
+</t>
+<t>
+   As a result, implementations &MUST-NOT; forward a message with a
+   Content-Length header field value that is known to be incorrect, unless
+   its value is being used for message delimitation. Specific versions of HTTP
```
reschke, line 3646: I don't get the "unless" clause. Should it say *not* used for message delimitation?

reschke (changes requested):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3634,6 +3634,21 @@ Upgrade: websocket
    (<xref target="attack.protocol.element.length"/>).
 </t>
 <t>
+   Because it is used for message delimitation in HTTP/1.1, the Content-Length
+   header field value can affect the security of the protocol. This is true even
+   when the current connection is not using HTTP/1.1, because HTTP is an
+   intermediated protocol, and messages might be transmitted using other versions
+   of the protocol.
```
royfielding, line 3641:
```suggestion
   Because Content-Length is used for message delimitation in HTTP/1.1,
   its field value can impact how the message is parsed by downstream
   recipients even when the immediate connection is not using HTTP/1.1.
   If the message is forwarded by a downstream intermediary, a Content-Length
   field value that is inconsistent with the received message framing might
   cause a security failure due to request smuggling or response splitting.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3634,6 +3634,21 @@ Upgrade: websocket
...
+</t>
+<t>
+   As a result, implementations &MUST-NOT; forward a message with a
+   Content-Length header field value that is known to be incorrect, unless
+   its value is being used for message delimitation. Specific versions of HTTP
+   place additional requirements on implementations to enforce this general one.
```
royfielding, line 3647:
```suggestion
   Content-Length header field value that is known to be incorrect.
```

draft-ietf-httpbis-semantics-latest.xml
royfielding: right, both that and the ending sentence are unnecessary.

royfielding (changes requested):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -3634,6 +3634,20 @@ Upgrade: websocket
...
+   field value that is inconsistent with the received message framing might
+   cause a security failure due to request smuggling or response splitting.
+</t>
+<t>
+   As a result, implementations &MUST-NOT; forward a message with a
+   Content-Length header field value that is known to be incorrect.
+</t>
+<t>
+   Likewise, implementations &MUST-NOT; forward a message with a Content-Length
+   header field value that does not match the ABNF above, with one exception:
```
royfielding, line 3650:
```suggestion
   As a result, a sender &MUST-NOT; forward a message with a
   Content-Length header field value that is known to be incorrect.
</t>
<t>
   Likewise, a sender &MUST-NOT; forward a message with a Content-Length
   header field value that does not match the ABNF above, with one exception:
```

royfielding (approved):

reschke (approved):

PR #761, opened 2021-02-10
Method semantics are inviolate

### LINKED ISSUE #691: Method semantics win
[Quoting from Section 3.1](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.3.1.p.2):

> One design goal of HTTP is to separate resource identification from request semantics, which is made possible by vesting the request semantics in the request method (Section 9) and a few request-modifying header fields. If there is a conflict between the method semantics and any semantic implied by the URI itself, as described in [Section 9.2.1](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#safe.methods), the method semantics take precedence.

This is good, but I think that the phrasing here could be more direct.  "any semantic implied by the URI itself" - when you read [the referenced text from Section 9.2.1](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.9.2.1.p.6) suggests that you might infer something from a URI about what it does.  However, I think that this underplays the key point.

The point that should be made here is that a resource cannot decide to change the semantics of a method to suit its purposes (I guess unless it is willing to assume all responsibility for doing so, the usual caveat we apply here for things like request logging).

Separately, the specific reference to 9.2.1 might be read to imply that safeness is special, but resources don't get to override the semantics of idempotent methods either (or any other well-defined properties of methods).

So I would instead phrase this more directly and use the citation as an example only:

> A resource cannot treat a request in a manner inconsistent with the semantics of the method of the request.  For example, though the URI of a resource might imply semantics that are not safe, a client can expect the resource to avoid actions that are unsafe when processing a request with a safe method; see Section 9.2.1.

martinthomson:
Closes #691.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -553,9 +553,11 @@
    request semantics, which is made possible by vesting the request
    semantics in the request method (<xref target="methods"/>) and a few
    request-modifying header fields.
-   If there is a conflict between the method semantics and any semantic
-   implied by the URI itself, as described in <xref target="safe.methods"/>,
-   the method semantics take precedence.
+   A resource cannot treat a request in a manner inconsistent with the
+   semantics of the method of the request. For example, though the URI of a
+   resource might imply semantics that are not safe, a client can expect the
+   resource to avoid actions that are unsafe when processing a request with a
+   safe method (see <xref target="safe.methods"/>).
 </t>
 <t>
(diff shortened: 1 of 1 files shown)


mnot:
Can you please factor the whitespace changes out into a separate PR?

martinthomson:
Sorry, my editor did that without my knowledge.  I didn't realize the draft was so messy to start with.  
Will do.

royfielding (approved):

PR #763, opened 2021-02-10
Cleanup whitespace

martinthomson:
Janitorial work: removing trailing whitespaces, replacing tabs.  Courtesy of emacs M-x whitespace-cleanup.


mnot:
I think we should sit on this until we merge a few more PRs, to make sure there aren't conflicts...

reschke:
I don't think this is a good idea. WS cleanup is nice, a good git history is more important though.

reschke:
OOBE see #818

PR #764, opened 2021-02-11
Explain implications of statelessness

### LINKED ISSUE #743: Semantics: 7.8 Upgrade
Clarify if the request needs to be the 1st request after a connection is opened.

mnot:
Fixes #743


royfielding (approved):

royfielding:
though I do wonder if HPACK/QPACK needs to be explained as well, since the order in which messages are processed does impact their compression, but not their interpretation.

martinthomson:
I would probably say that as the *semantics* are still stateless, this document probably doesn't need to say anything about that.  But that's a weak preference only.

reschke (approved):

reschke:
The change looks good to me. That said, it might indeed be good to mention the impact of header compression.

mnot:
I clarified that this was about semantics. I'm hesitant to talk about header compression, as that's likely to just confuse the matter.

royfielding:
I thought that was already clear. We might need to mention HPACK/QPACK anyway because they are only stateful with regard to the in-transit encoding between peers, and thus deliver stateless messages at the HTTP layer. Or something.

PR #767, opened 2021-02-11
Remove Canonicalization and Text Defaults

### LINKED ISSUE #703: Is a prohibition against mixed line termination a requirement?
[This text](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.8.3.3.p.2) implies a requirement: 

> HTTP allows the transfer of text media with plain CR or LF alone representing a line break, when such line breaks are consistent for an entire representation.

I would have said instead that HTTP itself doesn't care whether content adheres to this stricture, though some implementations could (or MAY) normalize line endings before sending or after receiving content that uses a `text` media type.

mnot:
Fixes #703.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -11823,16 +11823,6 @@ Content-Type: text/plain
   </x:source>
 </reference>
 
-<reference anchor="RFC2045">
-  <front>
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Not seeing anything here worth saving.

royfielding (approved):
This was a huge deal back in 1995-96 and was necessary to bless deployed practice of sending native line endings and expecting the user agents to process them. I guess we can ignore that now?

Don't be surprised if JK brings this up in IESG last call.

reschke (approved):

PR #768, opened 2021-02-11
Clarify large chunk handling issues

### LINKED ISSUE #749: Messaging: 7.1 chunked transfer coding: add a note about the risks of too large chunks
The text says

> Note that HTTP/1.1 does not define any means to limit the size of a chunked response such that an intermediary can be assured of buffering the entire response.

I'd want to add:

"Implementations must be careful about accurately parsing large values or rejecting the messages, as failure to accurately represent the advertised value due to overflows or loss of precision may have serious security consequences."

I think it is important to remind, because implementations seem to have progressively shifted towards 64 bit repressentations since 7230, 32-bit ones are still much present, and the risk of desynchronization by advertising more than 4GB is high. By the way, some languages do no provide more than 52 bits of accuracy because they use floats to represent any number... The risk remains low (1 injected byte every 4 PB) but its worth asking to be careful.

mnot:
Fixes #749.


royfielding (approved):

royfielding:
Note that if we say it has serious security hazards and we don't mention that in Security Considerations, someone might complain.

reschke (approved):

mnot:
PTAL

reschke (approved):

royfielding (approved):

PR #770, opened 2021-02-16
Clarify what cache proxy considerations apply to

### LINKED ISSUE #729: Identity of the proxy in the cache key
[This text](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#rfc.section.3.1.p.1) says that a proxy doesn't store things like Proxy-Authenticate unless the cache key includes the proxy identity.

> Header fields that are specific to a client's proxy configuration MUST NOT be stored, unless the cache incorporates the identity of the proxy into the cache key.

Should this be the identity of the proxy's client instead? Maybe the intent of this was to say that a proxy cache that uses authentication is prohibited from storing these header fields unless they might be valid when reused.  But including the identity of the *proxy* (as opposed to the user of the proxy) could lead to these fields always being stored, because the proxy identity is probably the same for all its users.

The effect of including the identity of the *client* in the cache key would be to effectively make the responses containing these fields private, which works to ensure that different clients don't end up with bad authentication challenges being sent to them from cache.  That might be OK.

That all said, caching something like Proxy-Authenticate and playing it back later seems like it might cause problems too, so maybe I'm missing something.

mnot:
Fixes #729.

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -379,7 +379,7 @@
    private (<xref target="cache-response-directive.private"/>) cache
    directives can have arguments that prevent storage of header fields by all
    caches and shared caches, respectively.</li>
-   <li>Header fields that are specific to a client's proxy configuration
+   <li>Header fields that are specific to the proxy that a cache uses when forwarding a request
    &MUST-NOT; be stored, unless the cache incorporates the identity of the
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Clearer now.

reschke (approved):

royfielding (approved):

PR #773, opened 2021-02-16
Discuss forwarding and extensibility

### LINKED ISSUE #692: Assumed properties of new methods
In general, an unknown or new method can be handled by a resource by rejecting the request (with 406 for instance).  However, some entities (intermediaries in particular) might be forced to engage with methods that are unknown to them.

I couldn't find text in [Section 7.6](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#message.forwarding) or Section 16.1 on this.  However, the draft might benefit from advice on this.  There are a few things that might be said, with reference to Section 9, which already says much of this:

* Intermediaries are expected to forward new methods.  This enables the deployment of new methods.

* New methods might be safe, idempotent, or cacheable.  Or they might not be.  Actions taken need to consider this possibility and act accordingly.  For instance, an intermediary that might be able to automatically retry a request cannot do so for an unknown method, which it does not know whether it is idempotent (9.2) or a cache can't cache a response unless the method is known to be cacheable (9.3).

mnot:
Fixes #692.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -2717,6 +2717,11 @@ Host: www.example.org
    can enhance (or interfere) with either direction of the stream.
 </t>
 <t>
+   Intermediaries are expected to forward messages even when protocol elements
+   such as methods, status codes, and fields are not recognized, as doing so
+   preserves the extensibility of the protocol.
+</t>
+<t>
    An intermediary not acting as a tunnel &MUST; implement the
    <x:ref>Connection</x:ref> header field, as specified in
    <xref target="field.connection"/>, and exclude fields from being forwarded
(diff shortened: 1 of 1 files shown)


martinthomson (approved):

martinthomson:
I'm disappointed that RFC 6919 doesn't list "ARE EXPECTED TO".

mnot:
Time for a bis?

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2716,6 +2716,11 @@ Host: www.example.org
    architecture, there are no inherent limits to the extent an intermediary
    can enhance (or interfere) with either direction of the stream.
 </t>
+<t>
+   Intermediaries are expected to forward messages even when protocol elements
+   such as methods, status codes, and fields are not recognized, as doing so
+   preserves the extensibility of the protocol.
```
royfielding, line 2722:
```suggestion
   Intermediaries are expected to forward messages even when protocol elements
   are not recognized (e.g., new methods, status codes, or field names), since that
   preserves extensibility for downstream recipients.
```

royfielding (approved):

PR #774, opened 2021-02-16
Refine version requirement for intermediaries

### LINKED ISSUE #751: messaging 2.3: version (rewriting by intermediaries)
It's written:

> Intermediaries that process HTTP messages ... MUST send their own HTTP-version in forwarded messages```

Actually I'd rather say "MUST send a version no higher than their own in forwarded messages" (poor wording, I know, maybe someone can propose better).

Indeed, if an intermediary receives an HTTP/1.0 request and passes it as HTTP/1.1, the server will wrongly assume that the client can deal with 1.1 (e.g. chunks) and the message may have to be degraded by the intermediary (such as de-chunking and rely on close only). Furthermore, seeing 1.0 for a server is often an indication of very limited (or possibly bogus) client. For example some intermediaries might avoid compressing or caching when facing HTTP/1.0 messages, and as such it's preferable to let such versions be properly advertised in messages that could be considered as potentially unsafe.

Maybe a different wording should be "intermediaries... MUST NOT pass a message showing a version they do not support, and MUST make sure the message always conforms to the advertised version".

mnot:
Fixes #751.

### DIFF
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -405,7 +405,8 @@
 <t>
    Intermediaries that process HTTP messages (i.e., all intermediaries
    other than those acting as tunnels) &MUST; send their own HTTP-version
-   in forwarded messages.  In other words, they are not allowed to blindly
+   in forwarded messages, unless it is purposefully downgraded as a workaround
+   for an upstream issue. In other words, they are not allowed to blindly
    forward the <x:ref>start-line</x:ref> without ensuring that the
    protocol version in that message matches a version to which that
    intermediary is conformant for both the receiving and
@@ -3094,6 +3095,7 @@ https://www.example.org
 
 <section title="Since draft-ietf-httpbis-messaging-14" anchor="changes.since.14">
 <ul x:when-empty="None yet.">
+   <li>In <xref target="http.version"/>, refine version requirements for intermediaries (<eref target="https://github.com/httpwg/http-core/issues/751"/>)</li>
 </ul>
 </section>
 </section>
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -406,7 +406,7 @@
    Intermediaries that process HTTP messages (i.e., all intermediaries
    other than those acting as tunnels) &MUST; send their own HTTP-version
    in forwarded messages, unless it is purposefully downgraded as a workaround
-   for an upstream issue. In other words, they are not allowed to blindly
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -405,7 +405,8 @@
 <t>
    Intermediaries that process HTTP messages (i.e., all intermediaries
    other than those acting as tunnels) &MUST; send their own HTTP-version
-   in forwarded messages.  In other words, they are not allowed to blindly
+   in forwarded messages, unless it is purposefully downgraded as a workaround
```
martinthomson, line 408:
```suggestion
   in forwarded messages, unless a request is purposefully downgraded as a workaround
```

This only applies to requests, right?  The expectation is that a server is able to produce a lower-version in its response.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -405,7 +405,8 @@
...
+   for an upstream issue. In other words, they are not allowed to blindly
    forward the <x:ref>start-line</x:ref> without ensuring that the
    protocol version in that message matches a version to which that
    intermediary is conformant for both the receiving and
```
martinthomson, line 412: This sentence is very long and convoluted.  While you are here, why not just:

> forward the start-line without ensuring that they support the indicated version.  Forwarding an HTTP message without rewriting [...]

(I think that implies both the sending of requests and the receiving of responses amply.)

martinthomson (approved):

reschke (approved):
but see feedback from  @martinthomson

draft-ietf-httpbis-messaging-latest.xml
mnot: While Willie talked about the request side, it might be relevant on the response side too.

royfielding: both ways

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -405,7 +405,8 @@
 <t>
    Intermediaries that process HTTP messages (i.e., all intermediaries
    other than those acting as tunnels) &MUST; send their own HTTP-version
-   in forwarded messages.  In other words, they are not allowed to blindly
+   in forwarded messages, unless it is purposefully downgraded as a workaround
+   for an upstream issue. In other words, they are not allowed to blindly
```
royfielding, line 409:
```suggestion
   for an upstream issue. In other words, an intermediary is not allowed to blindly
```

royfielding (approved):

PR #775, opened 2021-02-16
A CONNECT tunnel starts after the response header section

### LINKED ISSUE #737: SEMANTICS describes CONNECT in HTTP/1.1 terms
I think the goal was for SEMANTICS to stay out of message framing and leave that to the individual mappings.  It currently says this about CONNECT:

> Any 2xx (Successful) response indicates that the sender (and all inbound proxies) will switch to tunnel mode immediately after the **blank line that concludes the successful response's header section; data received after that blank line** is from the server identified by the request target.

Perhaps this should simply talk about data received after the header section, and leave the boundaries of that version-specific?

royfielding:
Fixes #737

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -4918,21 +4918,20 @@ Host: server.example.com
    into the protocol's wire format.
 </t>
 <t>
-   CONNECT is intended only for use in requests to a proxy.
+   CONNECT is intended for use in requests to a proxy.
+   The recipient can establish a tunnel either by directly connecting to
+   the server identified by the request target or, if configured to use
+   another proxy, by forwarding the CONNECT request to the next inbound proxy.
    An origin server &MAY; accept a CONNECT request, but most origin servers
    do not implement CONNECT.
 </t>
 <t>
-   The recipient proxy can establish a tunnel either by directly connecting to
-   the request target or, if configured to use another proxy, by forwarding
-   the CONNECT request to the next inbound proxy.
    Any <x:ref>2xx (Successful)</x:ref> response indicates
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4918,21 +4918,20 @@ Host: server.example.com
...
-   The recipient proxy can establish a tunnel either by directly connecting to
-   the request target or, if configured to use another proxy, by forwarding
-   the CONNECT request to the next inbound proxy.
    Any <x:ref>2xx (Successful)</x:ref> response indicates
    that the sender (and all inbound proxies) will switch to tunnel mode
-   immediately after the blank line that concludes the successful response's
-   header section; data received after that blank line is from the server
-   identified by the request target.
+   immediately after the response header section; data received after that
+   header section is from the server identified by the request target.
```
martinthomson, line 4932: I think that this is OK.  HTTP/2 and /3 have subtly different treatment here in that they only forward DATA frames to the tunnel, whereas HTTP/1.1 has no similar means of distinguishing tunnel data from other stuff.  I think that's OK at this level though.

martinthomson (approved):

mnot (approved):

reschke (approved):

MikeBishop (approved):
This is good for the issue in question.  I wonder if it's also worth mentioning that any data following the request header section is prospectively-sent data for the tunnel, even if the proxy hasn't sent a response yet?

reschke:
@MikeBishop  - I'll merge this; if you feel we need to address https://github.com/httpwg/http-core/pull/775#pullrequestreview-592301267 please open a new issue.

PR #781, opened 2021-02-24
Adjust handling of invalid Age values

### LINKED ISSUE #778: Last minute Age adjustments
Apologies for the lateness, but going through [my tests](https://cache-tests.fyi/#age-parse), I think we need to adjust handling of invalid `Age` headers:

* Right now, it's [specified](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#field.age) that caches SHOULD treat responses with invalid `Age` values as stale. However:
  * No cache tested treats a response with `Age: abc` in this manner; all ignore the field.
  * Only Fastly and Firefox treat a response with `Age: -7200` in this manner; the rest ignore the field.
  * Interestingly, most caches _do_ consider a response with `Age: 7200.0` to be stale, except for Apache and Chrome.

* We specify that the first field line should be used when multiple are present. That's almost universally adhered to (with one deviation by Safari), but we omit multiple values on the same field line from this rule (which are treated as invalids as per above). However:
   * When that line is `0, something_old`, all caches reuse the response.
   * When that line is `something_old, 0`, only Safari and Apache reuse it.

This suggests that we could align with current implementations and improve interoperability by:
1. Specifying that invalid `Age` fields should be ignored, not make the response stale. This is nearly universal practice.
2. Specifying that the first `Age` field value should be used when multiple can be parsed. This is also nearly universal practice.

mnot:
Fixes #778.

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1215,7 +1215,7 @@
    first member of the field value, discarding subsequent ones.
 </t>
 <t>
-   If the field value (after discarding additional lines, as per above) is invalid
+   If the field value (after discarding additional members, as per above) is invalid
    (e.g., it contains something other than a non-negative integer),
    a cache &SHOULD; ignore the field.
 </t>
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1211,13 +1211,13 @@
 </t>
 <t>
    Although it is defined as a singleton header field, a cache encountering a
-   message with multiple Age field lines &SHOULD; use the
-   first field line, discarding subsequent ones.
+   message with a list-based Age field value &SHOULD; use the
+   first member of the field value, discarding subsequent ones.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1211,13 +1211,13 @@
...
-   message with multiple Age field lines &SHOULD; use the
-   first field line, discarding subsequent ones.
+   message with a list-based Age field value &SHOULD; use the
+   first member of the field value, discarding subsequent ones.
 </t>
 <t>
    If the field value (after discarding additional lines, as per above) is invalid
-   (e.g., it contains a list or something other than a non-negative integer),
-   a cache &SHOULD; consider the response to be stale.
+   (e.g., it contains something other than a non-negative integer),
```
reschke, line 1219: "as per above" is not longer true...

reschke (changes requested):

reschke (approved):

royfielding (approved):

PR #783, opened 2021-02-24
Restore recipient requirements for invalid characters in field values

### LINKED ISSUE #683: field-value production rules out CTL characters, but these are common in the wild
draft-ietf-httpbis-semantics-latest currently says:

> Field values containing control (<x:ref>CTL</x:ref>) characters such as CR or LF are invalid; recipients &MUST; either reject a field value containing control characters, or convert them to SP before processing or forwarding the message.

However, when I implemented an HTTP/1.1 parsing library that followed this rule, it turned out to broken in the real world. Specifically, users reported that sites using Google Analytics were setting cookies with values containing the ASCII character `\x01`, which is a control character (https://github.com/python-hyper/h11/issues/57). And in my investigations at the time, browsers and popular clients like curl all supported this just fine.

It would be great if the next round of RFCs could look into this in more detail and define the `field-content` production in a way that would be acceptable to e.g. the browsers.

In my library we currently reject NUL and exotic whitespace (`\r`, `\v`, `\f`, `\n`), on the grounds that those are high-risk for interop bugs and splitting exploits, but allow all other bytes to pass through, including control characters. I don't know if that's the best solution, but in our experience it's at least closer to being real-world compatible than what the RFC draft currently says.

mnot:
Fixes #683.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -12979,7 +12979,7 @@ Content-Type: text/plain
    (<xref target="resources"/>)
 </t>
 <t>
-   Clarify that control characters in field values are to be rejected or
+   Clarify that CR and NUL in field values are to be rejected or
    mapped to SP.
    (<xref target="field-values"/>)
 </t>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -1650,9 +1650,13 @@ Content-Type: text/plain
    opaque data.
 </t>
 <t>
-   Field values containing CR or NUL characters are invalid; recipients
-   &MUST; either reject the message, or convert these characters to SP before
-   processing or forwarding the message.
+   Field values containing CR or NUL characters are invalid and dangerous, due
+   to the varying ways that implementations might parse and interpret those characters;
+   a recipient of CR or NUL within a field value &MUST; either reject the message or
+   replace each of those characters with SP before further processing or forwarding
+   of that message. Field values containing other CTL characters are also invalid;
+   however, recipients &MAY; retain such characters for the sake of robustness if
+   they only appear within safe field value contexts (e.g., data not required by HTTP).
 </t>
 <t>
    Leading and trailing whitespace in raw field values is removed upon field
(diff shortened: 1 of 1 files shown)


reschke (changes requested):
I though the proposal was to go back to what RFC723* said - CTL characters are invalid, but do not require them to be rejected or converted.

mnot:
That's what this does; the text above already has that effect.

It does raise the question of whether we should add CTL to `obs-text`, however. Thoughts?

reschke:
> That's what this does; the text above already has that effect.

I believe the proposed text now is somewhat misleading, as it does *not* mention that CTLs are (and always have been invalid.

> It does raise the question of whether we should add CTL to obs-text, however. Thoughts?

No, that would be a breaking change.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1650,10 +1650,9 @@ Content-Type: text/plain
    opaque data.
 </t>
 <t>
-   Field values containing control (<x:ref>CTL</x:ref>) characters such as
-   CR or LF are invalid; recipients &MUST; either reject a field value
-   containing control characters, or convert them to SP before processing or
-   forwarding the message.
+   Field values containing CR or NUL characters are invalid; recipients
+   &MUST; either reject the message, or convert these characters to SP before
+   processing or forwarding the message.
```
royfielding, line 1655:
```suggestion
   Field values containing CR or NUL characters are invalid and dangerous, due
   to the varying ways that implementations might parse and interpret those characters;
   a recipient of CR or NUL within a field value &MUST; either reject the message or
   replace each of those characters with SP before further processing or forwarding
   of that message. Field values containing other CTL characters are also invalid;
   however, recipients &MAY; retain such characters for the sake of robustness if
   they only appear within safe field value contexts (e.g., data not required by HTTP).
```

royfielding (changes requested):
How about this variation?

reschke:
We still need to rewrite or remove

> Clarify that control characters in field values are to be rejected or mapped to SP. (Section 5.5)

reschke (approved):

royfielding (approved):

PR #784, opened 2021-02-24
Refine Referer

### LINKED ISSUE #709: Important notes regarding Referer
[Section 10.1.3](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#field.referer) defines Referer and says:

> Most general-purpose user agents do not send the Referer header field when the referring resource is a local "file" or "data" URI. 

There are two other things that I think are relevant to mention here:

1. Referer is often suppressed when the referring resource is "https" at a different origin than the request target.
2. Referer can contain only an origin rather than the referring resource identity.

[Referrer-Policy](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-header) might not be worth citing here as it is probably too specific to browsers, but these other constraints are worth noting.  Especially the first as this has real security consequences.  Though the text in [Section 17.9](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#sensitive.information.in.uris) is excellent as a high-level principle, the steps that are taken to avoid URI leakage are meaningful and very relevant to this section.

mnot:
Fixes #709

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -5320,7 +5320,7 @@ Expect: 100-continue
    user agents do not send the Referer header field when the referring
    resource is a local "file" or "data" URI. A user agent &SHOULD-NOT; send a
    <x:ref>Referer</x:ref> header field if the referring resource was accessed with
-   a secure protocol, and the request target has an origin differing from that
+   a secure protocol and the request target has an origin differing from that
    of the referring resource, unless the referring resource explicitly allows it.
    A user agent &MUST-NOT; send a
    <x:ref>Referer</x:ref> header field in an unsecured HTTP request if the
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -5309,7 +5309,7 @@ Expect: 100-continue
 </t>
 <t>
    The Referer header field value need not convey the full URI of the referring
-   page; it &MAY; be truncated to the referring origin.
+   resource; it &MAY; be truncated to the referring origin.
 </t>
 <t>
    The Referer header field has the potential to reveal information about the request
@@ -5319,12 +5319,12 @@ Expect: 100-continue
    behind a firewall or internal to a secured service). Most general-purpose
    user agents do not send the Referer header field when the referring
    resource is a local "file" or "data" URI. A user agent &SHOULD-NOT; send a
-   <x:ref>Referer</x:ref> header field if the referring page was received with
+   <x:ref>Referer</x:ref> header field if the referring resource was accessed with
    a secure protocol, and the request target has an origin differing from that
-   of the referring page, unless the referring page explicitly allows it.
+   of the referring resource, unless the referring resource explicitly allows it.
    A user agent &MUST-NOT; send a
    <x:ref>Referer</x:ref> header field in an unsecured HTTP request if the
-   referring page was received with a secure protocol.
+   referring resource was accessed with a secure protocol.
    See <xref target="sensitive.information.in.uris"/> for additional
    security considerations.
 </t>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -5309,7 +5309,7 @@ Expect: 100-continue
 </t>
 <t>
    The Referer header field value need not convey the full URI of the referring
-   resource; it &MAY; be truncated to the referring origin.
+   resource; a user agent &MAY; truncate parts other than the referring origin.
 </t>
 <t>
    The Referer header field has the potential to reveal information about the request
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -5308,13 +5308,21 @@ Expect: 100-continue
    or send it with a value of "about:blank".
 </t>
 <t>
+   The Referer header field value need not convey the full URI of the referring
+   page; it &MAY; be truncated to the referring origin.
+</t>
+<t>
    The Referer header field has the potential to reveal information about the request
    context or browsing history of the user, which is a privacy concern if the
    referring resource's identifier reveals personal information (such as an
    account name) or a resource that is supposed to be confidential (such as
    behind a firewall or internal to a secured service). Most general-purpose
    user agents do not send the Referer header field when the referring
-   resource is a local "file" or "data" URI. A user agent &MUST-NOT; send a
+   resource is a local "file" or "data" URI. A user agent &SHOULD-NOT; send a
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5307,14 +5307,22 @@ Expect: 100-continue
...
    context or browsing history of the user, which is a privacy concern if the
    referring resource's identifier reveals personal information (such as an
    account name) or a resource that is supposed to be confidential (such as
    behind a firewall or internal to a secured service). Most general-purpose
    user agents do not send the Referer header field when the referring
-   resource is a local "file" or "data" URI. A user agent &MUST-NOT; send a
+   resource is a local "file" or "data" URI. A user agent &SHOULD-NOT; send a
+   <x:ref>Referer</x:ref> header field if the referring page was received with
+   a secure protocol, and the request target has an origin differing from that
+   of the referring page, unless the referring page explicitly allows it.
```
martinthomson, line 5324: The "referring page" is a bit of a weird concept here.  I know that the spec is bad about letting browser UI leak into the language in places, but it would be good if it didn't have to here.

In essence, it's the referring resource that makes the call here, but it's a little indirect even then.  Can this just be:

```suggestion
   of the referring resource, unless the referring resource explicitly allows it.
```

Using resource throughout is probably OK, even allowing for the fact that referrer policy isn't necessarily expressed by the referring resource.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5307,14 +5307,22 @@ Expect: 100-continue
...
+   A user agent &MUST-NOT; send a
    <x:ref>Referer</x:ref> header field in an unsecured HTTP request if the
    referring page was received with a secure protocol.
```
martinthomson, line 5327:
```suggestion
   referring resource was received with a secure protocol.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5307,14 +5307,22 @@ Expect: 100-continue
...
 <t>
    The Referer header field has the potential to reveal information about the request
    context or browsing history of the user, which is a privacy concern if the
    referring resource's identifier reveals personal information (such as an
    account name) or a resource that is supposed to be confidential (such as
    behind a firewall or internal to a secured service). Most general-purpose
    user agents do not send the Referer header field when the referring
-   resource is a local "file" or "data" URI. A user agent &MUST-NOT; send a
+   resource is a local "file" or "data" URI. A user agent &SHOULD-NOT; send a
+   <x:ref>Referer</x:ref> header field if the referring page was received with
```
martinthomson, line 5322:
```suggestion
   <x:ref>Referer</x:ref> header field if the referring resource was received with
```

martinthomson (approved):
This is right.

reschke (approved):
with the changes suggested by @martinthomson

draft-ietf-httpbis-semantics-latest.xml
mnot: I used `page` because that was the already existing language. If we want to change it, it should probably be changed to `representation`; you can't receive a resource (at least not in the HTTPWG; WHATWG may feel differently).

martinthomson: As I said, we're talking about an indirection at some level here, but it is still the resource.

In one case, the resource asks by using Referrer-Policy, which is - I think - pretty directly the resource acting here.  In some of the other cases (`<meta>`, link decorations), the requesting is indirect in that the resource provides a representation that contains the request, but the effect is still the same.

"Page" is just the wrong abstraction.  As "resource" is used a few paragraphs up in the introduction of the field, consistency can't be the reason to use "page".

mnot: The resource acts by sending `Referrer-Policy`, but the resource is _not_ "received". "Accessed", perhaps.

draft-ietf-httpbis-semantics-latest.xml
royfielding:
```suggestion
   <x:ref>Referer</x:ref> header field if the reference was received with
```

draft-ietf-httpbis-semantics-latest.xml
martinthomson: I can live with that.  I'm just objecting to the use of "page".

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5307,16 +5307,24 @@ Expect: 100-continue
    bookmarks/favorites), the user agent &MUST; either exclude the Referer header field
    or send it with a value of "about:blank".
 </t>
+<t>
+   The Referer header field value need not convey the full URI of the referring
+   resource; it &MAY; be truncated to the referring origin.
```
royfielding, line 5312:
```suggestion
   resource; a user agent &MAY; truncate parts other than the referring origin.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5307,16 +5307,24 @@ Expect: 100-continue
...
    The Referer header field has the potential to reveal information about the request
    context or browsing history of the user, which is a privacy concern if the
    referring resource's identifier reveals personal information (such as an
    account name) or a resource that is supposed to be confidential (such as
    behind a firewall or internal to a secured service). Most general-purpose
    user agents do not send the Referer header field when the referring
-   resource is a local "file" or "data" URI. A user agent &MUST-NOT; send a
+   resource is a local "file" or "data" URI. A user agent &SHOULD-NOT; send a
+   <x:ref>Referer</x:ref> header field if the referring resource was accessed with
+   a secure protocol, and the request target has an origin differing from that
```
royfielding, line 5323:
```suggestion
   a secure protocol and the request target has an origin differing from that
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -5307,16 +5307,24 @@ Expect: 100-continue
...
    referring resource's identifier reveals personal information (such as an
    account name) or a resource that is supposed to be confidential (such as
    behind a firewall or internal to a secured service). Most general-purpose
    user agents do not send the Referer header field when the referring
-   resource is a local "file" or "data" URI. A user agent &MUST-NOT; send a
+   resource is a local "file" or "data" URI. A user agent &SHOULD-NOT; send a
+   <x:ref>Referer</x:ref> header field if the referring resource was accessed with
+   a secure protocol and the request target has an origin differing from that
+   of the referring resource, unless the referring resource explicitly allows it.
+   A user agent &MUST-NOT; send a
```
royfielding, line 5325:
```suggestion
   of the referring resource, unless the referring resource explicitly allows
   Referer to be sent. A user agent &MUST-NOT; send a
```

royfielding (approved):

PR #786, opened 2021-02-25
The extent of a protection space is unknowable

### LINKED ISSUE #710: Realm is under-specified
[Realm is described](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.11.5.p.1) as:

> The realm authentication parameter is reserved for use by authentication schemes that wish to indicate a scope of protection.

This says that it is a parameter, but does it appear on challenges or responses?  [Section 11.2](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#auth.params) only establishes that authentication parameters parameterize authentication schemes, there is no mention of how those relate to what is sent.

I think that this first problem only requires a mention of WWW-Authenticate and Proxy-Authenticate.

The second problem is in [this text](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.11.5.p.3):

> The protection space determines the domain over which credentials can be automatically applied. If a prior request has been authorized, the user agent MAY reuse the same credentials for all other requests within that protection space for a period of time determined by the authentication scheme, parameters, and/or user preferences (such as a configurable inactivity timeout). 

Clients would seem to have no way of knowing whether reuse is likely to be successful.  A protection space is defined as the tuple of origin and realm, but there is no acknowledgment that how a protection space might correspond to the URI space is only known to the server.  (The next sentence, which I omitted from this quote acknowledges that the client needs special knowledge in order to understand that a protection space might span origins; that's very useful information.)

I think that this requires only that the text acknowledge this uncertainty and note that clients could decide to provide authentication information on every request made to the origin, without knowledge of the extent of the protection space.  It might also note that particular authentication schemes might define mechanisms that allow clients to decide where to use credentials.  RFC 7616 defines `domain`, which allows for scoping; RFC 7617 has a section on [reusing credentials](https://tools.ietf.org/html/rfc7617#section-2.2).

martinthomson:
When the text suggests that credentials can be automatically applied to all
requests made within a protection space, that begs the question: how do I know
what requests are in the same protection space?

The answer is "well, you don't really".  Authentication schemes might define
something, but otherwise you are left to guess.  This PR says that as directly
as I could manage.

I considered adding another sentence here that says "In the absence of
specific information about the extent of a protection space, clients &MAY;
assume that the protection space extent is the origin of the server."  I'd
like thoughts on whether that is helpful.  I think that it might be, as otherwise
this isn't really implementable.

Closes #710.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -13465,6 +13465,7 @@ Content-Type: text/plain
   <li>In <xref target="field.content-length"/>, be more explicit about invalid and incorrect values (<eref target="https://github.com/httpwg/http-core/issues/748"/> and <eref target="https://github.com/httpwg/http-core/issues/749"/>)</li>
   <li>Move discussion of statelessness from <xref target="intermediaries"/> to <xref target="connections"/> (<eref target="https://github.com/httpwg/http-core/issues/753"/>)</li>
   <li>In <xref target="status.101"/>, clarify that the upgraded protocol is in effect after the 101 response (<eref target="https://github.com/httpwg/http-core/issues/776"/>)</li>
+  <li>In <xref target="protection.scope"/>, explain that the protection scope is not defined without additional information (<eref target="https://github.com/httpwg/http-core/issues/710"/>)</li>
 </ul>
 </section>
 </section>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -5917,9 +5917,15 @@ Expect: 100-continue
    user agent &MAY; reuse the same credentials for all other requests within
    that protection space for a period of time determined by the authentication
    scheme, parameters, and/or user preferences (such as a configurable
-   inactivity timeout). Unless specifically allowed by the authentication
-   scheme, a single protection space cannot extend outside the scope of its
-   server.
+   inactivity timeout).
+</t>
+<t>
+   The extent of a protection space, and therefore the requests to which
+   credentials might be automatically applied, is not necessarily known to
+   clients without additional information. An authentication scheme might
+   define parameters that describe the extent of a protection space. Unless
+   specifically allowed by the authentication scheme, a single protection
+   space cannot extend outside the scope of its server.
 </t>
 <t>
    For historical reasons, a sender &MUST; only generate the quoted-string syntax.


mnot:
Please factor out whitespace changes...

martinthomson:
Ugh, it's like picking up toys after children.

mnot (approved):

reschke (approved):
(but should be added to change log)

martinthomson:
Neither of you have comments about saying what the extent of the protection scope is in the absence of information?

mnot:
I'd rather leave that out.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -13459,6 +13465,7 @@ Content-Type: text/plain
   <li>In <xref target="field.content-length"/>, be more explicit about invalid and incorrect values (<eref target="https://github.com/httpwg/http-core/issues/748"/> and <eref target="https://github.com/httpwg/http-core/issues/749"/>)</li>
   <li>Move discussion of statelessness from <xref target="intermediaries"/> to <xref target="connections"/> (<eref target="https://github.com/httpwg/http-core/issues/753"/>)</li>
   <li>In <xref target="status.101"/>, clarify that the upgraded protocol is in effect after the 101 response (<eref target="https://github.com/httpwg/http-core/issues/776"/>)</li>
+  <li>In <xref target="protection.scope"/>, explain that the protection scope is not defined without additional information (<eref target="https://github.com/httpwg/http-core/issues/710"/>)</li>
```
reschke, line 13468: ...please move up (these are sorted by github issue #)

reschke:
> Neither of you have comments about saying what the extent of the protection scope is in the absence of information?

AFAIU, it depends on the authentication scheme.

At the end of the day, we can only document here what we inherited from earlier specs, and what is actually done in practice. If you have information on the latter which helps us making this better, great. Do you?

martinthomson:
I proposed maybe adding:

> In the absence of specific information about the extent of a protection space, clients &MAY; assume that the protection space extent is the origin of the server.

Which I believe matches what clients do when they produce credentials and the authentication scheme doesn't define a clear scope.  Do you think that helps?

reschke:
Not sure. What problem are we trying to solve here? Auth scheme definitions that fail to define the protection space? In which case, shouldn't we say something about this in the "Considerations for new schemes" section?

royfielding (approved):

PR #787, opened 2021-02-25
Change semantic conformance requirement to an explanation and permision to workaround

### LINKED ISSUE #687: MUST NOT lie
> A sender MUST NOT generate protocol elements that convey a meaning that is known by that sender to be false. 

-- https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.2.2.p.5

This doesn't seem to be enforceable in any way.  This isn't really talking about whether the syntax is valid, but about the semantics, so it's easy to disavow knowledge of falsehood (my browser isn't necessary aware that me typing "the sky is pink" is false).  I think that it might be better to simply observe that it is contrary to function - and possibly interests - of a sender to generate protocol elements that convey semantics that are incorrect or contrary to the interests of the sender.

Normative language implies interoperability failure if it is ever violated, and I don't think that is the case here.

royfielding:
Fixes #687


mnot (approved):

martinthomson (approved):
This works very nicely.  Thanks Roy.

You can point very clearly at the DNT problems, but it's very subtle.

reschke (approved):

PR #790, opened 2021-03-04
Remove TRACE requirement on message/http media type (last normative ref to HTTP/1.1)

### LINKED ISSUE #690: Is -messaging part of the core definition of HTTP?
Sorry for this being a question.

I answered the same question for -caching by searching through and realizing that while caching might be a version-independent extension, a practical consequence of the structure of documents is that it is inextricably linked to -semantics.  That's cool.  Having the definition of HTTP/N rely on -caching is eminently sensible.

There is a normative reference to -messaging though.  Why would HTTP/3 care about that?  If this were software, I'd be horrified by pulling in a dependency like that.

Most of the references to -messaging are strictly examples.  I could only see two cases where there was a potentially normative dependency: [TRACE](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#TRACE) recommends "message/http" and [content codings prohibit overlap with transfer codings](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#content.coding.registry).  These seem like they could be broken (the second easier than the first, certainly) and the reference made normative.

mnot:
Fixes #690.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -13042,6 +13042,11 @@ Content-Type: text/plain
    (<xref target="OPTIONS"/>)
 </t>
 <t>
+   Removed normative requirement to use the "message/http" media type in
+   TRACE responses.
+   (<xref target="TRACE"/>)
+</t>
+<t>
    Restore list-based grammar for <x:ref>Expect</x:ref> for compatibility with
    RFC 2616.
    (<xref target="field.expect"/>)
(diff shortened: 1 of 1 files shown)


reschke:
I think it would be better to separate this into the purely editorial changes (all except TRACE), and then give the change to TRACE the full treatment separately (it's a normative change after all).

Will work on the former first.

reschke (approved):

royfielding (approved):

PR #791, opened 2021-03-04
whitespace removal is at time of use, not parsing

### LINKED ISSUE #697: Whitespace is not removed from field values in HTTP/2 or HTTP/3
[This claim](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.5.5.p.5) in Section 5.5 is likely incorrect:

> Leading and trailing whitespace in raw field values is removed upon field parsing (e.g., Section 5.1 of [Messaging]).

Does it need qualification (in some versions) or permission (could be removed)?

mnot:
Fixes #697


mnot:
Note that this is still operating on the leading and trailing whitespace on a field value, not a field line value. That makes me wonder whether text like this:

>    A recipient &MAY; combine multiple field lines within a field section that
   have the same field name
   into one field line, without changing the semantics of the message, by
   appending each subsequent non-empty field line value to the initial field line value
   in order, separated by a comma and <x:ref>OWS</x:ref> (optional whitespace).
   For consistency, use comma SP.

is really correct; do we need to say something about leading and trailing whitespace on field line values?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1678,8 +1678,8 @@ Content-Type: text/plain
    they only appear within safe field value contexts (e.g., data not required by HTTP).
 </t>
 <t>
-   Leading and trailing whitespace in raw field values is removed upon field
-   parsing (e.g., <xref target="field.parsing"/>). Field definitions where leading or trailing
+   Leading and trailing whitespace in field values &MUST; be removed before the
+   field value is used (but not forwarded). Field definitions where such
    whitespace in values is significant will have to use a container syntax such
    as quoted-string (<xref target="quoted.strings"/>).
```
scalablecory, line 1684: This language to me might suggest that forwarders should not trim whitespace and that it may have some semantic value. It seems better to explicitly call out that scenario as a MAY exception.

Something like:

```suggestion
   Leading and trailing whitespace in field values &MUST; be removed before the
   field value is used. Field definitions where such
   whitespace in values is significant will have to use a container syntax such
   as quoted-string (<xref target="quoted.strings"/>).
   
   A request being forwarded &MAY; leave whitespace intact.
```

reschke (commented):
+1 to making the forward case a bit clearer.

Also: if we are adding a "MUST" here, is this a new normative requirement to be listed in Changes?

mnot:
I'll adjust the language around forwarding (it's not just requests).

WRT change notes -- this is just surfacing a requirement that was latent in the ABNF before; I don't think it's necessary to list.

reschke (changes requested):
We are changing "is removed" to "MUST be removed".

This IMHO is because in HTTP/1.1, this was a side effect of parsing, whereas in HTTP/2 and 3 it is not. So essentially we are saying: if your wire format *can* transport leading/trailing whitespace, you *MUST* remove it before consumption.

That seems like a new normative requirement on wire formats to me.

mnot:
Change note added. Please also see question above re: field lines.

reschke (approved):

martinthomson (approved):

mnot:
@royfielding can you take a look at this?

royfielding (changes requested):
I looked at it, decided that the change was "correct" but that the section doesn't make any sense give the ABNF specifically excludes leading and trailing whitespace, tried to rewrite the section, and then got overwhelmed by other stuff. So close, and yet so far... I also don't think it helps to add "consume" without defining what that means, since recipients don't eat messages -- they just receive, interpret, and possibly forward them. Anyway, there's stuff here that can be said in Semantics, stuff that doesn't need to be said because the field value does not include leading or trailing whitespace, and other stuff that probably needs to be added to Messaging so that senders do the right thing. I'll add some commits.

reschke:
Just for the record: at the end should be clear that in HTTP 2 and 3, where there's no text-based message format constraining the field value syntax, leading and trailing whitespace is still prohibited and that that still needs to be enforced somehow.

royfielding:
> Just for the record: at the end should be clear that in HTTP 2 and 3, where there's no text-based message format constraining the field value syntax, leading and trailing whitespace is still prohibited and that that still needs to be enforced somehow.

Right, but that is almost entirely enforced by the messaging format, except for this bit about the field value not allowing any leading or trailing whitespace being a requirement of semantics (so that we don't have to include OWS around every field definition).

royfielding (approved):
I have cleaned up the section to more clearly distinguish the semantics of field values from the messaging syntax defined elsewhere, and some other bits where character limitations were described twice and differences from 2616 are no longer relevant. Please have another look.

reschke (approved):
Looks good. The one thing I'm not quite sure about is whether, for messaging formats that technically do allow leading/trailing whitespace, we should requirement not to send it as well...

mnot:
LGTM.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1658,44 +1656,52 @@ Content-Type: text/plain
   <x:ref>field-vchar</x:ref>    = <x:ref>VCHAR</x:ref> / <x:ref>obs-text</x:ref>
 </sourcecode>
 <t>
+   A field value does not include leading or trailing whitespace. When a
+   specific version of HTTP allows such whitespace to appear in a message,
+   a field parsing implementation &MUST; exclude such whitespace prior to
```
martinthomson, line 1661: should this be "field processing" instead of "field parsing"?

martinthomson (commented):
This is very good.  Sorry I wasn't fast enough to review.  One small question though.

draft-ietf-httpbis-semantics-latest.xml
royfielding: Nope, parsing is used when we are talking about parsing the field, whereas processing would be performing the action associated with the semantics of that field.

PR #792, opened 2021-03-04
Remove mid-stream trailers

### LINKED ISSUE #740: WGLC: Mid-stream trailers
[6](https://www.ietf.org/archive/id/draft-ietf-httpbis-semantics-14.html#name-message-abstraction)

> However, multiple trailer sections interleaved with content have only been deployed as frame extensions.

Trailers that come after the body has well-defined semantics (i.e. metadata applicable to the entire body). 

Interleaved, mid-stream trailers are considered and implemented as extensions, and maybe we should delay adding them to the core semantics. 

This is however an important topic for HTTPAPI, IMO.

royfielding:
Fixes #740


reschke (approved):

mnot (approved):

martinthomson (commented):
WFM, thanks.

Issue #771, opened 2021-02-16
artwork -> sourcecode


mnot:
In most of the recent RFCs I've had published, the RPC has a preference for using `<sourcecode>` for example protocol elements, so we should probably switch.

Also, we've settled on using `<sourcecode type="http-message">` for example HTTP messages, and we have tooling to validate them.

reschke:
That sounds like an issue that should be raised with the RPC. Examples in general are not sourcecode.

The aforementioned tooling can be extended to apply to artwork as well.

mnot:
It can, but they have a registry of types, and the appropriate type is `http-message`, not `message/http`.

reschke:
...which you registered yourself, instead of just using the one we had in the source for a decade :-).

Anyway, we can switch the type string if you want; but examples are not sourcecode.

mnot:
We should ask them before submission; they've been doing it very consistently, and if it's going to change, I'd rather we do it.

reschke:
see https://github.com/rfc-format/draft-iab-xml2rfc-v3-bis/issues/195

reschke:
see also https://github.com/rfc-format/draft-iab-xml2rfc-v3-bis/issues/195#issuecomment-782245779

PR #794, opened 2021-03-05
Change HTTP messages to <sourcecode>

mnot:
For #771.


reschke (commented):
I don't think this is a good change.

As long as it is not clear what the right approach is, changing this is premature.

mnot:
It's a draft, for when we get clarity. Based upon the current state of discussion on 7991bis, it seems likely we'll go this way.

reschke:
FWIW, re the argument "minimize the amount of diffs during AUTH48" - we already have ton of unavoidable diffs, due to the extensions we use (for instance, for cross-references between documents).

reschke:
FWIW, removing "x:indent-with" will change the indentation in the text output, creating lot of diffs. Is this intended and ok?

mnot:
I think so; it aligns with other usage.

PR #797, opened 2021-03-05
Clarify that HTTP self-descriptive messages have an exception …

### LINKED ISSUE #700: Messages aren't always self-descriptive

royfielding:
…  in that the request must be understood in order to parse and interpret the response.

Fixes #700

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -2169,7 +2169,13 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
    looking at the message itself, after decoding or reconstituting parts that
    have been compressed or elided in transit, without requiring an
    understanding of the sender's current application state (established via
-   prior messages).
+   prior messages). However, there is one significant exception: each
+   response relies on prior understanding of its corresponding request.
+   A client might need to remember the method, target URI, HTTP version, and
+   content negotiation fields that it used for a given request in order to
+   correctly parse, interpret, or cache the response. For example, responses
+   to the <x:ref>HEAD</x:ref> method look just like the beginning of a
+   response to <x:ref>GET</x:ref>, but cannot be parsed the same.
 </t>
 <t>
    Note that this message abstraction is a generalization across many versions
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -2169,10 +2169,9 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
    looking at the message itself, after decoding or reconstituting parts that
    have been compressed or elided in transit, without requiring an
    understanding of the sender's current application state (established via
-   prior messages). However, a response must be understood within the context
-   of its corresponding request in order to
-   correctly parse, interpret, or cache the response. For example, responses
-   to the <x:ref>HEAD</x:ref> method look just like the beginning of a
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2169,7 +2169,13 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
...
    have been compressed or elided in transit, without requiring an
    understanding of the sender's current application state (established via
-   prior messages).
+   prior messages). However, there is one significant exception: each
+   response relies on prior understanding of its corresponding request.
+   A client might need to remember the method, target URI, HTTP version, and
+   content negotiation fields that it used for a given request in order to
+   correctly parse, interpret, or cache the response. For example, responses
+   to the <x:ref>HEAD</x:ref> method look just like the beginning of a
+   response to <x:ref>GET</x:ref>, but cannot be parsed the same.
```
mnot, line 2178:
```suggestion
   prior messages). However, a response must be understood within the context
   of its corresponding request in order to
   correctly parse, interpret, or cache the response. For example, responses
   to the <x:ref>HEAD</x:ref> method look just like the beginning of a
   response to <x:ref>GET</x:ref>, but cannot be parsed in the same manner.
```

mnot (approved):
This is fine, but can be simplified.

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2169,7 +2169,11 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
    looking at the message itself, after decoding or reconstituting parts that
    have been compressed or elided in transit, without requiring an
    understanding of the sender's current application state (established via
-   prior messages).
+   prior messages). However, a response must be understood within the context
+   of its corresponding request in order to
+   correctly parse, interpret, or cache the response. For example, responses
+   to the <x:ref>HEAD</x:ref> method look just like the beginning of a
+   response to <x:ref>GET</x:ref>, but cannot be parsed in the same manner.
```
royfielding, line 2176:
```suggestion
   prior messages). However, a client &MUST; retain knowledge of the request when
   parsing, interpreting, or caching a corresponding response. For example,
   responses to the <x:ref>HEAD</x:ref> method look just like the beginning of a
   response to <x:ref>GET</x:ref>, but cannot be parsed in the same manner.
```

royfielding:
Well, the simple version is a normative statement in sheep's clothing. Let's just make it normative.

reschke (approved):

mnot (approved):

martinthomson (approved):

PR #802, opened 2021-03-24
clarify how the target URI is extracted from 1.1 request

### LINKED ISSUE #722: Default value for Host
In determining the request target, the following text is included after failing to find an authority in configuration, the authority form target, or the Host header:

> Otherwise, the authority component is assigned the default name configured for the server and, if the connection's incoming TCP port number differs from the default port for the target URI's scheme, then a colon (":") and the incoming port number (in decimal form) are appended to the authority component.

This is a security risk because it means that servers might not agree with clients about the identity of the resource that the request is directed to.  This is especially bad as a TLS certificate is good for every port and often good for multiple names, which can mean a fairly large scope over which an attacker can exploit this confusion.

The origin form description doesn't use 2119 language, it says "A Host header field is also sent", citing [-semantics](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.7.2).  And that section is a bit weaselly in that regard, using a "SHOULD...unless".

> The target URI's authority information is critical for handling a request. A user agent SHOULD generate Host as the first field in the header section of a request unless it has already generated that information as an ":authority" pseudo-header field.

If this were a "MUST ... unless" in -semantics, this problem would not exist (at least in theory).

Just to confuse things, the absolute-form says [this](https://httpwg.org/http-core/draft-ietf-httpbis-messaging-latest.html#rfc.section.3.2.2.p.6), which might be read to imply that Host truly is mandatory:

> A client MUST send a Host header field in an HTTP/1.1 request even if the request-target is in the absolute-form

The asterisk-form is the only other option that naturally lacks an authority, and that says nothing, so that might be fixed up similarly.

royfielding:
... components, depending on the form of request-target, and when authority might be empty.

Fixes #722

This is an alternative to #732

### DIFF
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -3134,6 +3134,7 @@ http://www.example.org:8080
 
 <section title="Since draft-ietf-httpbis-messaging-14" anchor="changes.since.14">
 <ul x:when-empty="None yet.">
+   <li>In <xref target="reconstructing.target.uri"/>, clarify how the target URI is reconstructed when the request-target is not in absolute-form and highlight risk in selecting a default host (<eref target="https://github.com/httpwg/http-core/issues/722"/>)</li>
    <li>In <xref target="chunked.encoding"/>, clarify large chunk handling issues (<eref target="https://github.com/httpwg/http-core/issues/749"/>)</li>
    <li>In <xref target="message.parsing"/>, explicitly close the connection after sending a 400 (<eref target="https://github.com/httpwg/http-core/issues/750"/>)</li>
    <li>In <xref target="http.version"/>, refine version requirements for intermediaries (<eref target="https://github.com/httpwg/http-core/issues/751"/>)</li>
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -761,7 +761,7 @@ https://www.example.org
    request.
 </t>
 <t>
-   However, supplying a default name for authority within the context of a
+   Supplying a default name for authority within the context of a
    secured connection is inherently unsafe if there is any chance that the
    user agent's intended authority (used to verify the secured connection)
    might differ from the selected default. A server that can uniquely
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -754,9 +754,10 @@ https://www.example.org
    If the received authority is empty and the indicated URI scheme
    requires a non-empty authority (as is the case for "http" and "https"),
    the server can reject the request or determine whether a configured
-   default applies that is consistent with the incoming connection's context
-   (scheme, address, port, and locally-defined fields specific to that
-   server's configuration). In the latter case, an empty authority would be
+   default applies that is consistent with the incoming connection's context.
+   Context might include connection details like address and port, what
+   security has been applied, and locally-defined information specific to that
+   server's configuration. An empty authority is
    replaced with the configured default before further processing of the
    request.
 </t>
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -686,49 +686,44 @@ Host: www.example.org:8001
 <section title="Reconstructing the Target URI" anchor="reconstructing.target.uri">
   <x:anchor-alias value="h1.effective.request.uri"/>
 <t>
-   Since the request-target often contains only part of the user agent's
-   target URI, a server constructs its value to properly service the request
-   (<xref target="target.resource"/>).
-</t>
-<t>
    If the <x:ref>request-target</x:ref> is in <x:ref>absolute-form</x:ref>,
    the target URI is the same as the request-target. Otherwise, the
-   target URI is constructed as follows:
+   server reconstructs the target URI from the connection context and various
+   parts of the request message in order to identify the target resource
+   (<xref target="target.resource"/>):
 </t>
 <ul>
-<li>
+<li><t>
    If the server's configuration (or outbound gateway) provides a fixed URI
    scheme, that scheme is used for the target URI.
    Otherwise, if the request is received over a secured connection,
    the target URI's scheme is "https"; if not, the scheme is "http".
-</li>
-<li>
-   If the server's configuration (or outbound gateway) provides a fixed URI
-   <x:ref>authority</x:ref> component, that authority is used for the
-   target URI. If not, then if the request-target is in
-   <x:ref>authority-form</x:ref>, the target URI's authority
-   component is the same as the request-target.
-   If not, then if a <x:ref>Host</x:ref> header field is supplied with a
-   non-empty field value, the authority component is the same as the
-   Host field value. Otherwise, the authority component is assigned
-   the default name configured for the server and, if the connection's
-   incoming TCP port number differs from the default port for the target
-   URI's scheme, then a colon (":") and the incoming port number (in
-   decimal form) are appended to the authority component.
+</t></li>
+<li><t>
+   If the request-target is in <x:ref>authority-form</x:ref>,
+   the target URI's authority component is the same as the request-target.
+   Otherwise, the request-target is in <x:ref>origin-form</x:ref> and:
+</t>
+   <ul>
+   <li><t>if a valid <x:ref>Host</x:ref> header field is supplied with a
+       non-empty field value, the target URI's authority component is the
+       same as that Host field value; else,</t></li>
+   <li><t>the target URI's authority component is empty.</t></li>
+   </ul>
 </li>
 <li>
    If the request-target is in <x:ref>authority-form</x:ref> or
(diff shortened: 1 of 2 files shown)


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -685,50 +685,45 @@ Host: www.example.org:8001
...
-   Host field value. Otherwise, the authority component is assigned
-   the default name configured for the server and, if the connection's
-   incoming TCP port number differs from the default port for the target
-   URI's scheme, then a colon (":") and the incoming port number (in
-   decimal form) are appended to the authority component.
+</t></li>
+<li><t>
+   If the request-target is in <x:ref>authority-form</x:ref>,
+   the target URI's authority component is the same as the request-target.
+   Otherwise, the request-target is in <x:ref>origin-form</x:ref> and:
```
martinthomson, line 705: Where are the asterisk-form clauses for determining authority?

I think that you have a three-fold conditional here, which this presentation doesn't seem to work quite right for:

1. If authority-form, authority is right there.
2. If origin-form AND Host is valid, authority is the value of Host.
3. Otherwise, the authority component is empty.

It looks like you could write that as three sentences without the extra nested list.

Though do you need the "if origin-form" condition on the second item?  Can the same apply to asterisk-form?  That is, condition 2 is "If Host is valid, authority is the value of Host."  The `OPTIONS *` example below suggests that it isn't necessary.

Incidentally, I initially read "Otherwise, the request is in origin-form" as a statement of fact rather than a condition, which was very confusing at first.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -756,10 +751,30 @@ Host: www.example.org
...
+   requires a non-empty authority (as is the case for "http" and "https"),
+   the server can reject the request or determine whether a configured
+   default applies that is consistent with the incoming connection's context
+   (scheme, address, port, and locally-defined fields specific to that
+   server's configuration). In the latter case, an empty authority would be
+   replaced with the configured default before further processing of the
+   request.
+</t>
+<t>
+   However, supplying a default name for authority within the context of a
```
martinthomson, line 764:
```suggestion
   Supplying a default name for authority within the context of a
```

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -756,10 +751,30 @@ Host: www.example.org
...
+   secured connection is inherently unsafe if there is any chance that the
+   user agent's intended authority (used to verify the secured connection)
```
martinthomson, line 766:
```suggestion
   user agent's intended authority
```

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -756,10 +751,30 @@ Host: www.example.org
...
+   might differ from the selected default. A server that can uniquely
+   identify an authority from the request context &MAY; use that identity as
+   a default without this risk. Alternatively, it might be better to redirect
+   the request to a safe resource that explains how to obtain a new client.
```
martinthomson, line 770: I laughed.  It's great advice. Keep it.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -756,10 +751,30 @@ Host: www.example.org
...
-   Recipients of an HTTP/1.0 request that lacks a <x:ref>Host</x:ref> header
-   field might need to use heuristics (e.g., examination of the URI path for
-   something unique to a particular host) in order to guess the
-   target URI's authority component.
+   If the received authority is empty and the indicated URI scheme
+   requires a non-empty authority (as is the case for "http" and "https"),
+   the server can reject the request or determine whether a configured
+   default applies that is consistent with the incoming connection's context
+   (scheme, address, port, and locally-defined fields specific to that
+   server's configuration). In the latter case, an empty authority would be
```
martinthomson, line 759:
```suggestion
   default applies that is consistent with the incoming connection's context.
   Context might include connection details like address and port, what
   security has been applied, and locally-defined information specific to that
   server's configuration. An empty authority is
```

martinthomson (commented):
I really like the text on filling in defaults.  It balances my concerns very nicely.

I struggled with the authority component bullet point here.  I have some suggestions for how it might be improved, but nothing concrete.

draft-ietf-httpbis-messaging-latest.xml
royfielding: You are probably right, but I need sleep in order to think it through again. Will try again tomorrow.

royfielding: Right, done in https://github.com/httpwg/http-core/commit/0cf513d7c8119c3cac1db863d00d24684d37c4ff

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -703,29 +703,22 @@ Host: www.example.org:8001
...
-   Host field value. Otherwise, the authority component is assigned
-   the default name configured for the server and, if the connection's
-   incoming TCP port number differs from the default port for the target
-   URI's scheme, then a colon (":") and the incoming port number (in
-   decimal form) are appended to the authority component.
+   If the request-target is in <x:ref>authority-form</x:ref>,
+   the request-target is the target URI's authority component.
+   Otherwise, the target URI's authority component is the field value of the
+   <x:ref>Host</x:ref> header field, if that value is valid and non-empty, or
+   the authority component is empty if Host is absent, empty, or invalid.
```
martinthomson, line 710: I think that three sentences works better, but our language doesn't do ternary conditions very well.

```suggestion
   Otherwise, the target URI's authority component is the field value of the
   <x:ref>Host</x:ref> header field, if that value is valid and non-empty.
   In all other cases, the authority component is empty if <x:ref>Host<x:ref>
   is absent, empty, or invalid.
```

or even

```suggestion
   For other request forms, if the <x:ref>Host</x:ref> header field is valid
   and non-empty, the target URI's authority component is that value.
   Otherwise, the authority component is empty if <x:ref>Host<x:ref> is
   absent, empty, or invalid.
```

or, to make the construction more parallel, though I don't know how to say "empty is the authority component" without it seeming like a bad Yoda impression


```suggestion
   For other request forms, if the <x:ref>Host</x:ref> header field is valid
   and non-empty, that value is the target URI's authority component.
   Otherwise, the authority component is empty if <x:ref>Host<x:ref> is
   absent, empty, or invalid.
```

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -703,29 +703,22 @@ Host: www.example.org:8001
...
 </li>
 <li>
    If the request-target is in <x:ref>authority-form</x:ref> or
    <x:ref>asterisk-form</x:ref>, the target URI's combined
-   <x:ref>path</x:ref> and <x:ref>query</x:ref> component is empty. Otherwise,
-   the combined <x:ref>path</x:ref> and <x:ref>query</x:ref> component is the
-   same as the request-target.
+   <x:ref>path</x:ref> and <x:ref>query</x:ref> component is empty.
+   Otherwise, the request-target is the target URI's combined
+   <x:ref>path</x:ref> and <x:ref>query</x:ref> component.
```
martinthomson, line 717: I only just noticed this, but there is a difference in treatment between the items in this list.

The authority component uses "the authority component is X", whereas this uses "X is the combined path and query component".  Not sure if that is useful information or whether it just makes this harder to manage, but it does make reading this slightly harder than might be ideal.  (Though this change makes it a lot clearer.)

martinthomson (approved):
Much improved.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -686,14 +686,14 @@ Host: www.example.org:8001
 <section title="Reconstructing the Target URI" anchor="reconstructing.target.uri">
   <x:anchor-alias value="h1.effective.request.uri"/>
 <t>
-   Since the request-target often contains only part of the user agent's
-   target URI, a server constructs its value to properly service the request
-   (<xref target="target.resource"/>).
+   If the <x:ref>request-target</x:ref> is in <x:ref>absolute-form</x:ref>,
+   the request-target is the target URI. A server will parse that URI into
```
reschke, line 690:
```suggestion
   it is the target URI. A server will parse that URI into
```

maybe...

reschke (approved):

draft-ietf-httpbis-messaging-latest.xml
royfielding: Yeah, on third thought it is more consistent to describe it as variable assignment.

draft-ietf-httpbis-messaging-latest.xml
royfielding: it's better not to tempt ambiguity.

draft-ietf-httpbis-messaging-latest.xml
royfielding: I used a fourth option in https://github.com/httpwg/http-core/pull/802/commits/69dbe8ff030a94c9183e9ec023d6c4fb241eb51a

reschke (approved):

martinthomson (approved):
better again

PR #807, opened 2021-03-25
use uri-host:port to define the authority form

### LINKED ISSUE #806: Definition of authority-form in Messaging must require host:port
In 3.2.3. authority-form it says

> The authority-form of request-target is only used for CONNECT requests (Section 9.3.6 of [Semantics]).
> 
>  authority-form = authority

Whereas it needs to specifically require host:port, as in the definition of [CONNECT](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#CONNECT):

> CONNECT uses a special form of request target, unique to this method, consisting of only the host and port number of the tunnel destination, separated by a colon. There is no default port; a client MUST send the port number even if the CONNECT request is based on a URI reference that contains an authority component with an elided port (Section 4.1).

royfielding:
Fixes #806

### DIFF
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -2950,6 +2950,12 @@ https://www.example.org
    conflicts with the use of ranks in the <x:ref>TE</x:ref> header field.
    (<xref target="transfer.coding.registry"/>)
 </t>
+<t>
+   The ABNF definition of <x:ref>authority-form</x:ref> has changed from the
+   more general authority component of a URI (in which port is optional) to
+   the specific host:port format that is required by CONNECT.
+   (<xref target="authority-form"/>)
+</t>
 </section>
 </section>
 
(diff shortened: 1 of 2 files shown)


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -2661,7 +2668,7 @@ https://www.example.org
 <x:ref>absolute-path</x:ref> = &lt;absolute-path, see <xref target="Semantics" x:fmt="," x:sec="4"/>&gt;
 <x:ref>asterisk-form</x:ref> = "*"
 <x:ref>authority</x:ref> = &lt;authority, see <xref target="RFC3986" x:fmt="," x:sec="3.2"/>&gt;
-<x:ref>authority-form</x:ref> = authority
+<x:ref>authority-form</x:ref> = uri-host ":" port
```
reschke, line 2671:
```suggestion
<x:ref>authority-form</x:ref> = <x:ref>uri-host</x:ref> ":" <x:ref>port</x:ref>
```

reschke (commented):
does this merit a change log entry?

mnot (approved):

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -2939,6 +2950,12 @@ https://www.example.org
    conflicts with the use of ranks in the <x:ref>TE</x:ref> header field.
    (<xref target="transfer.coding.registry"/>)
 </t>
+<t>
+   The ABNF definition of <x:ref>authority-form</x:ref> has changed from the
+   more general authority component of a URI (in which port is optional) to
+   the specific host:port format that is required by CONNECT.
+   (<xref target="authority-form"/>)
```
martinthomson, line 2957: It also doesn't allow `userinfo@` :)

PR #810, opened 2021-03-25
Note that Trailer isn't hop-by-hop

### LINKED ISSUE #793: Trailers header field doesn't need to be hop-by-hop
I added "Remove Trailer from existing header fields" for the dechunking algorithm in RFC7230, and at some point we added Trailers to the list of fields to be removed hop-by-hop even when not present in Connection. However, now that preserving trailer fields is optional and separate from header fields, it makes more sense to retain Trailers as a sort of tombstone that indicates "something might be in the trailers" or "something might have been in the trailers but we lost it along the way".

mnot:
Fixes #793

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -2798,7 +2798,6 @@ Host: www.example.org
    <li>Proxy-Connection (<xref target="compatibility.with.http.1.0.persistent.connections"/>)</li>
    <li>Keep-Alive (<xref x:sec="19.7.1" x:fmt="of" target="RFC2068"/>)</li>
    <li>TE (<xref target="field.te"/>)</li>
-   <li>Trailer (<xref target="field.trailer"/>)</li>
    <li>Transfer-Encoding (<xref target="field.transfer-encoding"/>)</li>
    <li>Upgrade (<xref target="field.upgrade"/>)</li>
 </ul>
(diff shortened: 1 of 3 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -13485,6 +13489,7 @@ Content-Type: text/plain
   <li>Move discussion of statelessness from <xref target="intermediaries"/> to <xref target="connections"/> (<eref target="https://github.com/httpwg/http-core/issues/753"/>)</li>
   <li>In <xref target="status.101"/>, clarify that the upgraded protocol is in effect after the 101 response (<eref target="https://github.com/httpwg/http-core/issues/776"/>)</li>
   <li>Throughout, remove unneeded normative references to <xref target="Messaging"/> (<eref target="https://github.com/httpwg/http-core/issues/795"/>)</li>
+  <li>In <xref target="field.trailer"/>, note that Trailers can be used to discover deleted trailers (<eref target="https://github.com/httpwg/http-core/issues/793"/>)</li>
```
royfielding, line 13492:
```suggestion
  <li>In <xref target="field.trailer"/>, note that the Trailer field can be used to discover deleted trailers (<eref target="https://github.com/httpwg/http-core/issues/793"/>)</li>
```

royfielding (changes requested):
We would also want to remove

>   Remove Trailer from existing header fields

from 'Decoding Chunked" in Messaging. Otherwise, LGTM.

reschke (approved):

royfielding (approved):

PR #814, opened 2021-03-26
Clarify edge case of If-Match and If-Unmodified-Since for retrieval requests like GET (fixes #800)

### LINKED ISSUE #800: GET conditioned over If-Match
Consider a GET request conditioned over `If-Match` (I know it's nonsense, but the spec allows it). If the origin server does not have a current representation for the target resource, then the condition is false:
https://github.com/httpwg/http-core/blob/df2fefd530357474ecc7b0ef2f5b324371cdc878/draft-ietf-httpbis-semantics-latest.xml#L6963-L6978
Therefore, the server should respond 412:
https://github.com/httpwg/http-core/blob/df2fefd530357474ecc7b0ef2f5b324371cdc878/draft-ietf-httpbis-semantics-latest.xml#L7450-L7459
However, both Apache and Nginx responded 404.
Did I misunderstand some part of the specification, or are the servers violating the spec?

reschke:


reschke:
Something went wrong here. Will recreate.

PR #816, opened 2021-03-26
Clarify edge case of If-Match and If-Unmodified-Since for retrieval requests like GET (fixes #800)

### LINKED ISSUE #800: GET conditioned over If-Match

reschke:


royfielding (changes requested):
This doesn't look right -- I don't want a 200 response if the precondition fails unless it is specifically If-Modified-Since or If-None-Match. The edge case was about 404, not 200.

royfielding:
And I'll reiterate again that we do not need to specify this edge case. The response should be 404 because there is no selected representation upon which to test the 412.

reschke:
Hm. The spec doesn't seem to allow the 404, though.

I guess we should park this and get the specs out though.

royfielding:
> Hm. The spec doesn't seem to allow the 404, though.

This is defined in "When to Evaluate":

> A server MUST ignore all received preconditions if its response to the same request without those conditions, prior to processing the request content, would have been a status code other than a <x:ref>2xx (Successful)</x:ref> or <x:ref>412 (Precondition Failed)</x:ref>.

IOW, if the request would have been 404 without the precondition then it MUST be 404 with the precondition.

royfielding:
I closed the issue instead.

liyishuai:
Let me check my understanding:
A server MUST first preprocess the request without conditions, and check whether the status code can be determined, before evaluating the conditions.
If so, should we replace all occurrences "precondition" with "condition"? since they're not evaluated at the very first, conceptually.

PR #817, opened 2021-03-26
clarify extensibility of preconditions

### LINKED ISSUE #715: Unknown preconditions aren't safe
[Section 13](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#conditional.requests) implies that a precondition failure is a good safe mode.  However, a precondition that isn't recognized by a server will not share this property.  

> Regardless, if the mapping is inconsistent and the server is unable to select the appropriate representation, then no harm will result when the precondition evaluates to false.

A server that does not recognize a header as a precondition will ignore it and likely proceed with handling the request even if the precondition expressed should have failed.

Indeed, the same is true if a server ignores preconditions, though the specification effectively [mandates that they be supported](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.13.2.2.p.2).

[Section 13.2.2](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#precedence) is the only text I could find on defining new preconditions.

(One thing I really want to avoid is server's applying special heuristics to headers that start with `If-`, because that is one of many poor outcomes if this is left.)

royfielding:
Fixes #715 without changing section numbering

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -13509,6 +13509,7 @@ Content-Type: text/plain
   <li>Simplify description of reactive content negotiation in <xref target="reactive.negotiation"/> (<eref target="https://github.com/httpwg/http-core/issues/712"/>)</li>
   <li>In <xref target="charset"/>, remove the "charset" ABNF production, and clarify where charsets appear (<eref target="https://github.com/httpwg/http-core/issues/713"/>)</li>
   <li>In <xref target="field.accept-encoding"/>, clarify that selection <em>between</em> multiple acceptable codings is only relevant when they have the same purpose (<eref target="https://github.com/httpwg/http-core/issues/714"/>)</li>
+  <li>In <xref target="conditional.requests"/>, rewrite introduction, mentioning extensibility (<eref target="https://github.com/httpwg/http-core/issues/715"/>)</li>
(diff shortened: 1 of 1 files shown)


reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6921,32 +6921,45 @@ Vary: accept-encoding, accept-language
...
 
 <section title="Preconditions" anchor="preconditions">
+   <iref primary="false" item="selected representation"/>
+<t>
+   Preconditions are usually defined with respect to a state of the target
+   resource as a whole (its current value set) or the state as observed in a
+   previously obtained representation (one value in that set). If a resource
+   has multiple current representations, each with its own observable state,
+   a precondition will assume that the mapping of requests to a
+   <x:ref>selected representation</x:ref> (<xref target="representations"/>)
```
mnot, line 6933: The differing pluralisation in 'mapping of requests to a selected representation' tripped me up a bit. Suggest: 

```suggestion
   a precondition will assume that the mapping of requests to their
   <x:ref>selected representations</x:ref> (<xref target="representations"/>)
```

mnot (approved):
One grammatical nit.

draft-ietf-httpbis-semantics-latest.xml
royfielding:
```suggestion
   a precondition will assume that the mapping of each request to a
   <x:ref>selected representation</x:ref> (<xref target="representations"/>)
```

Issue #456, opened 2020-09-29
Reduce use of HTTP in the third person


royfielding:
We talk a lot about HTTP in the third person. For historical context, that's fine, but we don't need to say it when we are defining HTTP itself. So, phrases like

   "HTTP defines messages as being …" :arrow_right: "Messages are …"

This is related to preferring active over passive voice, particularly with regard to targeting requirements on actors instead of stating a requirement as if it happens by magic.

royfielding:
After fixing the passive sentences, the remaining use of HTTP seems to be reasonable (or at least too ingrained to be rewritten at this point). Maybe the RFC editor will go further.

PR #820, opened 2021-03-29
use active instead of passive voice and introduce status code classes before requirements

royfielding:
this is only editorial, but worth reviewing first.


reschke (approved):

royfielding:
Related to #456

mnot (approved):

Issue #827, opened 2021-04-01
Media type of 303 response


ioggstream:
## I wish

To clarify that a 303 response that does not return an hypertext (eg. application/json) is not weird

The spec states that 

> the representation of a 303 response ought to contain a short hypertext note with a hyperlink to the same URI reference provided in the Location header field

imho this implies in some way that returning eg. application/json is not an interoperable behavior.

See https://github.com/httpwg/http-extensions/pull/690

reschke:
No, it doesn't imply non-interoperable behaviour. If the text was concerned with interoperability, it would use a BCP14 -style requirement.

It's more like "it's a good idea to do this".

ioggstream:
In my experience, `ought` is typically used to indicate a correct behavior WRT an incorrect one (eg. implying that if you don't do it, you are not behaving correctly. Maybe some native speaker can help here.

Technically speaking, the fact that the spec references `a short hypertext` implies that 303 is tighly related to web browsers instead of generic user agents (eg. API clients).
WDYT?

royfielding:
It is the same thing we say about all non-success responses, regardless of client type. The only difference here is that a redirection implies a direction be given, which means some form of hypertext. If the JSON has a link in it and some form of processing model to use it, that's a form of hypertext. It is not specific to HTML and, as Julian said, all requirements have requirement language. The "ought" is specifically to emphasize that it is not a requirement.

ioggstream:
>  If the JSON has a link in it and some form of processing model to use it, that's a form of hypertext

Ok, this fixes then!

MikeBishop:
> In my experience, `ought` is typically used to indicate a correct behavior WRT an incorrect one (eg. implying that if you don't do it, you are not behaving correctly. Maybe some native speaker can help here.

I believe this is where we direct you to RFC 6919.....

ioggstream:
> I believe this is where we direct you to RFC 6919.....

Already been there :P https://github.com/httpwg/http-extensions/pull/690#issuecomment-775091066

PR #828, opened 2021-04-01
Fix:  #827. Other media-types are fine for 303 responses

ioggstream:
## This PR

Tries to clarify that it is perfectly interoperable to use other media-types (eg. application/json) in 303 responses.

IMHO the current wording suggests that not returning a short hypertext is not a best practice.

See #827


reschke (changes requested):
I don't think a change is needed.

But, *if* we add something like this it ought to say "user agent", not "browser".

That said: how is the server supposed to know whether the client is a browser?

ioggstream:
short application/json with links is hypertext

reschke:
"If the JSON has a link in it and some form of processing model to use it, that's a form of hypertext."

So yes, JSON *can* be hypertext, but it requires some processing model. *Usually* that requires a more specifc media type then application/json.

PR #833, opened 2021-04-06
fix misleading statement in Acks (fixes #824)

### LINKED ISSUE #824: misleading text in ACKs
This changed "previous authors, editors, and Working Group Chairs" to "current and previous authors"; that's misleading, as it actually does not contain any *current* authors.

_Originally posted by @reschke in https://github.com/httpwg/http-core/issues/821#issuecomment-810778776_

reschke:


royfielding (changes requested):
I think we should just properly edit the acks instead of hacking on this bit leftover from 7230.

Do we want to ack substantial contributors? If so, I will go through the old RFCs and pick out a list of substance. The last example of that is RFC2616 (based on what I wrote for 1945 and 2068). If not, we should ack all of the specs and not the individuals, instead pointing to the real acks in those specs.

royfielding (approved):
This is what I had in mind in terms of bringing back the original acks.

royfielding:
I am seeing a lot of warnings about lines being too long (longer that 72 characters), but they are unrelated to these changes.

mnot (approved):

PR #834, opened 2021-04-06
Abstract: note that we updated RFC 3864 (fixes #829)

### LINKED ISSUE #829: Should the update to RFC3864 be mentioned in the abstract?
Nits on semantics flags this: 

 -- The draft header indicates that this document updates RFC3864, but the
     abstract doesn't seem to mention this, which it should.

reschke:


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -110,6 +110,7 @@
 <t>
    This document obsoletes RFC 2818, RFC 7231, RFC 7232, RFC 7233,
    RFC 7235, RFC 7538, RFC 7615, RFC 7694, and portions of RFC 7230.
+   Furthermore, it updates RFC 3864.
```
royfielding, line 113: Though I still think placing this stuff in the Abstract is a crime against humanity, I would slightly prefer
```suggestion
   This document updates RFC 3864 and
   obsoletes RFC 2818, RFC 7231, RFC 7232, RFC 7233,
   RFC 7235, RFC 7538, RFC 7615, RFC 7694, and portions of RFC 7230.
```

royfielding (changes requested):

draft-ietf-httpbis-semantics-latest.xml
mnot: +1 to that, and the distaste for doing this sort of thing in the abstract.

PR #835, opened 2021-04-06
Align prose about content in HEAD requests with description of GET (fixes #826)

### LINKED ISSUE #826: HEAD with content is inconsistent with GET and DELETE
HEAD says

> A content within a HEAD request message has no defined semantics; sending content in a HEAD request might cause some existing implementations to reject the request.

whereas the new text for GET is

> A client SHOULD NOT generate content in a GET request. Content received in a GET request has no defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request and close the connection because of its potential as a request smuggling attack (Section 11.2 of [Messaging]).

and similarly for DELETE. We should update HEAD as well to be

> A client SHOULD NOT generate content in a HEAD request. Content received in a HEAD request has no defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request and close the connection because of its potential as a request smuggling attack (Section 11.2 of [Messaging]).

reschke:


royfielding (approved):

mnot (approved):

royfielding:
Should we leave these open until after IESG last call, or just apply them now?

mnot:
Good Q. @tfpauly ?

tfpauly:
I think it'd be preferable to not cut another version until we get through the IETF last call. However, if we want to merge this and just keep a note of the changes, I don't see any issue.

reschke:
> I think it'd be preferable to not cut another version until we get through the IETF last call. 

I think it *would* be preferable, just to avoid people raising the same issues again, and also to get the Index into the last call version.

PR #840, opened 2021-04-29
For "HTTP3" add an RFC Editor note to rename to "RFCnnn" before publication (fixes #815)

### LINKED ISSUE #815: SEMANTICS uses named references inconsistently
HTTP3 is a named reference, but the other HTTP RFCs are referenced only by RFC number.  That makes some sense for the older HTTP RFCs which are being obsoleted by this document, but I'd expect at least H2 and H3 to have parity.

reschke:


royfielding:
If we are going to make this change, we might as well globally replace HTTP3 with RFCxxH3, Messaging with RFCxxH1, Semantics with RFCxxHS, and Caching with RFCxxHC. OTOH, that would practically guarantee that our RFCs would be the first to have five digit numbers.

reschke:
> OTOH, that would practically guarantee that our RFCs would be the first to have five digit numbers.

That part I didn't quite get.

royfielding:
> > OTOH, that would practically guarantee that our RFCs would be the first to have five digit numbers.
> 
> That part I didn't quite get.

Murphy's Law. If we carefully craft a reference so that it fits RFC1234 format (so that it doesn't change the paragraph wrapping when the RFC editor does a global replace), then we will be the first to have an RFC12345 spec.

Seriously, I prefer short names. H1, H2, H3, HC (or CACHE), HS (or SEMANTICS), TLS, etc.

martinthomson:
Short mnemonics are best for these things.  We overdid it with QUIC and [QUIC-TRANSPORT] is too much (of course, we're not about to change it now...)

PR #844, opened 2021-05-20
Application handling of field names

### LINKED ISSUE #843: Security considerations: `_` in field names
Many implementations normalise `-` to `_` when passing field names to applications, for historical reasons. This is a vector for smuggling attacks; see [here](https://github.security.telekom.com/2020/05/smuggling-http-headers-through-reverse-proxies.html).

Suggestions:

1. Consistently recommend that `_` not be used in field names.
2. When an implementation normalises `-` to `_`, it MUST remove fields whose names contain `_`; it MAY make them available through an alternative method.
3.

mnot:
Fixes #843

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -10606,7 +10606,7 @@ Content-Type: text/plain
    When an implementation makes HTTP fields available to an application, it
    is often necessary to map field names into an application-specific name
    space. Most commonly, this is done by changing all instances of "-" to
-   "_", and prepending "HTTP_". For example, "Content-Length" might become
+   "_", and prepending "HTTP_", in order to make the field names suitable for use in environment variables.  For example, "Content-Length" might become
    "HTTP_CONTENT_LENGTH".
 </t>
 <t>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -12641,6 +12641,16 @@ Content-Type: text/plain
   <seriesInfo name="RFC" value="3864"/>
 </reference>
 
+<reference anchor="RFC3875">
+  <front>
+    <title>The Common Gateway Interface (CGI) Version 1.1</title>
+    <author initials="D." surname="Robinson" fullname="David Robinson"/>
+    <author initials="K." surname="Coar" fullname="Ken A. L. Coar"/>
+    <date year="2004" month="October"/>
+  </front>
+  <seriesInfo name="RFC" value="3875"/>
+</reference>
+
 <reference anchor="RFC5322">
   <front>
     <title>Internet Message Format</title>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -10016,9 +10016,9 @@ Content-Type: text/plain
    practice some implementations place limits on the characters they accept
    in field-names. To be interoperable, new field names &SHOULD; constrain
    themselves to alphanumeric characters, "-", and ".", and &SHOULD;
-   begin with an alphanumeric character. In particular, the "_" character
-   can be problematic because of how some implementations treat it;
-   see <xref target="underscore.in.fields"/>.
+   begin with an alphanumeric character. For example, the underscore
+   ("_") character can be problematic when passed through non-HTTP
+   gateway interfaces (see <xref target="underscore.in.fields"/>).
 </t>
 <t>
    Field names ought not be prefixed with "X-"; see
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -10611,7 +10611,7 @@ Content-Type: text/plain
 </t>
 <t>
    This practice can lead to security vulnerabilities. For example, if an
-   attacker were to send a header named "Http_Content_Length", a naive
+   attacker were to send a field named "Content_Length", a naive
    application might use that value instead of the intended one. Most often,
    the resulting vulnerability takes the form of request smuggling (<xref
    target="request.smuggling"/>).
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -10599,6 +10601,35 @@ Content-Type: text/plain
 </t>
 </section>
 
+<section title="Application Handling of Field Names" anchor="underscore.in.fields">
+<t>
+   When an implementation makes HTTP fields available to an application, it
+   is often necessary to map field names into an application-specific name
+   space. Most commonly, this is done by changing all instances of "-" to
```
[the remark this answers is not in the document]
reschke: I tripped over that as well. Maybe we should really mention CGI, or systems that map fields to environment variables.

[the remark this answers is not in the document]
mnot: I'm a bit wary of mentioning CGI specifically, because this convention is copied by other systems (eg WSGI), and readers might think that they aren't vulnerable because they're not actually using CGI. 

Mentioning environment vars makes sense. I can also see saying that it's mostly server-side, but again, I don't want to artificially limit it.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -10599,6 +10601,35 @@ Content-Type: text/plain
...
+   "_", and prepending "HTTP_". For example, "Content-Length" might become
+   "HTTP_CONTENT_LENGTH".
```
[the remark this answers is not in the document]
reschke: "for use in identifiers, as needed when mapping to env variables"?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -10599,6 +10601,35 @@ Content-Type: text/plain
...
+</t>
+<t>
+   This practice can lead to security vulnerabilities. For example, if an
+   attacker were to send a header named "Http_Content_Length", a naive
```
reschke, line 10614:
```suggestion
   attacker were to send a field named "Content_Length", a naive
```

reschke (changes requested):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -10016,7 +10016,9 @@ Content-Type: text/plain
    practice some implementations place limits on the characters they accept
    in field-names. To be interoperable, new field names &SHOULD; constrain
    themselves to alphanumeric characters, "-", and ".", and &SHOULD;
-   begin with an alphanumeric character.
+   begin with an alphanumeric character. In particular, the "_" character
+   can be problematic because of how some implementations treat it;
+   see <xref target="underscore.in.fields"/>.
```
royfielding, line 10021:
```suggestion
   begin with an alphanumeric character. For example, the underscore
   ("_") character can be problematic when passed through non-HTTP
   gateway interfaces (see <xref target="underscore.in.fields"/>).
```

royfielding (commented):
I need to clean up the comments to improve the text.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -10599,6 +10601,35 @@ Content-Type: text/plain
...
+</t>
+<t>
+   To mitigate the associated risks, implementations that perform such
+   mapping are advised to handle an incoming field whose name contains "_"
+   carefully. For example, the field might be completely removed, or it might
+   be made available with a different prefix, to distinguish it from fields
+   containing "-".
+</t>
+
+</section>
```
royfielding, line 10631:
```suggestion
<section title="Application Handling of Field Names" anchor="underscore.in.fields">
<t>
   Servers often use non-HTTP gateway interfaces and frameworks to process a received
   request and produce content for the response. For historical reasons related to the
   Common Gateway Interface (CGI), such interfaces often pass received field names as
   external variable names, using a name mapping suitable for environment variables.
</t>
<t>
   For example, the CGI-defined mapping consists of prepending "HTTP_" to
   each name and changing all instances of hyphen ("-") to underscore ("_").
   A received <x:ref>Content-Length</x:ref> field would be passed
   as the variable "HTTP_CONTENT_LENGTH" with a string value matching the
   received field's value. (This historical practice is why
   <xref target="considerations.for.new.field.names"/> discourages the creation
   of new field names that contain an underscore.)
</t>
<t>
   Unfortunately, mapping field names to different interface names can lead to
   security vulnerabilities if the mapping is incomplete or ambiguous. For example,
   if an attacker were to send a field named "Content_Length", a naive interface
   might map that to the same name as the "Content-Length" field, resulting in
   a potential request smuggling vulnerability (<xref target="request.smuggling"/>).
</t>
<t>
   To mitigate the associated risks, implementations that perform such
   mapping are advised to make the mapping unambiguous and complete
   for the full range of potential octets received as a name (including those
   that are discouraged or forbidden by the HTTP grammar).
   For example, a field with an unexpected name character might
   result in the request being blocked, the specific field being removed,
   or the name being passed with a different prefix to distinguish it from
   other fields.
</t>
</section>
```

royfielding (changes requested):
This is my suggested rewrite.

reschke (approved):

royfielding:
section 4.1.8 is better as a reference

   https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.18

though it reminds be that CONTENT_LENGTH is already protected, so our example is wrong. Maybe we should change it to a Content-Transfer-Encoding example.

reschke:
> though it reminds me that CONTENT_LENGTH is already protected

By?

royfielding:
The interface maps the Content-Length header field to CONTENT_LENGTH, not HTTP_CONTENT_LENGTH, and CGI-based applications will always look to the CONTENT_LENGTH value regardless of the presence of HTTP_CONTENT_LENGTH.

royfielding:
Though, if someone were foolish enough to write an HTTP-CGI-HTTP gateway, all bets are off.

royfielding:
I edited my suggestion again to be more specific about CGI and CGI legacy, and to fix the incorrect example of CONTENT_LENGTH. This may require adding an informative reference for RFC3875.

draft-ietf-httpbis-semantics-latest.xml
mnot: Works for me.

Issue #852, opened 2021-06-10
Lars Eggert COMMENT draft-ietf-httpbis-messaging-16


reschke:
- [x] Section 9.1. , paragraph 2, comment:
>    protocols.  Each connection applies to only one transport link.

I'm not sure I understand what is meant by "transport link" and how connections
would "apply" to one.

Editors: https://github.com/httpwg/http-core/pull/891

-------------------------------------------------------------------------------
- [x] Section 9.4. , paragraph 4, comment:
>    consumes server resources.  Furthermore, using multiple connections
>    can cause undesirable side effects in congested networks.

Using larger number of multiple connections can even cause side effects in
otherwise uncongested networks, because their aggregate and initially
synchronized sending behavior can *cause* congestion that would not have been
present if fewer parallel connections had been used.

Editors: https://github.com/httpwg/http-core/pull/891

-------------------------------------------------------------------------------
- [x] Section 13.1. , paragraph 14, comment:
>    [Welch]    Welch, T. A., "A Technique for High-Performance Data
>               Compression", IEEE Computer 17(6), June 1984.

This can become an informative reference, so to not create a DOWNREF, since it's
only used in the description of an IANA codepoint.

Editors: https://github.com/httpwg/http-core/issues/852#issuecomment-858566800

-------------------------------------------------------------------------------
- [X] Section 13.2. , paragraph 12, comment:
>    [RFC7230]  Fielding, R., Ed. and J. F. Reschke, Ed., "Hypertext
>               Transfer Protocol (HTTP/1.1): Message Syntax and Routing",
>               RFC 7230, DOI 10.17487/RFC7230, June 2014,
>               <https://www.rfc-editor.org/info/rfc7230>.

I think it is common practice to normatively cite an RFC that is being
obsoleted.

Editors: https://github.com/httpwg/http-core/issues/852#issuecomment-858567716

-------------------------------------------------------------------------------
All comments below are about very minor potential issues that you may choose to
address in some way - or ignore - as you see fit. Some were flagged by
automated tools (via https://github.com/larseggert/ietf-reviewtool), so there
will likely be some false positives. There is no need to let me know what you
did with these suggestions.

-------------------------------------------------------------------------------
- [X] Section 9.3. , paragraph 3, nit:
~~~
-    Connection header field (Section 7.6.1 of [Semantics]), if any:
+    based on the protocol version and Connection header field in the
~~~

This got garbled, the suggestion was to rephrase to:

[...]

PR #853, opened 2021-06-10
Lars Eggert COMMENT draft-ietf-httpbis-messaging-16  (for #852)

reschke:


royfielding:
It is spelled as Acknowledgments in every RFC. Fix the one inconsistent case instead.

reschke:
> It is spelled as Acknowledgments in every RFC. Fix the one inconsistent case instead.

FWIW, I checked the style guide. Will check again.

royfielding:
Well, FFS, the style guide also says to use either US spelling or British spelling, but not both. It is spelled Acknowledgments in the US and all RFCs prior to that style guide. The RFC editor deliberately changed my text to Acknowledgments the last time this came up. I don't care which way it is spelled.

royfielding (approved):
I think we should commit this now and let the RFC editor decide on spelling.

Issue #854, opened 2021-06-10
Roman Danyliw COMMENT draft-ietf-httpbis-semantics-16


reschke:
Roman Danyliw has entered the following ballot position for
draft-ietf-httpbis-semantics-16: Yes

When responding, please keep the subject line intact and reply to all
email addresses included in the To and CC lines. (Feel free to cut this
introductory paragraph, however.)

Please refer to https://www.ietf.org/iesg/statement/discuss-criteria.html
for more information about DISCUSS and COMMENT positions.

The document, along with other ballot positions, can be found here:
https://datatracker.ietf.org/doc/draft-ietf-httpbis-semantics/

COMMENT:

- [x] ** Section 2.2.  Per “Additional (social) requirements are placed on
implementations …”, what’s a “social” requirement?

Editors: removed in #903 (https://github.com/httpwg/http-core/issues/854#issuecomment-877502247)

-------------------------------------------------------------------------------
- [x] ** Section 4.2.2.

> Resources made available via the "https" scheme have no shared
> identity with the "http" scheme.  They are distinct origins with
>    separate namespaces.  However, an extension to HTTP that is defined
>    to apply to all origins with the same host, such as the Cookie
>    protocol [RFC6265], can allow information set by one service to
>    impact communication with other services within a matching group of
>    host domains.
> 

It would be worth reiterating that it might be risky for extensions to treat
http + https origins on the same host uniformly.

Editors: added text in #903

-------------------------------------------------------------------------------
- [x] ** Section 5.1.  Convention question.  Per “Field names … ought to be
registered within the …” HTTP field name registry, I have a question about the
strength of the recommendation based on the use of the verb “ought” – is that
RECOMMENDED? SHOULD?  Section 8.3.1, 8.3.2, 8.4.1, 8.8.3, etc use a similar
construct.

Editors: we intentionally do not to use normative language for registration requirements.

-------------------------------------------------------------------------------
- [x] ** Section 7.2.  Does this text allow for the possibility for both a Host and
:authority be included?

Editors: it's syntactically possible in H2 and H3. It's up to those specs to describe what that would mean.

-------------------------------------------------------------------------------
- [x] ** Section 9.2.1.  In addition to example of the access logs files filling the
disk, there could be significant CPU load if the target is a script.

Editors: https://github.com/httpwg/http-core/issues/854#issuecomment-862243176

-------------------------------------------------------------------------------
- [x] ** Section 17.1.  The text provides helpful caution by stressing that “… the
[...]

PR #855, opened 2021-06-10
Roman Danyliw COMMENT draft-ietf-httpbis-semantics-16  (for #854)

reschke:


royfielding:
The first is an incorrect change -- "advance configuration" means configured prior to the request, not "advanced".

The second is indeed a typo.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -10408,8 +10408,9 @@ Content-Type: text/plain
    default configuration).
 </t>
 <t>
-   Users need to be aware that intermediaries are no more trustworthy than
-   the people who run them; HTTP itself cannot solve this problem.
+   Users need to be aware that intermediaries are no more trustworthy that
+   the entities that operate them and the policies governing them; HTTP
+   itself cannot solve this problem.
```
royfielding, line 10413:
```suggestion
   Intermediaries are no more trustworthy than the people and policies
   under which they operate; HTTP cannot solve this problem.
```

royfielding (changes requested):
or just delete the last paragraph.

royfielding (approved):

PR #870, opened 2021-06-16
Editorial suggestions from Benjamin Kaduk's IESG review (-semantics)

kaduk:
There's a range of stuff in here, from simple typo fixes to attempts to reword places that I found confusing; the latter may need additional work in order to be correct and acceptable.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -1012,7 +1012,8 @@ Content-Type: text/plain
    Each protocol element in HTTP that allows a URI reference will indicate
    in its ABNF production whether the element allows any form of reference
    (URI-reference), only a URI in absolute form (absolute-URI), only the
-   path and optional query components, or some combination of the above.
+   path and optional query components (partial-URI),
+   or some combination of the above.
    Unless otherwise indicated, URI references are parsed
    relative to the target URI (<xref target="target.resource"/>).
 </t>
@@ -1076,7 +1077,8 @@ Content-Type: text/plain
 <t>
    The origin server for an "http" URI is identified by the
    <x:ref>authority</x:ref> component, which includes a host identifier
-   and optional port number (<xref target="RFC3986" x:fmt="," x:sec="3.2.2"/>).
+   (<xref target="RFC3986" x:fmt="," x:sec="3.2.2"/>)
+   and optional port number (<xref target="RFC3986" x:fmt="," x:sec="3.2.3"/>).
    If the port subcomponent is empty or not given, TCP port 80 (the
    reserved port for WWW services) is the default.
    The origin determines who has the right to respond authoritatively to
@@ -1116,6 +1118,7 @@ Content-Type: text/plain
 <t>
    The origin server for an "https" URI is identified by the
    <x:ref>authority</x:ref> component, which includes a host identifier
+   (<xref target="RFC3986" x:fmt="," x:sec="3.2.2"/>)
    and optional port number (<xref target="RFC3986" x:fmt="," x:sec="3.2.2"/>).
    If the port subcomponent is empty or not given, TCP port 443
    (the reserved port for HTTP over TLS) is the default.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1574,8 +1577,8 @@ Example-Field: Baz
 <t>
    contains two field lines, both with the field name "Example-Field". The
    first field line has a field line value of "Foo, Bar", while the second
-   field line value is "Baz". The field value for "Example-Field" is a list
-   with three members: "Foo", "Bar", and "Baz".
+   field line value is "Baz". The field value for "Example-Field" is the list
+   "Foo, Bar, Baz".
```
kaduk, line 1581: Giving an example of the actual list structure as it appears on the wire seems more useful to me than enumerating the list members individually.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4740,7 +4744,8 @@ Content-Encoding: gzip
 <t>
    An origin server &SHOULD; verify that the PUT representation is
    consistent with any constraints the server has for the target
-   resource that cannot or will not be changed by the PUT.  This is
+   resource, in that the constraints cannot or will not be changed
+   by the PUT.  This is
```
kaduk, line 4748: It feels weird to me for there to be "constraints" that sometimes constrain the acceptable input and sometimes can be modified by the input -- the latter doesn't feel like much of a "constraint".

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6437,8 +6442,9 @@ WWW-Authenticate: Basic realm="simple", Newauth realm="apps",
 <t>
    Most of these header fields, where indicated, define a wildcard value ("*")
    to select unspecified values. If no wildcard is present, values that are
-   not explicitly mentioned in the field are considered unacceptable, except
-   for within <x:ref>Vary</x:ref> where it means the variance is unlimited.
+   not explicitly mentioned in the field are considered unacceptable.
+   Within <x:ref>Vary</x:ref>, the wildcard value means that the variance
+   is unlimited.
```
kaduk, line 6447: Please check this one; I was inferring that the "except" matches up to the "most" in the previous sentence.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8373,7 +8379,7 @@ Content-Range: exampleunit 11.2-14.3/25
    though an origin server &MAY; generate content of zero length.
    If no content is desired, an origin server ought to send
    <x:dfn>204 (No Content)</x:dfn> instead.
-   For CONNECT, no content is allowed because the successful result is a
+   For CONNECT, there is no content because the successful result is a
```
kaduk, line 8382: "no content is allowed" can parse as "the situation where there is no content is a permitted situation" or as "it is forbidden/impossible to have content".  I propose this rewording because this is not an arbitrary restriction but rather an inherent property of how connect works.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8548,7 +8554,8 @@ Content-Range: exampleunit 11.2-14.3/25
    selected representation's complete length.
 </t>
 <t>
-   A sender that generates a 206 response with an <x:ref>If-Range</x:ref>
+   A sender that generates a 206 response for a request with
```
kaduk, line 8557: I did not see anything to suggest that If-Range is usable as a response header, so made some inference here.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -9830,9 +9837,9 @@ Content-Type: text/plain
 </t>
 <t>
    The definition of a new status code ought to specify whether or not it is
-   cacheable. Note that all status codes can be cached if the response they
+   heuristically cacheable. Note that all status codes can be cached if the response they
```
kaduk, line 9840: "heuristically cacheable" is the phrasing used in Section 15.1, so this was just an attempt to increase consistency.

royfielding:
closing so that this can be done in #899

draft-ietf-httpbis-semantics-latest.xml
reschke: In retrospective, this is IMHO incorrect.

PR #871, opened 2021-06-16
reference 'sameness' in URI spec (fixes #863)

### LINKED ISSUE #863: GET "sameness"
> GET is the primary mechanism of information retrieval and the focus of almost all performance optimizations. Hence, when people speak of retrieving some identifiable information via HTTP, they are generally referring to making a GET request. A successful response reflects the quality of "sameness" identified by the target URI. (...)

Need to explain or reference explanation.

reschke:


royfielding (approved):
good, but now I see other problems with that paragraph.

royfielding (approved):
okay, I think this reads better, or we can split them into separate PRs if needed

reschke:
LGTM.

PR #872, opened 2021-06-17
Editorial suggestions from Benjamin Kaduk's IESG review (-messaging)

kaduk:


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -644,6 +645,7 @@ GET http://www.example.org/pub/WWW/TheProject.html HTTP/1.1
 <sourcecode type="http-message">
 CONNECT www.example.com:80 HTTP/1.1
 Host: www.example.com
+
 </sourcecode>
```
kaduk, line 649: The full request message, of course, includes the blank line to end the headers section.
Whether it is worth trying to be pedantically correct about this across all the `<sourcecode>` elements is not entirely clear, though, given that the presentation form of it is unlikely to show the blank line in a meaninful way.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1404,8 +1409,9 @@ Transfer-Encoding: gzip, chunked
   <dd>See <xref target="gzip.coding"/>.</dd>
 </dl>
 <t>
-   The compression codings do not define any parameters. Their presence
-   &SHOULD; be treated as an error.
+   The compression codings do not define any parameters. The presence
+   of parameters with any of these compression codings &SHOULD; be treated
+   as an error.
```
kaduk, line 1414: A similar statement appears for the chunked transfer coding, where "their presence" is disambiguated since only the parameters appear as a plural noun, with only the single coding being discussed.

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -2086,7 +2092,7 @@ Connection: close
 <section title="Security Considerations" anchor="security.considerations">
 <t>
    This section is meant to inform developers, information providers, and
-   users of known security considerations relevant to HTTP message syntax
+   users about known security considerations relevant to HTTP message syntax
```
kaduk, line 2095: "inform ... users of X" can be ambiguous about whether it is "users of X" that are being informed, or that users are being informed of the information X, which in this case leads to a "garden path sentence".

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -2180,7 +2186,8 @@ Connection: close
 <t>
    Extensions to the protocol might be used to mitigate the risk of unwanted
    modification of messages by intermediaries, even when the "https" scheme is
-   used. Integrity might be assured by using hash functions or digital
+   used. Integrity might be assured by using message authentication codes
+   or digital
```
kaduk, line 2190: Hash functions do not provide integrity protection in the face of modification of messages by intermediaries.

royfielding (approved):

PR #874, opened 2021-06-17
Editorial suggestions from Benjamin Kaduk's IESG review (cache)

kaduk:


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -577,9 +577,9 @@
    When a cache receives a request that can be satisfied by a stored response
    that has a <x:ref>Vary</x:ref> header field (<xref target="field.vary"/>),
    it &MUST-NOT; use that response unless all the selecting header fields
-   nominated by the Vary header field match in both the original request
+   nominated by the Vary header field in both the original request
    (i.e., that associated with the stored response), and the presented
-   request.
+   request, match.
```
mnot, line 582: I don't think this is an improvement.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -608,7 +608,7 @@
    there.
 </t>
 <t>
-   A <x:ref>Vary</x:ref> header field value containing a member "*" always fails to match.
+   A request with a <x:ref>Vary</x:ref> header field value containing a member "*" always fails to match.
```
mnot, line 611: Vary is a response header field.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -705,7 +705,7 @@
 </t>
 <t>
    Clients can send the max-age or min-fresh request directives (<xref
-   target="cache-request-directive" />) to constrain or relax freshness
+   target="cache-request-directive" />) to attempt to constrain or relax freshness
```
mnot, line 708: 'to attempt to' doesn't read well. Can we restructure this?

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -753,7 +753,8 @@
 </ul>
 <t>
    Note that this calculation is not vulnerable to clock skew, since all of
-   the information comes from the origin server.
+   the information comes from the origin server, except when the cache must
+   use the time the message was received due to an absent Date header field.
```
mnot, line 757:
```suggestion
   the information comes from the origin server (except when the cache must
   use the time the message was received due to an absent Date header field).
```

mnot (changes requested):

draft-ietf-httpbis-cache-latest.xml
kaduk: I was confused about "match in both", since "match in both X and Y" would imply that there is some other third thing that both are matched against.  I'll try again...

draft-ietf-httpbis-cache-latest.xml
kaduk: Oops.

draft-ietf-httpbis-cache-latest.xml
kaduk: I'll try a rewording.

royfielding:
Thanks for the good comments. This has been replaced by #900 so that we can edit the suggested text.

PR #880, opened 2021-07-02
Explain the "Content-" prefix

### LINKED ISSUE #878: Content and Payload
I'm a little uneasy that the majority of headers that are prefixed with `Content-` are actually scoped to the selected representation.

E.g., `Content-Length`, `Content-Type`, `Content-Encoding`, `Content-Language`. The only registered, current `Content-` headers that actually about the content (using the current definition in Semantics) are `Content-Location` and `Content-Disposition`, AFAICT. Not sure about `Content-Range`; it's defined in terms of the selected representation, but could equally be defined in terms of content, I suspect.

It might help to add a small note to the [content section](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#content) that explains that the `Content-` prefix doesn't necessarily imply that the header is strictly about the content. Also might be worth mentioning in the [registration considerations](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#considerations.for.new.field.names).

mnot:
Fixes #878

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -2335,6 +2335,15 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
    it does not include the chunk lengths, chunked framing syntax, nor the
    trailer fields (<xref target="trailer.fields"/>).
 </t>
+<aside>
+   <t>
+      &Note; Some field names have a "Content-" prefix. This is an informal
+      convention; while some of these fields refer to the content of the
+      message as defined above, others are scoped to the selected representation
+      (<xref target="representations"/>). See the individual field's
+      definition to disambiguate.
+   </t>
+</aside>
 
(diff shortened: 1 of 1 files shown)


reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2335,6 +2335,15 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
    it does not include the chunk lengths, chunked framing syntax, nor the
    trailer fields (<xref target="trailer.fields"/>).
 </t>
+<aside>
+   <t>
+      &Note; Some field names have a "Content-" prefix. This is an informal
+      convention; while some of these fields refer to the content of the
+      message as defined above, others are scoped to the selected representation
```
royfielding, line 2342:
```suggestion
      message, as defined above, others are scoped to the selected representation
```

royfielding (approved):

Issue #857, opened 2021-06-11
Martin Duke COMMENT on draft-ietf-httpbis-semantics-16


mnot:
- [x] (7.6.3) Via

"If a port is not provided, a recipient MAY interpret that as meaning it was
received on the default TCP port, if any, for the received-protocol."

So if received-protocol is "3", it's a UDP port.

If received-protocol is "1" or "1.1", is the default port 80 or 443? IIUC the
scheme isn't included to determine this.

Editors: https://github.com/httpwg/http-core/pull/881

---

- [x] (7.7) Message Transformations

"A proxy that transforms the content of a 200 (OK) response can inform
downstream recipients that a transformation has been applied by changing the
response status code to 203 (Non-Authoritative Information)"

Why not an normative word, instead of "can"?

Editors: https://github.com/httpwg/http-core/issues/857#issuecomment-872025414

---

- [x] (12.5.3) Is it correct that "identity" and having no field value for
Accept-Encoding are synonymous?

Editors: https://github.com/httpwg/http-core/issues/857#issuecomment-872759458

---

- [x] "Servers that fail a request due to an unsupported content coding ought to
respond with a 415 (Unsupported Media Type) status"

Why not s/ought to/SHOULD ?

Editors: https://github.com/httpwg/http-core/issues/857#issuecomment-872939272

---

- [x] (14.3) Why can only origin servers send "Accept-Ranges: bytes"? Why not
intermediaries?

Editors: https://github.com/httpwg/http-core/pull/882

---

- [x] (15.3.7) "A sender that generates a 206 response with an If-Range header
field"... (13.1.5) leads me to believe that only clients can send If-Range. So
how can there be a response with If-Range?

Editors: https://github.com/httpwg/http-core/pull/883

---

- [x] (15.3.7.2) The last instance of THIS_STRING_SEPARATES has a trailing '--'. If
this is intentional, it ought to be explained.

Editors: https://github.com/httpwg/http-core/issues/857#issuecomment-872032533

---

- [x] (16.3.1) says field names SHOULD begin with a letter, but (16.3.2.1) says they
SHOULD begin with "an alphanumeric character". More broadly, the "Field name:"
description in (16.3.1) should probably refer to (16.3.2.1) unless I'm
misunderstanding the scope of these sections.

Editors: https://github.com/httpwg/http-core/pull/890

---

- [x] (17.13) s/TCP behavior/TCP or QUIC behavior

Editors: https://github.com/httpwg/http-core/pull/884

---

[...]

PR #882, opened 2021-07-02
Accept-Ranges is for Origin Servers

mnot:
For #857

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -7868,7 +7868,7 @@ bytes=500-700,601-999
   <x:anchor-alias value="Accept-Ranges"/>
   <x:anchor-alias value="acceptable-ranges"/>
 <t>
-   The "Accept-Ranges" header field allows a server to indicate that it
+   The "Accept-Ranges" header field allows an origin server to indicate that it
    supports range requests for the target resource.
 </t>
 <sourcecode type="abnf7230"><iref primary="true" item="Grammar" subitem="Accept-Ranges"/><iref primary="true" item="Grammar" subitem="acceptable-ranges"/>
@@ -7888,7 +7888,7 @@ Accept-Ranges: bytes
    involved. Range units are defined in <xref target="range.units"/>.
 </t>
 <t>
-   A server that does not support any kind of range request for the target
+   An origin server that does not support any kind of range request for the target
    resource &MAY; send
 </t>
 <sourcecode type="http-message">
@@ -13608,6 +13608,7 @@ Content-Type: text/plain
 
 <section title="Since draft-ietf-httpbis-semantics-16" anchor="changes.since.16">
 <ul x:when-empty="None yet.">
+   <li>In <xref target="field.accept-ranges"/>, clarify that it's only sent by origin servers (<eref target="https://github.com/httpwg/http-core/issues/857"/>)</li>
 </ul>
 </section>
 </section>


mnot:
See, however, [this comment from Roy in 1996](https://web.archive.org/web/20010222161642/http://www.ics.uci.edu/pub/ietf/http/hypermail/1996q2/0615.html):

>The following paragraphs are absolutely wrong and must be deleted:
>
>The Accept-Ranges MUST NOT be added to a response by a proxy (i.e., it
>may only be added by the origin server), MUST NOT be forwarded by a
>proxy that does not support the Range header, and MUST NOT be returned
>to a client whose HTTP version is less than HTTP/1.1.
> 
>  Note: this rule allows a client to be reasonably sure that a
>  subsequent Range request on the entity will not accidentally
>  transfer the entire entity.  However, the subsequent Range request
>  may follow a different request chain, and so there is a small
>  chance that the entire entity will be returned in a response.
>
> and the paragraph below it should begin with "Servers ..." instead
> of "Origin servers ...".
>
> The reason is because the Range header field is interpreted by the
> next inbound server that has the entity cached -- not necessarily
> the origin server -- and thus 14.5 is SEVERELY BROKEN in draft 04a.

Roy, any thoughts? To my mind, a proxy (as opposed to a gateway) adding Accept-Ranges to a response is... surprising. It's conflating origin capabilities with proxy capabilities.

reschke:
Hm. Why would we forbid non-origin servers to implement ranges?

royfielding:
> Roy, any thoughts? To my mind, a proxy (as opposed to a gateway) adding Accept-Ranges to a response is... surprising. It's conflating origin capabilities with proxy capabilities.

I think that was why it was left as is -- I was reading it as a server capability, not an origin capability, and the paragraphs as a limitation on using Range requests. Someone else wanted it to be only an origin header field. So, fine. I am pretty sure implementations ignore it anyway.

mnot:
The problem is that the current text uses both 'server' and 'origin server', so it sends conflicting messages.

reschke:
So why not remove the restriction on "origin servers" consistently?

MikeBishop:
I'm familiar with a caching layer that will request the entire object upstream for caching purposes and then serve range requests out of the cached resource.  It seems entirely reasonable that an intermediary might do something similar.

mnot:
@MikeBishop I'm familiar with that one too -- but it's a gateway, acting as an origin.

reschke:
My understanding is that we allow all servers to process range requests, but only allow origin servers to indicate support (and even that inconsistently: "A server that does not support any kind of range request for the target resource MAY send Accept-Ranges: none to advise the client not to attempt a range request.")

As far as I can tell, this is clearly a mistake, and removing the single instance of "origin" here would fix that.

royfielding:
I think "fixing" this field is hopeless. It was added late in the game, as a form of advertising for a feature that wasn't always supported, and didn't take into consideration all of the reasons for making Range requests. Hence, I think we should go with Mark's clarification that this field is only set by origin servers, but re-emphasize that it is advisory only and that most clients ignore it because they want a 200 response when range is not supported.

reschke:
Hm, me confused. Isn't the right fox just to replace all mentions of "origin server" with just "server"? What am I missing?

royfielding:
Okay, I went back and re-read the full discussion at

  https://lists.w3.org/Archives/Public/ietf-http-wg-old/1996MayAug/0387.html
  https://lists.w3.org/Archives/Public/ietf-http-wg-old/1996MayAug/0411.html

and the final resolution should have matched Lou's comment. Specifically, the purpose of Netscape's implementation of Accept-Ranges was to indicate if any of the servers/caches along the chain supported Range requests.

So, now I am back to agreeing with 1996-me and Julian that the fault is in the spec.

royfielding:
replaced by #902

Issue #869, opened 2021-06-16
Caching and Messaging should reference Semantics Setion 3 (Terminology)


reschke:

mnot:
Are we thinking just something like "See also [ref] for additional terminology" in the introduction?

reschke:
Yes.

PR #889, opened 2021-07-05
reference "Terminology" + minor cleanups (#869)

reschke:


mnot (approved):

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -249,6 +248,15 @@
...
 </section>
 
+<section title="Terminology and Core Concepts" anchor="terminoly">
+<t>
+  See <xref target="Semantics" x:fmt="of" x:rel="#terminology"/> for
+  general terminology and core concepts of HTTP.
+</t>
+</section>
+
+</section>
```
royfielding, line 258: Does this still build? It looks like an extra section closure was added.

reschke: Yes, I changed the nesting of a section.

royfielding:
This change doesn't make any sense from an editorial viewpoint. We don't need a subsection to make reference to a different document. The subsection title would imply that this section is about the terminology of caching, which is defined in the next section (Overview). Note that this is unlike the same section in Messaging (which does define the terminology for HTTP messages).

What I would prefer is to not change the nesting of delta-seconds, avoid adding a new section, and instead add the new reference after the initial sentence in first paragraph of section 2, and then split the paragraph after, so it would read like:
```
2. Overview of Cache Operation

Proper cache operation preserves the semantics of HTTP transfers while reducing the transmission of information already held in the cache. See Section 3 of [Semantics] for the general terminology and core concepts of HTTP.

Although caching is an entirely OPTIONAL feature of HTTP, it can be assumed that reusing a cached response is desirable and that such reuse is the default behavior when no requirement or local configuration prevents it. Therefore, HTTP cache requirements are focused on preventing a cache from either storing a non-reusable response or reusing a stored response inappropriately, rather than mandating that caches always store and reuse particular responses.
```

draft-ietf-httpbis-cache-latest.xml
martinthomson: BTW, the anchor on that section has a typo.

reschke:
Please go ahead with that (note we need the same type of change for Messaging).

reschke:
@royfielding  -  do you still want to get something changed here?

royfielding:
yes, done in https://github.com/httpwg/http-core/commit/6664fa43c300473b45f16ca5967c0b5775995850

PR #892, opened 2021-07-06
Clarify that only selectable responses are updated

### LINKED ISSUE #839: Validation, Last-Modified and scope of updates
[Caching 3.4.3](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#freshening.responses) says:

> The stored response(s) to update are identified by using the first match (if any) of:
[...]
> If the new response contains one or more strong validators (see Section 8.8.1 of [Semantics]), then each of those strong validators identify the selected representation for update. All the stored responses with one of those same strong validators are identified for update. 

[Semantics 8.8.2.2](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#lastmod.comparison) says about `Last-Modified`:

> A Last-Modified time, when used as a validator in a request, is implicitly weak unless it is possible to deduce that it is strong, using the following rules:
[...]
> The validator is being compared by an intermediate cache to the validator stored in its cache entry for the representation, and [t]hat cache entry includes a Date value which is at least one second after the Last-Modified value and the cache has reason to believe that they were generated by the same clock or that there is enough difference between the Last-Modified and Date values to make clock synchronization issues unlikely.

Because we're saying that validation occurs against _all_ responses for a given URI (see #832), this has the effect of allowing a `304` response with a `Last-Modified` header to update *all* of the responses sent for that URI in that second (provided the requirement about `Date` is met) -- which would lead to e.g., `304` response with `Content-Language: English` updating other languages. That probably isn't desirable.

The obvious way to fix this is to avoid promoting `Last-Modified` to a strong validator when it's being used by a cache. However, see #510 -- it's necessary for it to be strong for `If-Range`.

It might also be worth refining caching 3.4.3 to differentiate between entity-tag updates and last-modified updates.

mnot:
Fixes #839.

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1081,7 +1081,7 @@
 </t>
 <t>
    The initial set of stored responses to update are those that could have been selected for
-   that request -- i.e., those that meet the requirements in <xref
+   that request &mdash; i.e., those that meet the requirements in <xref
    target="constructing.responses.from.caches"/>, except the last requirement
    to be fresh, able to be served stale or just validated.
 </t>
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -1075,36 +1075,39 @@
 
 <section title="Freshening Stored Responses upon Validation" anchor="freshening.responses">
 <t>
-   When a cache receives a <x:ref>304 (Not Modified)</x:ref> response and
-   already has one or more stored <x:ref>200 (OK)</x:ref> responses for the
-   applicable request URI, the cache needs to identify which (if any) are to
-   be updated by the new information provided, and then do so.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-cache-latest.xml
```diff
@@ -1075,36 +1075,39 @@
...
+   When a cache receives a <x:ref>304 (Not Modified)</x:ref> response, it needs
+   to identify stored responses that are suitable for updating with the new information
+   provided, and then do so.
 </t>
-<iref item="strong validator" />
 <t>
-   The stored response(s) to update are identified by using the first match
-   (if any) of:
+   The initial set of stored responses to update are those that could have been selected for
+   that request -- i.e., those that meet the requirements in <xref
```
reschke, line 1084:
```suggestion
   that request &mdash; i.e., those that meet the requirements in <xref
```

reschke (approved):

royfielding (approved):

PR #897, opened 2021-07-07
Only use absolute-form if it's a proxy

### LINKED ISSUE #896: Ben Kaduk's DISCUSS on Semantics - absolute URIs as request targets
_From @kaduk -_

Let's discuss whether the currently specified procedures for reconstructing the target URI from a request-target in absolute-form provide adequate security properties, at the origin server. I'm specifically concerned about taking the scheme directly from the request target, i.e., making the distinction between the "http" and "https" schemes. The simple procedure of "take the scheme from the request-target" would seem to allow for the client to cause the server to engage processing for the "https" origin without receiving the protection that https is supposed to provide. (The converse case does not immediately seem to present much risk but is probably worth preventing as well on general principles of retaining consistency.) I don't remember seeing any text that would require the server to validate the scheme from the request-target against the actual properties of the transport (or the configured fixed URI scheme as might be provisioned with a trusted outbound gateway, etc.) While we do reference §7.4 of [Semantics] with a note that reconstructing the target URI is only part of the process of identifying a target resource, that part of [Semantics] does not mention scheme validation as part of rejecting misdirected requests.

Does the origin server need to validate the scheme from an absolute-form request-target? What is the scope of consequences if it fails to do so?

mnot:
Fixes #896.

### DIFF
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -701,13 +701,11 @@ Host: www.example.org:8001
 <section title="Reconstructing the Target URI" anchor="reconstructing.target.uri">
   <x:anchor-alias value="h1.effective.request.uri"/>
 <t>
-   The target URI is the <x:ref>request-target</x:ref> when the
-   request-target is in <x:ref>absolute-form</x:ref>. In that case,
-   a server will parse the URI into its generic components for further
-   evaluation.
+   When a request-target in <x:ref>absolute-form</x:ref> is received by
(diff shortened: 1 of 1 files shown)


reschke (changes requested):
That seems to leave the algorithm for determining the reequest URI undefined if the form is "absolute-form", and the recipient it not a proxy.

mnot:
Ah - I forgot to update that, thanks. Will do when I’m next in front of computer.

royfielding (changes requested):
Absolutely not.

royfielding:
Closing as this is not in scope.

PR #898, opened 2021-07-08
Check target URIs for scheme semantic mismatches

### LINKED ISSUE #896: Ben Kaduk's DISCUSS on Semantics - absolute URIs as request targets

mnot:
Fixes #896.

Note that if we go in this directions, there are some editorial improvements in #897 that we should pull in separately.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -2719,9 +2719,9 @@ Host: www.example.org
    considerations regarding message routing.
 </t>
 <t>
-   Unless the caveat above applies, an origin server &MUST; reject or redirect
+   Unless the caveat above applies, an origin server &MUST; reject
    a request if any scheme-specific requirements are not met.
-   In particular, a request for an "https" resource &MUST; be refused unless it has been
+   In particular, a request for an "https" resource &MUST; be rejected unless it has been
    received over a connection that has been secured via a certificate
    valid for that target URI's origin, as defined by <xref target="https.uri"/>.
 </t>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -13635,6 +13635,7 @@ Content-Type: text/plain
   <li>Align <xref target="fields.registry"/> with <xref target="considerations.for.new.field.names"/> (<eref target="https://github.com/httpwg/http-core/issues/857"/>)</li>
   <li>In <xref target="field.via"/>, don't specify TCP (<eref target="https://github.com/httpwg/http-core/issues/865"/>)</li>
   <li>In <xref target="content"/>, explain the "Content-" prefix (<eref target="https://github.com/httpwg/http-core/issues/878"/>)</li>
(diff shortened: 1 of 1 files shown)


reschke (changes requested):
The new text looks good to me. But... are we introducing a new requirement here? In that case it would need to be mentioned in the Changes section...

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2718,6 +2718,12 @@ Host: www.example.org
    cache. See <xref target="security.considerations"/> for security
    considerations regarding message routing.
 </t>
+<t>
+   In particular, an origin server &MUST; reject a request whose target URI
+   has a scheme that implies it is to be served over a cryptographically
+   secured connection (such as "https") when the connection it was received
+   upon was not secured, unless the caveat above applies.
+</t>
```
royfielding, line 2726:
```suggestion
<t>
   Unless the caveat above applies, an origin server &MUST; reject or redirect
   a request if the target URI's minimum requirements for access to the identified
   resource are not being met by the request context.
   In particular, a request for an "https" resource is illegitimate unless it has been
   received over a connection that has been secured via a certificate
   valid for that target URI's origin, as defined by <xref target="https.uri"/>.
</t>
```

royfielding (changes requested):
I think this version is more accurate (and more general to other schemes).

wtarreau:
I find Roy's proposal more precise. The overall text does slightly help me find how to figure a scheme in H1.

draft-ietf-httpbis-semantics-latest.xml
mnot: I don't mind that approach, but:

* This phrasing is too obfuscated: "... the target URI's minimum requirements for access to the identified resource are not being met by the request context"

Suggestion: "... any scheme-specific requirements are not met."

* "illegitimate" isn't used anywhere else, and it's unclear what it means here. Suggest "MUST be refused" or "will be refused" if you want to avoid the reinforcing requirement.

reschke (approved):

royfielding (approved):

PR #899, opened 2021-07-11
forked suggestions by kaduk with further edits

royfielding:
Replaces #870 by @kaduk (minus the bits already committed) with adjustments by me.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -1012,7 +1012,8 @@ Content-Type: text/plain
    Each protocol element in HTTP that allows a URI reference will indicate
    in its ABNF production whether the element allows any form of reference
    (URI-reference), only a URI in absolute form (absolute-URI), only the
-   path and optional query components, or some combination of the above.
+   path and optional query components (partial-URI),
+   or some combination of the above.
    Unless otherwise indicated, URI references are parsed
    relative to the target URI (<xref target="target.resource"/>).
 </t>
@@ -1076,7 +1077,8 @@ Content-Type: text/plain
 <t>
    The origin server for an "http" URI is identified by the
    <x:ref>authority</x:ref> component, which includes a host identifier
-   and optional port number (<xref target="RFC3986" x:fmt="," x:sec="3.2.2"/>).
+   (<xref target="RFC3986" x:fmt="," x:sec="3.2.2"/>)
+   and optional port number (<xref target="RFC3986" x:fmt="," x:sec="3.2.3"/>).
    If the port subcomponent is empty or not given, TCP port 80 (the
    reserved port for WWW services) is the default.
    The origin determines who has the right to respond authoritatively to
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1116,6 +1118,7 @@ Content-Type: text/plain
 <t>
    The origin server for an "https" URI is identified by the
    <x:ref>authority</x:ref> component, which includes a host identifier
+   (<xref target="RFC3986" x:fmt="," x:sec="3.2.2"/>)
    and optional port number (<xref target="RFC3986" x:fmt="," x:sec="3.2.2"/>).
```
reschke, line 1122:
```suggestion
   and optional port number (<xref target="RFC3986" x:fmt="," x:sec="3.2.3"/>).
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1745,7 +1748,8 @@ Example-Dates: "Sat, 04 May 1996", "Wed, 14 Sep 2005"
    that allows both unquoted (token) and quoted (quoted-string) syntax for
    a parameter value (<xref target="parameter"/>). Use of common syntax
    allows recipients to reuse existing parser components. When allowing both
-   forms, the meaning of a parameter value ought to be the same whether it
+   forms, the meaning of a parameter value (with no internal commas)
```
reschke, line 1751: What is the "(with no internal commas)" about?

reschke (changes requested):

draft-ietf-httpbis-semantics-latest.xml
royfielding: no idea, will remove this part

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1745,7 +1748,8 @@ Example-Dates: "Sat, 04 May 1996", "Wed, 14 Sep 2005"
...
+   ought to be the same whether it
```
royfielding, line 1752:
```suggestion
   forms, the meaning of a parameter value ought to be the same whether it
```

draft-ietf-httpbis-semantics-latest.xml
kaduk: You can't use the token syntax if there are commas in the parameter value.  If there are such "internal commas" in the parameter value, then only the quoted-string syntax can be used for that particular value, regardless of whether or not the parameter in question allows both token and quoted-string form.
So this statement about "meaning ought to be the same whether it was received as [...]" is nonsensical for a parameter value that has internal commas.

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
reschke: No, the statement is true for syntactically valid tokens and quoted-strings.

mnot (approved):

PR #900, opened 2021-07-11
(editorial) tweaks inspired by kaduk suggestions from IESG review

royfielding:
replaces #874 with my own edits.

Note that I added "without revalidation" to the description of Vary since the cache can use the stored response, regardless of Vary, once it has been revalidated by the origin server (e.g., 304 with etag).


reschke (approved):

mnot (approved):
First changeset is pretty wordy...

PR #902, opened 2021-07-12
clarify that Accept-Ranges can be sent by any server

royfielding:
Clarify that Accept-Ranges can be sent by any server, remove "none" from the ABNF because it is now a reserved range unit, and allow the field to be sent in a trailer section while noting why that is much less useful than as a header field

For #857 

this is an alternative to #882

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -7911,6 +7911,11 @@ Accept-Ranges: bytes
    improving performance and reducing unnecessary network transfers.
 </t>
 <t>
+   Conversely, a client &MUST-NOT; assume that receiving an Accept-Ranges field means that future requests
+   will return partial responses. The server might not support future range requests, or the header might have
+   been added by an intermediary on a different path.
+</t>
+<t>
    A server that does not support any kind of range request for the target
    resource &MAY; send
 </t>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -7885,24 +7885,29 @@ bytes=500-700,601-999
   <x:anchor-alias value="Accept-Ranges"/>
   <x:anchor-alias value="acceptable-ranges"/>
 <t>
-   The "Accept-Ranges" header field allows a server to indicate that it
-   supports range requests for the target resource.
+   The "Accept-Ranges" field in a response indicates whether an upstream
+   server supports range requests for the target resource.
 </t>
 <sourcecode type="abnf7230"><iref primary="true" item="Grammar" subitem="Accept-Ranges"/><iref primary="true" item="Grammar" subitem="acceptable-ranges"/>
   <x:ref>Accept-Ranges</x:ref>     = <x:ref>acceptable-ranges</x:ref>
-  <x:ref>acceptable-ranges</x:ref> = 1#<x:ref>range-unit</x:ref> / "none"
+  <x:ref>acceptable-ranges</x:ref> = 1#<x:ref>range-unit</x:ref>
 </sourcecode>
 <t>
-   An origin server that supports byte-range requests for a given target
-   resource &MAY; send
+   For example, a server that supports
+   <xref target="byte.ranges">byte-range requests</xref> can send the field
 </t>
 <sourcecode type="http-message">
 Accept-Ranges: bytes
 </sourcecode>
 <t>
-   to indicate what range units are supported. A client &MAY; generate range
-   requests without having received this header field for the resource
-   involved. Range units are defined in <xref target="range.units"/>.
+   to indicate that the request's target resource supports range requests,
+   thereby encouraging its use by the client for future partial requests.
+   Range units are defined in <xref target="range.units"/>.
+</t>
+<t>
+   A client &MAY; generate range requests regardless of having received an
+   Accept-Ranges field. The information only provides advice for the sake of
+   improving performance and reducing unnecessary network transfers.
 </t>
 <t>
    A server that does not support any kind of range request for the target
@@ -7912,7 +7917,14 @@ Accept-Ranges: bytes
 Accept-Ranges: none
 </sourcecode>
 <t>
-   to advise the client not to attempt a range request.
+   to advise the client not to attempt a range request. The range unit "none"
+   is reserved for this purpose.
+</t>
+<t>
+   The Accept-Ranges field &MAY; be sent in a trailer section, but is preferred
+   to be sent as a header field because the information is particularly useful
+   for restarting large information transfers that have failed in mid-content
+   (before the trailer section is received).
 </t>
 </section>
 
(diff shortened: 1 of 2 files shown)


reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7885,24 +7885,29 @@ bytes=500-700,601-999
...
 <sourcecode type="http-message">
 Accept-Ranges: bytes
 </sourcecode>
 <t>
-   to indicate what range units are supported. A client &MAY; generate range
-   requests without having received this header field for the resource
-   involved. Range units are defined in <xref target="range.units"/>.
+   to indicate that the request's target resource supports range requests,
+   thereby encouraging its use by the client for future partial requests.
+   Range units are defined in <xref target="range.units"/>.
```
mnot, line 7905: This isn't necessarily true; the target resource doesn't always support them if the header was added by a proxy.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7912,7 +7917,14 @@ Accept-Ranges: bytes
 Accept-Ranges: none
 </sourcecode>
 <t>
-   to advise the client not to attempt a range request.
+   to advise the client not to attempt a range request. The range unit "none"
+   is reserved for this purpose.
```
mnot, line 7921: You need to do something like this to properly qualify it for the possibility that a proxy added the value:

```suggestion
   to advise the client not to attempt a range request on the current request path. The range unit "none"
   is reserved for this purpose.
```

mnot (changes requested):
If we're going to do this, we need to properly account for the cases where a value has been added by a proxy, and then the client's proxy changes (either through configuration or behind the scenes).

I believe most people think of this as metadata about the resource's capabilities -- indeed, the text you just wrote make that assumption. It would be much cleaner and more useful to restrict this information to the origin server, no matter what the historical intent was. That's how it's currently used today.

If we can't agree on this, we should take it to the list.

reschke:
> It would be much cleaner and more useful to restrict this information to the origin server, no matter what the historical intent was. That's how it's currently used today.

IMHO... allowing non-origins to service range requests, but disallowing them to advertise that is really really weird. At the end of the day, any "Accept-*" field value received is always a statement about the present, and not a promise for future requests. This is not different from other fields.

Do you believe it would actually be harmful if a client was mislead to believe it can do a range request, and then it receives the full content with a 200?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7885,24 +7885,29 @@ bytes=500-700,601-999
...
+   <xref target="byte.ranges">byte-range requests</xref> can send the field
 </t>
 <sourcecode type="http-message">
 Accept-Ranges: bytes
 </sourcecode>
 <t>
-   to indicate what range units are supported. A client &MAY; generate range
-   requests without having received this header field for the resource
-   involved. Range units are defined in <xref target="range.units"/>.
+   to indicate that the request's target resource supports range requests,
```
royfielding, line 7903:
```suggestion
   to indicate that it supports byte range requests for that target resource,
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7885,24 +7885,29 @@ bytes=500-700,601-999
...
 </t>
 <sourcecode type="http-message">
 Accept-Ranges: bytes
 </sourcecode>
 <t>
-   to indicate what range units are supported. A client &MAY; generate range
-   requests without having received this header field for the resource
-   involved. Range units are defined in <xref target="range.units"/>.
+   to indicate that it supports byte range requests for that target resource,
+   thereby encouraging its use by the client for future partial requests.
```
royfielding, line 7904:
```suggestion
   thereby encouraging its use by the client for future partial requests on
   the same request path.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7912,7 +7917,14 @@ Accept-Ranges: bytes
 Accept-Ranges: none
 </sourcecode>
 <t>
-   to advise the client not to attempt a range request.
+   to advise the client not to attempt a range request on the current request path. The range unit "none"
+   is reserved for this purpose.
```
royfielding, line 7922:
```suggestion
   to advise the client not to attempt a range request on the same request path.
   The range unit "none" is reserved for this purpose.
```

royfielding:
> I believe most people think of this as metadata about the resource's capabilities -- indeed, the text you just wrote make that assumption. It would be much cleaner and more useful to restrict this information to the origin server, no matter what the historical intent was. That's how it's currently used today.

A header field is defined by the intent of the people who minted and deployed it. That was Lou Montulli. Its original meaning was confused in the pre-RFC2068 draft because of an editorial mistake, not because the header field ever changed its meaning as deployed. The main use case was restarting a big download, and that is not specific to origin servers.

Today, this field is largely ignored because of the poor specification. We can improve such things without harm.

mnot:
I don't think this is worth too much more time, I'll just say I disagree -- it's not very useful in this form, and it differs from every other bit of similar metadata we put into headers (e.g., `Allow`, `Accept-*`, `Vary`).

I'll make a suggestion to try to make the ambiguity here a bit more clear.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7885,24 +7885,29 @@ bytes=500-700,601-999
...
+   to indicate that it supports byte range requests for that target resource,
+   thereby encouraging its use by the client for future partial requests.
+   Range units are defined in <xref target="range.units"/>.
+</t>
+<t>
+   A client &MAY; generate range requests regardless of having received an
+   Accept-Ranges field. The information only provides advice for the sake of
+   improving performance and reducing unnecessary network transfers.
 </t>
 <t>
```
mnot, line 7913:
```suggestion
</t>
<t>
   Conversely, a client &MUST-NOT; assume that receiving an Accept-Ranges field means that future requests
   will return partial responses. The server might not support future range requests, or the header might have
   been added by an intermediary on a different path.
</t>
<t>
```

royfielding:
> I don't think this is worth too much more time, I'll just say I disagree -- it's not very useful in this form, and it differs from every other bit of similar metadata we put into headers (e.g., `Allow`, `Accept-*`, `Vary`).

You are starting with an assumption that this is an origin characteristic and saying that this doesn't fit that assumption. I agree that the assumption does not fit the header field's definition, nor is it a fit for purpose. The field's purpose is to tell the client that range requests may help if there is a problem receiving the whole content. That's not limited to origin servers and has nothing in common with Allow or Vary.

If a proxy cache can process range requests on large cached items (e.g., a company training video) and its clients can benefit from that, why would we not want to advertise that fact to the client? The theory that this may cause a problem with multiple paths just doesn't hold for proxies (they are client chosen) and wouldn't matter anyway for range requests because the worst case is a 200 response. This was explained in 1996 and the definition was fixed in the next draft. It just wasn't fixed enough.

This doesn't change the fact that the header field isn't very useful, simply because it is advisory only and most clients will ignore that advice when it doesn't matter to them if 200 is the response. That's okay.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7885,24 +7885,35 @@ bytes=500-700,601-999
...
+</t>
+<t>
+   A client &MAY; generate range requests regardless of having received an
+   Accept-Ranges field. The information only provides advice for the sake of
+   improving performance and reducing unnecessary network transfers.
+</t>
+<t>
+   Conversely, a client &MUST-NOT; assume that receiving an Accept-Ranges field means that future requests
+   will return partial responses. The server might not support future range requests, or the header might have
+   been added by an intermediary on a different path.
```
royfielding, line 7916:
```suggestion
   Conversely, a client &MUST-NOT; assume that receiving an Accept-Ranges field
   means that future range requests will return partial responses. The content might
   change, the server might only support range requests at certain times or under
   certain conditions, or a different intermediary might process the next request.
```

reschke (approved):

mnot (approved):

PR #906, opened 2021-07-22
without changing the protocol, be explicit about when non-standard content in requests …

### LINKED ISSUE #904: Content on GET: requirement strength

royfielding:
… might be allowed by consenting adults

Fixes #904


martinthomson (approved):
If this is where we end up, I can live with that.

FWIW, this "by prior agreement" is still the very definition of not-HTTP (or basically not a standardized protocol), but if this is what gets this well-overdue document published, then "SHOULD NOT" is what it shall be.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -4621,12 +4621,15 @@ Content-Encoding: gzip
...
-   request. Content received in a GET request has no defined semantics,
+   Content received in a GET request has no generally defined semantics,
    cannot alter the meaning or target of the request, and might lead some
    implementations to reject the request and close the connection because of
    its potential as a request smuggling attack
    (<xref target="request.smuggling"/>).
+   A client &SHOULD-NOT; generate content in a GET request unless it is
+   made directly to an origin server that has previously indicated,
+   in or out of band, that such a request has a purpose and will be adequately
+   supported.
```
mnot, line 4632: This places a requirement on clients and what they generate, but not servers (and especially resources) regarding what they require. As discussed, the latter are often the real problem. So, suggest:

```suggestion
   A client &SHOULD-NOT; generate content in a GET request unless it is
   made directly to an origin server that has previously indicated,
   in or out of band, that such a request has a purpose and will be adequately
   supported. Because some clients might use proxies or other infrastructure
   outside the control of the server, it is NOT RECOMMENDED that resources
   require the use of content on a GET request to interoperate.
```

... with corresponding edits for `HEAD` and `DELETE`.

mnot (changes requested):

draft-ietf-httpbis-semantics-latest.xml
martinthomson: A double-SHOULD NOT is an improvement.

"resources &SHOULD-NOT; require the use of content" perhaps, just to keep this parallel.

reschke: By all means, do not include DELETE into this issue right now.

mnot: @reschke if we want to finish this, it'd really help to know _why_ you object.

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
reschke: I'm ok with this, although I don't think it's needed. If we do this, I would prefer "SHOULD NOT", as this would be the first occurrence of "NOT RECOMMENDED" in the spec.

royfielding: We don't make requirements on resources. It would make more sense to say that an origin server cannot rely on private agreements unless they control the entire request/response chain.

royfielding: https://github.com/httpwg/http-core/pull/906/commits/fbdad50aa9bb577b13518a610c12a2fa36d4f0f1

reschke (approved):

mnot (approved):

PR #907, opened 2021-07-23
switch to named references instead of RFC numbers (for HTTP family)

### LINKED ISSUE #846: anchor names for sibling specs
We currently have [Semantics], [Caching] etc.

Should we...:

1. Stick with those?
2. Use RFCnnnn notation (consistent with what we did in the past)?
3. Something else?

Also: if we stick with this style, should we rename "Messaging" to "HTTP11"?

royfielding:
Fixes #846

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -2161,6 +2161,8 @@ Expires: Thu, 01 Dec 1994 16:00:00 GMT
 </middle>
 
 <back>
+<displayreference target="HTTP11" to="HTTP/1.1"/>
+
 <references title="Normative References">
 
   <reference anchor="HTTP11">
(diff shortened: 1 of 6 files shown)


reschke (changes requested):
We can rename HTTP11 to HTTP/1.1 (and so on); see https://greenbytes.de/tech/webdav/rfc7991.html#element.displayreference; should I do that?

mnot:
I prefer HTTP11. It's not a protocol element, and using the same reference anchor as other specs is good.

reschke:
Doesn't compute :-) What's the relation to protocol elements? Re consistency: we can fix this in unpublished specs.

reschke:
I have added the displayreference elements as suggested by me. @royfielding - please take a look and either keep the change ore revert it if you feel like it. @mnot - if we go with this there's no problem to make the HTTP/3 spec consistent during AUTH48.

reschke (approved):

PR #909, opened 2021-07-25
add change note for #905

reschke:


royfielding (changes requested):
But this should be in HTTP/1.1 messaging, right? Not Semantics.

https://github.com/httpwg/http-core/commit/e77d794a4b2c43a36dae9abc651bbdc13597282a#diff-a83bf4936c8fbe0cf0ad1177a283dd11cf6a02d6cd4321bfa7c08a56c47d684c

reschke:
Indeed.

PR #929, opened 2021-08-03
move RFC5322 to normative references

### LINKED ISSUE #924: RFC 5322 is not cited normatively, yet used in the ABNF
From https://github.com/httpwg/http-core/issues/914:

> Should RFC 5322 be normative? We rely on it for, e.g., the "mailbox" ABNF construction.

royfielding:
fixes #924


mnot (approved):
Ugh. Sure.

reschke (approved):

reschke:
Note that we made it informative so we don't have a downref when we go to full standard...

PR #936, opened 2021-08-06
move Date and Trailer to a new (last) subsection of Message Abstraction

### LINKED ISSUE #934: Trailer and Date are bidirectional fields
>>    Section 10.2

> It seems like it might be worth listing the fields already defined in the previous section (as request context fields) that can also appear as response context fields.

Oops, that wasn't intentional. The Trailer field is the main one that goes both directions. It should really be in a different section. Date is the other.

royfielding:
Fixes #934 

This works well on multiple levels because these fields are not about the request or response context, but rather about the message itself. Date is now defined early and close to the date value definitions of 5.6.7. Trailer is now defined just after the description of trailers. Also, it avoids changing any xrefs from HTTP/3.

Only the section heading and intro are new -- the moved text has not been changed.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -2560,6 +2560,101 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
 </t>
 </section>
 </section>
+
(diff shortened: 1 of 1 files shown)


mnot (approved):
Looks good, want to see it in situ tho

PR #939, opened 2021-08-09
Rephrase description of Vary to be more explanatory

### LINKED ISSUE #938: Improve phrasing of Vary description
As mentioned in #914 ...

>Section 12.5.5
>
>>    An origin server SHOULD send a Vary header field when its algorithm
>>    for selecting a representation varies based on aspects of the request
>>    message other than the method and target URI, unless the variance
>>    cannot be crossed or the origin server has been deliberately
>>    configured to prevent cache transparency. [...]
>
> I don't think I know what it means to "cross" a variance. The example (elided from this comment) about Authorization not needing to be included gives some hint as to what is meant, but I still don't have a clear picture.

royfielding:
Fixes #938 

Note that this also rephrases the SHOULD requirement, but doesn't change its effect.


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6897,72 +6897,94 @@ Accept-Language: da, en-gb;q=0.8, en;q=0.7
    <x:anchor-alias value="Vary"/>
 <t>
    The "Vary" header field in a response describes what parts of a request
-   message, aside from the method and target URI,
-   might influence the origin server's process for selecting and representing
-   this response.
+   message, aside from the method and target URI, might influence the
```
mnot, line 6900:
```suggestion
   message, aside from the method and target URI, might have influenced the
```

Putting this in the past tense relative to the interpretation of the response makes it easier to understand what's going on.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6897,72 +6897,94 @@ Accept-Language: da, en-gb;q=0.8, en;q=0.7
...
 <sourcecode type="abnf7230"><iref primary="true" item="Grammar" subitem="Vary"/>
   <x:ref>Vary</x:ref> = #( "*" / <x:ref>field-name</x:ref> )
 </sourcecode>
 <t>
-   A Vary field value is a list of request field names, known as the selecting header fields,
-   that might have a role in selecting the representation for this response. Potential selecting
-   header fields are not limited to those defined by this specification.
+   A Vary field value is either the wildcard member "*" or a list of
+   request field names, known as the selecting header fields, that might
+   have a role in selecting the representation for this response.
```
mnot, line 6909:
```suggestion
   have had a role in selecting the representation for this response.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6897,72 +6897,94 @@ Accept-Language: da, en-gb;q=0.8, en;q=0.7
...
+   Potential selecting header fields are not limited to those defined by
+   this specification.
 </t>
 <t>
-   If the list contains "*", it signals that other aspects of the request might
-   play a role in selecting the response representation, possibly including
-   elements outside the message syntax (e.g., the client's network address).
+   A list containing the member "*" signals that other aspects of the
+   request might play a role in selecting the response representation,
```
mnot, line 6915:
```suggestion
   request might have played a role in selecting the response representation,
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6897,72 +6897,94 @@ Accept-Language: da, en-gb;q=0.8, en;q=0.7
...
-   cache transparency. For example, there is no need to send the Authorization
-   field name in Vary because reuse across users is constrained by the field
-   definition (<xref target="field.authorization"/>). Likewise, an origin server might use
-   Cache-Control response directives (<xref target="field.cache-control"/>) to supplant Vary if it
-   considers the variance less significant than the performance cost of Vary's
-   impact on caching.
+   An origin server &SHOULD; send a Vary header field when doing so might
+   improve the recipient's content selection in a later request, unless
+   the resource has been deliberately configured to prefer cache reuse
+   regardless of changes to content negotiation preferences.
```
mnot, line 6963:
```suggestion
   An origin server &SHOULD; generate a Vary header on a cacheable response
   when it wishes that response to be selectively reused for subsequent requests.
   Generally, that is the case when the selecting header fields have been used to 
   modify the response in some fashion; for example, <x:ref>Accept-Language</x:ref>
   might have been used to generate a response that would be inappropriate for
   requests from clients with different preferences.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6897,72 +6897,94 @@ Accept-Language: da, en-gb;q=0.8, en;q=0.7
...
+   Vary is intended to prevent (until validated by the origin server)
+   a cached response from being reused to satisfy a subsequent request
+   that has different preferences in the selecting header fields.
+   However, this is a choice made by the origin server and communicated to
+   recipients, not an interoperability requirement whenever content
+   selection is present. Vary might be elided when an origin server
+   considers the variance in content selection to be less significant than
+   Vary's performance impact on caching, particularly when such variance
+   is already being limited by Cache-Control response directives
+   (<xref target="field.cache-control"/>).
```
mnot, line 6975:
```suggestion
   Vary might be elided when an origin server
   considers variance in content selection to be less significant than
   Vary's performance impact on caching.
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6897,72 +6897,94 @@ Accept-Language: da, en-gb;q=0.8, en;q=0.7
...
+   Cached content selection is only improved by Vary if there is some
+   chance that the origin server's selection algorithm might prefer a
+   different representation for the recipient in a subsequent request.
+   For example, there is no need to send the Authorization field name in
+   Vary because reuse of that response for a different user is prohibited
+   by the field definition (<xref target="field.authorization"/>).
+   Likewise, if the response content has been selected by network region
+   but the origin server wants the cached response to be reused even if
+   recipients move from one region to another, then there is no need for
+   the origin server to indicate such variance in Vary.
```
mnot, line 6987:
```suggestion
   There is no need to send the Authorization field name in
   Vary because reuse of that response for a different user is prohibited
   by the field definition (<xref target="field.authorization"/>).
   Likewise, if the response content has been selected or influenced by
   by network region but the origin server wants the cached response to 
   be reused even if recipients move from one region to another, then 
   there is no need for the origin server to indicate such variance in Vary.
```

mnot (changes requested):
I've tried to adjust; this text downplays the impact of not sending Vary -- in my experience, developers much, much more often forget to add it when it's required to reflect their intent; I very rarely see it added unnecessarily.

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
royfielding: So, you are advocating removal of all the explanations and the parts that are not cache-specific and the forward references to Caching? How is this addressing the issue raised?

draft-ietf-httpbis-semantics-latest.xml
royfielding: This suggestion is counter-productive.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6897,72 +6897,96 @@ Accept-Language: da, en-gb;q=0.8, en;q=0.7
...
   <x:ref>Vary</x:ref> = #( "*" / <x:ref>field-name</x:ref> )
 </sourcecode>
 <t>
-   A Vary field value is a list of request field names, known as the selecting header fields,
-   that might have a role in selecting the representation for this response. Potential selecting
-   header fields are not limited to those defined by this specification.
+   A Vary field value is either the wildcard member "*" or a list of
+   request field names, known as the selecting header fields, that might
+   have had a role in selecting the representation for this response.
+   Potential selecting header fields are not limited to those defined by
```
royfielding, line 6910:
```suggestion
   Potential selecting header fields are not limited to fields defined by
```

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6897,72 +6897,96 @@ Accept-Language: da, en-gb;q=0.8, en;q=0.7
...
-   definition (<xref target="field.authorization"/>). Likewise, an origin server might use
-   Cache-Control response directives (<xref target="field.cache-control"/>) to supplant Vary if it
-   considers the variance less significant than the performance cost of Vary's
-   impact on caching.
+   An origin server &SHOULD; generate a Vary header on a cacheable response
+   when it wishes that response to be selectively reused for subsequent requests.
+   Generally, that is the case when the selecting header fields have been used to 
+   modify the response in some fashion; for example, <x:ref>Accept-Language</x:ref>
+   might have been used to generate a response that would be inappropriate for
+   requests from clients with different preferences.
```
royfielding, line 6965:
```suggestion
   An origin server &SHOULD; generate a Vary header field on a cacheable response
   when it wishes that response to be selectively reused for subsequent requests.
   Generally, that is the case when the response content has been tailored to better
   fit the preferences expressed by those selecting header fields.
   For example, an origin server that has selected a language for the response
   content based on the request's <x:ref>Accept-Language</x:ref> field can
   include Accept-Language in the Vary field value to inform recipients that
   the content has been selected specifically for that Accept-Language.
```

reschke (approved):

mnot (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6896,72 +6896,88 @@ Accept-Language: da, en-gb;q=0.8, en;q=0.7
...
 </t>
 <sourcecode type="abnf7230"><iref primary="true" item="Grammar" subitem="Vary"/>
   <x:ref>Vary</x:ref> = #( "*" / <x:ref>field-name</x:ref> )
 </sourcecode>
 <t>
-   A Vary field value is a list of request field names, known as the selecting header fields,
-   that might have a role in selecting the representation for this response. Potential selecting
-   header fields are not limited to those defined by this specification.
+   A Vary field value is either the wildcard member "*" or a list of
+   request field names, known as the selecting header fields, that might
```
kaduk, line 6907: This seems to suggest that when the `*` form is used, there is no list...

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -6896,72 +6896,88 @@ Accept-Language: da, en-gb;q=0.8, en;q=0.7
...
+   have had a role in selecting the representation for this response.
+   Potential selecting header fields are not limited to fields defined by
+   this specification.
 </t>
 <t>
-   If the list contains "*", it signals that other aspects of the request might
-   play a role in selecting the response representation, possibly including
-   elements outside the message syntax (e.g., the client's network address).
+   A list containing the member "*" signals that other aspects of the
```
kaduk, line 6913: ...but this admits the possibility of a list containing `*` (and other members).

kaduk (commented):
The rewording prompted by request is much clearer; thank you!
One possible nit inline...

draft-ietf-httpbis-semantics-latest.xml
royfielding: Yes, that's because the ABNF now allows a list containing more than one *, or a mix of names and *. The reason for that is it allows Vary field lines to be appended multiple times. Regardless, the impact is the same for a recipient if they parse according to the ABNF and process as specified in "containing a member".

The way this is phrased is deliberate, since for any given Vary field line there is no reason to send both a * and field names, so we do want to discourage that even if it is allowed by the grammar for other reasons.

kaduk: Thank you for confirming.

PR #941, opened 2021-08-10
Clarify that a false If-Range means ignore the Range

### LINKED ISSUE #940: when If-Range doesn't match, ignore Range
part of #914 Section 13.2.2

        When the method is GET and both Range and If-Range are present,
        evaluate the If-Range precondition:
            if the validator matches and the Range specification is
            applicable to the selected representation, respond 206
            (Partial Content)

        Otherwise,
            all conditions are met, so perform the requested action and
            respond according to its success or failure.

I think that if the If-Range doesn't match, we're supposed to ignore the
Range header field when performing the requested action, which doesn't
seem to match up with this unadorned directive to "perform the requested
action" (which would include the Range header field).
(We might also change point (5) to use the "if true" phrasing that the
other items use in the context of evaluating the precondition.)

royfielding:
Fixes #940

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -7595,7 +7595,7 @@ If-Unmodified-Since: Sat, 29 Oct 1994 19:43:31 GMT
        <li>if true and the Range is
            applicable to the <x:ref>selected representation</x:ref>,
            respond <x:ref>206 (Partial Content)</x:ref></li>
-       <li>if false, ignore the Range header field and
+       <li>otherwise, ignore the Range header field and
            respond <x:ref>200 (OK)</x:ref></li>
      </ul>
    </li>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -7592,11 +7592,11 @@ If-Unmodified-Since: Sat, 29 Oct 1994 19:43:31 GMT
      <x:ref>Range</x:ref> and <x:ref>If-Range</x:ref> are present,
      evaluate the <x:ref>If-Range</x:ref> precondition:</t>
      <ul>
-       <li>if true and the Range is
+       <li>if true and the <x:ref>Range</x:ref> is
            applicable to the <x:ref>selected representation</x:ref>,
            respond <x:ref>206 (Partial Content)</x:ref></li>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7592,14 +7592,16 @@ If-Unmodified-Since: Sat, 29 Oct 1994 19:43:31 GMT
      <x:ref>Range</x:ref> and <x:ref>If-Range</x:ref> are present,
      evaluate the <x:ref>If-Range</x:ref> precondition:</t>
      <ul>
-       <li>if the validator matches and the Range specification is
-           applicable to the <x:ref>selected representation</x:ref>, respond
-           <x:ref>206 (Partial Content)</x:ref></li>
+       <li>if true and the Range is
+           applicable to the <x:ref>selected representation</x:ref>,
+           respond <x:ref>206 (Partial Content)</x:ref></li>
+       <li>if false, ignore the Range header field and
```
mnot, line 7598: Do you mean

```suggestion
       <li>otherwise, ignore the Range header field and
```

?

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7592,14 +7592,16 @@ If-Unmodified-Since: Sat, 29 Oct 1994 19:43:31 GMT
...
      evaluate the <x:ref>If-Range</x:ref> precondition:</t>
      <ul>
-       <li>if the validator matches and the Range specification is
-           applicable to the <x:ref>selected representation</x:ref>, respond
-           <x:ref>206 (Partial Content)</x:ref></li>
+       <li>if true and the Range is
+           applicable to the <x:ref>selected representation</x:ref>,
+           respond <x:ref>206 (Partial Content)</x:ref></li>
+       <li>otherwise, ignore the Range header field and
+           respond <x:ref>200 (OK)</x:ref></li>
```
royfielding, line 7599:
```suggestion
       <li>if true and the <x:ref>Range</x:ref> is
           applicable to the <x:ref>selected representation</x:ref>,
           respond <x:ref>206 (Partial Content)</x:ref></li>
       <li>otherwise, ignore the <x:ref>Range</x:ref> header field and
           respond <x:ref>200 (OK)</x:ref></li>
```

PR #948, opened 2021-08-11
clarify history of redirect status codes (closes #947)

### LINKED ISSUE #947: Section 15.4: tune text about history of redirect status codes
|  *Note:* In HTTP/1.0, the status codes 301 (Moved Permanently)
    |  and 302 (Found) were defined for the first type of redirect
    |  ([RFC1945], Section 9.3).  Early user agents split on whether
    |  the method applied to the redirect target would be the same as
    |  the original request or would be rewritten as GET.  Although
    |  HTTP originally defined the former semantics for 301 and 302
    |  (to match its original implementation at CERN), and defined 303
    |  (See Other) to match the latter semantics, prevailing practice
    |  gradually converged on the latter semantics for 301 and 302 as
    |  well.  The first revision of HTTP/1.1 added 307 (Temporary
    |  Redirect) to indicate the former semantics of 302 without being
    |  impacted by divergent practice.  For the same reason, 308
    |  (Permanent Redirect) was later on added in [RFC7538] to match
    |  301.  [...]

I had to read this text several times to find a way to understand it
that seems to make sense to me (but might still be wrong!).
I think part of my confusion is that the word "former" is being used in
two different senses (the first of the two choices, and the
historical/earlier version). Perhaps it's more clear to just talk about
"method rewriting" (and not rewriting) instead of using the overloaded
term.

reschke:


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8828,6 +8828,27 @@ Content-Range: bytes 7000-7999/8000
...
+    defined <x:ref>303 (See Other)</x:ref> to match the latter semantics (rewriting to GET),
+    prevailing practice gradually converged on the latter semantics for
+    <x:ref>301</x:ref> and <x:ref>302</x:ref> as well. The first revision of
+    HTTP/1.1 added <x:ref>307 (Temporary Redirect)</x:ref> to indicate the
+    former semantics of <x:ref>302</x:ref> without being impacted by divergent practice.
+    For the same reason, <x:ref>308 (Permanent Redirect)</x:ref> was later on
+    added in <xref target="RFC7538"/> to match <x:ref>301</x:ref>.
+    Over 10 years later, most user agents still do method rewriting for
+    <x:ref>301</x:ref> and <x:ref>302</x:ref>; therefore, <xref target="RFC7231"/>
+    made that behavior conformant if and only if the original request is POST.
```
royfielding, line 8849:
```suggestion
    &Note; In HTTP/1.0, the status codes <x:ref>301 (Moved Permanently)</x:ref>
    and <x:ref>302 (Found)</x:ref> were originally defined as method-preserving
    (<xref target="HTTP10" x:fmt="," x:sec="9.3"/>) to match their implementation
    at CERN; <x:ref>303 (See Other)</x:ref> was defined for a redirection that
    changed its method to GET. However, early user agents split on whether to
    redirect POST requests as POST (according to then-current specification)
    or as GET (the safer alternative when redirected to a different site).
    Prevailing practice eventually converged on changing the method to GET.
    <x:ref>307 (Temporary Redirect)</x:ref> and
    <x:ref>308 (Permanent Redirect)</x:ref> <xref target="RFC7538"/> were
    later added to unambiguously indicate method-preserving redirects, and
    <x:ref>301</x:ref>/<x:ref>302</x:ref> have been adjusted to allow a POST
    request to be redirected as GET.
```

royfielding: I think this rewrite is also needed to make the note understandable.

royfielding (changes requested):

mnot (approved):

PR #952, opened 2021-08-12
explicitly refer to how this spec defines access to http or https resources

### LINKED ISSUE #951: Refer to description of "http" and "https" direct access
In Section 7.3.3 (as noted by #914)

    If no proxy is applicable, a typical client will invoke a handler
    routine, usually specific to the target URI's scheme, to connect
    directly to an origin for the target resource. How that is
    accomplished is dependent on the target URI scheme and defined by its
    associated specification.

This document is the relevant specification for the "http" and "https"
URI schemes; a section reference to the corresponding procedures might
be in order.

royfielding:
fixes #951


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2794,18 +2796,35 @@ Host: www.example.org
    satisfy the request.  Proxy configuration is implementation-dependent,
    but is often based on URI prefix matching, selective authority matching,
    or both, and the proxy itself is usually identified by an "http" or
-   "https" URI.  If a proxy is applicable, the client connects inbound by
-   establishing (or reusing) a connection to that proxy.
+   "https" URI.
+</t>
+<t>
+   If an "http" or "https" proxy is applicable, the client connects
```
mnot, line 2802: Applicable, or 'configured'?

mnot (approved):

draft-ietf-httpbis-semantics-latest.xml
royfielding: *shrug*

reschke (approved):

PR #955, opened 2021-08-13
make clock a defined term and use it as such

### LINKED ISSUE #953: explain or xref "with a clock"
Section 10.2.2

    A recipient with a clock that receives a response message without a
    [...]
    A recipient with a clock that receives a response with an invalid

Are we using "with a clock" as shorthand for "have a clock capable of
providing a reasonable approximation of the current instant in
Coordinated Universal Time"? It might be worth clarifying if this
different phrasing than above is intended to convey different semantics.

royfielding:
fixes #953

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -2016,6 +2016,8 @@ Example-Dates: "Sat, 04 May 1996", "Wed, 14 Sep 2005"
 
 <section title="Date/Time Formats" anchor="http.date">
   <x:anchor-alias value="HTTP-date"/>
+  <x:anchor-alias value="clock"/>
+  <iref primary="true" item="clock"/>
 <t>
    Prior to 1995, there were three different formats commonly used by servers
    to communicate timestamps.  For compatibility with old implementations, all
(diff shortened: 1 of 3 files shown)


mnot (approved):
Looks good; should we also update caching's use of 'clock' to refer to this?

royfielding:
> Looks good; should we also update caching's use of 'clock' to refer to this?

Yes, I think so. Certainly the first time it is used (line 575) and we might want to remove or trim some of the redundant mentions elsewhere.

royfielding:
Well, tried but am having difficulty with this sentence

     A cache that does not have a clock available &MUST-NOT; use stored
     responses without revalidating them upon every use.

Do we really need the triple negative here?

I will try again in the morning.

martinthomson:
> A cache without a clock MUST revalidate stored responses upon every use.
?

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2046,13 +2048,17 @@ Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
...
-   values in the asctime format are assumed to be in UTC.
+   An HTTP-date value represents time as an instance of Coordinated
+   Universal Time (UTC). The first two formats indicate UTC by the
+   three-letter abbreviation for Greenwich Mean Time, "GMT", a predecessor
+   of the UTC name; values in the asctime format are assumed to be in UTC.
+</t>
+<t>
+   A <x:dfn>clock</x:dfn> is an implementation capable of providing a
+   reasonable approximation of the current instant in UTC.
    A sender that generates HTTP-date values from a local clock ought to use
```
martinthomson, line 2059:
```suggestion
   A sender that generates HTTP-date values from a local clock <xref target="RFC6919" section="4">ought to</xref> use
```

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
royfielding: Nope, I don't like all-caps. Note that the example OUGHT TO is a reference to RFC2616, though we use it more like a REALLY SHOULD.

royfielding:
> > A cache without a clock MUST revalidate stored responses upon every use.
> > ?

Much better.

mnot (approved):

PR #956, opened 2021-08-13
split when to use validators into other sections

### LINKED ISSUE #933: Old requirements in 8.8.4 When to Use Entity-Tags and Last-Modified Dates
This was part of #914

>    Section 8.8.4

>>        SHOULD send the Last-Modified value in non-subrange cache
>>        validation requests (using If-Modified-Since) if only a Last-
>>        Modified value has been provided by the origin server.
>>
>>        MAY send the Last-Modified value in subrange cache validation
>>        requests (using If-Unmodified-Since) if only a Last-Modified value
>>        has been provided by an HTTP/1.0 origin server. The user agent
>>        SHOULD provide a way to disable this, in case of difficulty.

> I'm failing to come up with an explanation for why it's important to specifically call out the HTTP/1.0 origin server in the latter case. What's special about an HTTP/1.1 origin server that only provided a Last-Modified value and subrange cache validation requests that makes the MAY inapplicable? (What's the actual expected behavior for that situation?)

There exists a bit of rot here. First, this reflects the original implementation of If-Unmodified-Since being applicable to GET and caches, which conflicts with the (new) last sentence of 13.1.4.

>> The If-Unmodified-Since header field can be ignored by caches and intermediaries because it is not applicable to a stored response.

Second, this function (subrange conditional GET requests) is why If-Range was introduced late in the HTTP/1.1 design, and the above text predates that field (which is why it doesn't say If-Range instead).

Third, the text is obsolete because nobody cares about byte range consistency in subrange requests to HTTP/1.0 origin servers, because that was a theoretical concern in early 1996 and there are no longer any HTTP/1.0 origin servers that implement Range, nor would we care if people occasionally got funky spliced content in HTTP/1.0 responses just because more than one modification was made within the same timestamp.

royfielding:
Move requirements on status codes to where they will be seen.
Move requirements on cache validation to where cache validation is defined.
Update both for the (slightly newer) If-Range.

This is mostly editorial but rephrases some existing requirements.

fixes #933

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -2591,6 +2591,7 @@ Expires: Thu, 01 Dec 1994 16:00:00 GMT
 <section title="Since draft-ietf-httpbis-cache-17" anchor="changes.since.17">
 <ul x:when-empty="None yet.">
   <li>Made reference to <xref target="HTTP11"/> informative only (<eref target="https://github.com/httpwg/http-core/issues/911"/>)</li>
+  <li>Move cache-related aspects of validator use from <xref target="HTTP"/> into <xref target="validation.sent"/> (<eref target="https://github.com/httpwg/http-core/issues/933"/>)</li>
 </ul>
 </section>
 </section>
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -974,21 +974,21 @@
    representations with the listed entity-tags).
 </t>
 <t>
-   A cache:
+   When generating a conditional request for validation, a cache:
 </t>
 <ul>
-   <li>&MUST; send the relevant entity-tags in any cache validation request
+   <li>&MUST; send the relevant entity-tags
      (using <x:ref>If-Match</x:ref>, <x:ref>If-None-Match</x:ref>, or
      <x:ref>If-Range</x:ref>) if the entity-tags were provided in the
      stored response(s) being validated.</li>
-   <li>&SHOULD; send the <x:ref>Last-Modified</x:ref> value in non-subrange
-     cache validation requests (using <x:ref>If-Modified-Since</x:ref>) if
-     a single stored response is being validated and that response contains
-     a Last-Modified value.</li>
-   <li>&MAY; send the <x:ref>Last-Modified</x:ref> value in subrange cache
-     validation requests (using <x:ref>If-Unmodified-Since</x:ref> or
-     <x:ref>If-Range</x:ref>) if a single stored response is being validated
-     and that response contains only a Last-Modified value
+   <li>&SHOULD; send the <x:ref>Last-Modified</x:ref> value (using 
+     <x:ref>If-Modified-Since</x:ref>) if the request is not for a subrange, 
+     a single stored response is being validated, and that response 
+     contains a Last-Modified value.</li>
+   <li>&MAY; send the <x:ref>Last-Modified</x:ref> value (using 
+     <x:ref>If-Unmodified-Since</x:ref> or <x:ref>If-Range</x:ref>) if 
+     the request is for a subrange, a single stored response is being 
+     validated, and that response contains only a Last-Modified value
      (not an entity-tag).</li>
 </ul>
 <t>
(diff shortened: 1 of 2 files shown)


royfielding:
fixes #933

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -8509,6 +8475,20 @@ Content-Range: exampleunit 11.2-14.3/25
...
    the method definition or explicit cache controls (see <xref target="heuristic.freshness"/>).
 </t>
+<t>
+   In 200 responses to GET or HEAD, an origin server &SHOULD; send any
+   available validator fields (<xref target="response.validator"/>) for the
+   <x:ref>selected representation</x:ref>, with both a strong entity-tag and
+   a <x:ref>Last-Modified</x:ref> date being preferred.
+</t>
+<t>
+   In 200 responses to state-changing methods, any validator fields
```
reschke, line 8485:
```suggestion
   In 2xx responses to state-changing methods, any validator fields
```

Or is this really specific to 200?

reschke (commented):
So what we had as "client" requirements now entirely moved into the Caching spec.

Is this because clients that are not caches really are not required to do anything with the validators.

...still not sure whether not saying anything about clients in Semantics is good...

draft-ietf-httpbis-semantics-latest.xml
royfielding: This is the section on 200. Yes, it is specific to 200. 201 has a slightly different requirement, and the other 2xx codes have unique requirements on metadata that don't need to be changed. The more generic statement (that validators in response to state-changing methods apply to the selected representation after the state change) is still in the section on validators.

royfielding:
> So what we had as "client" requirements now entirely moved into the Caching spec.

Yes, because that second group of requirements are on cache validation. Hence, they are requirements on caches, not clients.

> Is this because clients that are not caches really are not required to do anything with the validators?

Clients that are not caches cannot do cache validation. There are plenty of other things that clients do with validators that were not discussed by that section. The original was written by Jeff as part of the cache design document for 1.1 and thus did not include the non-cache uses of validators.

> ...still not sure whether not saying anything about clients in Semantics is good...

The stuff about clients is in the section on Conditional Requests (where it belongs), right? We probably should have combined them into a single section, like "Validators and Conditional Requests", but I don't want another reorg.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -973,6 +973,29 @@
...
+   <li>&SHOULD; send the <x:ref>Last-Modified</x:ref> value in non-subrange
+     cache validation requests (using <x:ref>If-Modified-Since</x:ref>) if
+     a single stored response is being validated and that response contains
+     a Last-Modified value.</li>
+   <li>&MAY; send the <x:ref>Last-Modified</x:ref> value in subrange cache
+     validation requests (using <x:ref>If-Unmodified-Since</x:ref> or
+     <x:ref>If-Range</x:ref>) if a single stored response is being validated
+     and that response contains only a Last-Modified value
+     (not an entity-tag).</li>
+</ul>
```
mnot, line 993:
```suggestion
<t>
   When generating a conditional request for validation, a cache:
</t>
<ul>
   <li>&MUST; send the relevant entity-tags
     (using <x:ref>If-Match</x:ref>, <x:ref>If-None-Match</x:ref>, or
     <x:ref>If-Range</x:ref>) if the entity-tags were provided in the
     stored response(s) being validated.</li>
   <li>&SHOULD; send the <x:ref>Last-Modified</x:ref> value (using 
     <x:ref>If-Modified-Since</x:ref>) if the request is not for a subrange, 
     a single stored response is being validated, and that response 
     contains a Last-Modified value.</li>
   <li>&MAY; send the <x:ref>Last-Modified</x:ref> value (using 
     <x:ref>If-Unmodified-Since</x:ref> or <x:ref>If-Range</x:ref>) if 
     the request is for a subrange, a single stored response is being 
     validated, and that response contains only a Last-Modified value
     (not an entity-tag).</li>
</ul>
```

Just specifying 'A cache... must send...' doesn't match the original text; it's when the cache chooses to generate a conditional request that's important.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -973,6 +973,29 @@
...
+<t>
+   In most cases, both validators are sent in cache validation requests,
```
mnot, line 995:
```suggestion
   In most cases, both validators are generated in cache validation requests,
```

mnot (changes requested):

mnot (approved):

reschke (approved):

PR #960, opened 2021-08-16
make preconditions consistent on when they are required to be evaluated …

### LINKED ISSUE #954: align required ignore of conditional request fields on some methods
Section 13.2.1

    server that can provide a current representation. Likewise, a server
    MUST ignore the conditional request header fields defined by this
    specification when received with a request method that does not
    involve the selection or modification of a selected representation,
    such as CONNECT, OPTIONS, or TRACE.

We do say "can be used with any method" regarding If-Match, earlier,
which is not very well aligned with this "MUST ignore".

royfielding:
… and when they can be ignored

fixes #954

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -7236,7 +7236,7 @@ If-None-Match: *
 <t>
    When an origin server receives a request that selects a representation
    and that request includes an If-None-Match header field,
-   the origin server &SHOULD; evaluate the If-None-Match condition as per
+   the origin server &MUST; evaluate the If-None-Match condition as per
    <xref target="evaluation"/> prior to performing the method.
 </t>
 <t>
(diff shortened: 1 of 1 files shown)


mnot (approved):

reschke (changes requested):
I do not disagree with the changes, but I am slightly concerned that we are changing normative requirements here:

- some requirements now only apply when the origin server actually process the conditional header field; that makes the mechanism essentially optional
- at least one MUST became a SHOULD

If we agree on these changes we may also need to document as change from RFC723x.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7219,8 +7234,10 @@ If-None-Match: *
    resource.
 </t>
 <t>
-   An origin server that receives an If-None-Match header field &MUST;
-   evaluate the condition as per <xref target="evaluation"/> prior to performing the method.
+   When an origin server receives a request that selects a representation
+   and that request includes an If-None-Match header field,
+   the origin server &SHOULD; evaluate the If-None-Match condition as per
```
royfielding, line 7239:
```suggestion
   the origin server &MUST; evaluate the If-None-Match condition as per
```

royfielding:
> I do not disagree with the changes, but I am slightly concerned that we are changing normative requirements here:
> 
>     * some requirements now only apply when the origin server actually process the conditional header field; that makes the mechanism essentially optional

The text above those requirements already requires that the origin server evaluate the condition, so there shouldn't be a change.

>     * at least one MUST became a SHOULD

Oops, that was my mistake when I pasted some text. Fixed.

> If we agree on these changes we may also need to document as change from RFC723x.

I don't think there are any, unless the part about caches and intermediaries ignoring IUS and If-Match was not documented previously.

reschke:
> The text above those requirements already requires that the origin server evaluate the condition, so there shouldn't be a change.

I see. I'm still confused about that change though; a sloppy reader like me might get the impression that this implies some optionality.

reschke:
> I don't think there are any, unless the part about caches and intermediaries ignoring IUS and If-Match was not documented previously.

Ack. I see https://greenbytes.de/tech/webdav/rfc7232.html#rfc.section.3.1.p.7 and https://greenbytes.de/tech/webdav/rfc7232.html#rfc.section.3.4.p.8 and

reschke (approved):

PR #961, opened 2021-08-16
use chosen response instead of selected response and avoid selected resource

### LINKED ISSUE #958: Selected [representation, response]
We use the phrases 'selected representation' and 'selected response' in Semantics and Caching (respectively) to refer to two completely different concepts.

Can we do better?

royfielding:
fixes #958


mnot (commented):
See #963.

reschke:
> See #963.

Me confused. Which PR is relevant now? Both?

mnot:
The other one is an evolution of this one.

PR #962, opened 2021-08-16
remove mention of obs-fold line folding from semantics

### LINKED ISSUE #923: removal of obs-fold from field-value causes inconsistencies
Citing Ben Kaduk (https://github.com/httpwg/http-core/issues/915):

> > A sender MUST NOT generate a message that includes line folding
(i.e., that has any field line value that contains a match to the
obs-fold rule) unless [...]

> Since we don't include the obs-fold production as a component of any
other production, and field-value excludes CRLF, it seems that any such
field line value would already be in violation of the ABNF and thus
forbidden. I don't really want to advocate for including obs-fold in
the field-value production in -semantics, though, so maybe accepting
this nit is the least bad choice here.

In RFC723x, field-value allowed obs-fold. We need to:

- tune the prose above, and
- somehow properly allow it in message/http (maybe by modifying the field-value for that specific case???)

royfielding:
… (now entirely handled by HTTP/1.1 messaging).
Include LF in discussion of what characters are invalid in field values.
Require obs-fold to be processed when consuming message/http data.

fixes #923

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -13776,6 +13776,7 @@ Content-Type: text/plain
   <li>In <xref target="field.te"/>, add a forward reference to "weight" (<eref target="https://github.com/httpwg/http-core/issues/914"/>)</li>
   <li>In <xref target="field.accept-encoding"/>, clarify that the examples contains multiple values; also remove obsolete HTTP/1.0 note about qvalues (<eref target="https://github.com/httpwg/http-core/issues/914"/>)</li>
   <li>In <xref target="status.3xx"/>, remove incorrect mention of Etag as request field (<eref target="https://github.com/httpwg/http-core/issues/914"/>)</li>
+  <li>Move text about obs-fold in message/http to <xref target="HTTP11"/>; also note that LF is forbidden in field values just as CR and NUL (<eref target="https://github.com/httpwg/http-core/issues/923"/>)</li>
   <li>In <xref target="message.transformations"/>, properly refer to text that has moved to <xref target="HTTP11"/> (<eref target="https://github.com/httpwg/http-core/issues/930"/>)</li>
   <li>Rewrite description of validators and move cache-related aspects into <xref target="CACHING"/> (<eref target="https://github.com/httpwg/http-core/issues/933"/>)</li>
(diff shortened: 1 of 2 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -1701,18 +1701,19 @@ Example-Field: Baz
...
    Specifications for newly defined fields &SHOULD; limit their values to
    visible US-ASCII octets (VCHAR), SP, and HTAB.
-   A recipient &SHOULD; treat other octets in field content (obs-text) as
-   opaque data.
+   A recipient &SHOULD; treat other allowed octets in field content
+   (i.e., <x:ref>obs-text</x:ref>) as opaque data.
 </t>
 <t>
-   Field values containing CR or NUL characters are invalid and dangerous,
+   Field values containing CR, LF, or NUL characters are invalid and dangerous,
```
martinthomson, line 1708: I reviewed the text on field *names* and noticed that there is nothing about COLON there.  Is it worth adding an explicit prohibition there of a similar shape to this?

martinthomson (approved):
This is a good change.

A few things that might be worth considering, when comparing this to recent HTTP/2 changes.  HTTP/2 prohibits additional things in field names: SP (0x20), COLON (0x3a), 0x01-0x1f, and 0x7f-0xff (inclusive), specifically.

The other things in h2 are specific to that.  That includes leading and trailing space in values (you just say "A field value does not include leading or trailing whitespace") and uppercase field characters are version-specific.

mnot (approved):

draft-ietf-httpbis-semantics-latest.xml
royfielding: Maybe, but I think that would be a different issue. It would be nice if we had one place for all these special handling requirements that are really protocol errors but have to be handled the same way by all versions.

royfielding: see #30

reschke (approved):

draft-ietf-httpbis-semantics-latest.xml
royfielding: Note that HTTP already limits the field-name as described, except for the lowercasing and allowing the pseudo-header-specific colon that are version-specific, and also doesn't allow token delimiters. I've added an issue on that for http2bis.

PR #963, opened 2021-08-17
Mnot 958

### LINKED ISSUE #958: Selected [representation, response]

mnot:
Additional instances of 'select'

Fixes #958

### DIFF
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -624,9 +624,10 @@
    (e.g., qvalues on <x:ref>Accept</x:ref> and similar request header
    fields), that mechanism &MAY; be used to choose a preferred response.
    If such a mechanism is not available, or leads to equally preferred
-   responses, the most recent response (as determined by the
-   <x:ref>Date</x:ref> header field) is chosen, as per <xref
-   target="constructing.responses.from.caches"/>.
+   responses, the most recent
+   response (as determined by the <x:ref>Date</x:ref> header field) is
+   chosen, as
+   per <xref target="constructing.responses.from.caches"/>.
 </t>
 <t>
    Some resources mistakenly omit the Vary header field from their default
--- a/draft-ietf-httpbis-cache-latest.xml
+++ b/draft-ietf-httpbis-cache-latest.xml
@@ -268,7 +268,7 @@
 </t>
 <iref item="cache key" />
 <t>
-   The <x:dfn>cache key</x:dfn> is the information a cache uses to select a response and
+   The <x:dfn>cache key</x:dfn> is the information a cache uses to choose a response and
    is composed from, at a minimum, the request method and target
    URI used to retrieve the stored response; the method determines under which
    circumstances that response can be used to satisfy a subsequent request. However, many
@@ -278,7 +278,7 @@
 <t>
    If a request target is subject to content negotiation, the cache might
    store multiple responses for it. Caches differentiate these responses
-   by incorporating values of the original request's selecting header fields
+   by incorporating chosen values from the original request's header fields
    into the cache key as well, using information in the <x:ref>Vary</x:ref>
    response header field, as per <xref target="caching.negotiated.responses"/>.
 </t>
@@ -518,7 +518,7 @@
    that of the stored response match, and</t></li>
    <li><t>the request method associated with the stored response allows it to
    be used for the presented request, and</t></li>
-   <li><t>selecting header fields nominated by the stored response (if any)
+   <li><t>request header fields nominated by the stored response (if any)
    match those presented (see <xref target="caching.negotiated.responses"
    />), and</t></li>
    <li><t>the stored response does not contain the no-cache cache directive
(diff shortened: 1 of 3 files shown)


mnot:
I went with 'nominated' in a few places where 'choosing' didn't seem workable. 'varying' could work too, I suppose, but we'd probably need to add it as a defined term.

royfielding:
I think that "selecting header fields" was fine and reads better than nominated or chosen fields — they are the request fields that the origin claims were used to select each representation, so the name fits even if we use them later to choose a response.

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -582,12 +582,12 @@
...
    the cache &MUST-NOT; use that stored response without revalidation unless
-   all the selecting header fields (nominated by that Vary field value) in the
-   present request match those fields in the original request (i.e., the
+   all the presented request header fields nominated by that Vary field value
+   match those fields in the original request (i.e., the
    request that caused the cached response to be stored).
 </t>
 <t>
-   The selecting header fields from two requests are defined to match if
+   The nominated header fields from two requests are defined to match if
```
mnot, line 590:
```suggestion
   The header fields from two requests are defined to match if
```

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -616,33 +616,34 @@
...
-   <x:dfn>selected response</x:dfn>.
-</t>
-<t>
-   If multiple selected responses are available (potentially including
-   responses without a Vary header field), the cache will need to choose one to use.
-   When a selecting header field has a known mechanism for doing so (e.g., qvalues on
-   <x:ref>Accept</x:ref> and similar request header fields), that mechanism &MAY; be
-   used to select a preferred response. If such a mechanism is not available, or leads to equally preferred responses, the most recent
-   response (as determined by the <x:ref>Date</x:ref> header field) is used, as
+   If multiple stored responses match the nominated header fields
```
mnot, line 620:
```suggestion
   If multiple stored responses match the request header fields nominated by Vary
```

mnot:
@royfielding see updates; I think I managed to avoid the awkwardness

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -278,7 +278,7 @@
 <t>
    If a request target is subject to content negotiation, the cache might
    store multiple responses for it. Caches differentiate these responses
-   by incorporating values of the original request's selecting header fields
+   by incorporating chosen values from the original request's header fields
```
royfielding, line 281:
```suggestion
   by incorporating some of the original request's header fields
```

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -616,33 +616,34 @@
...
-</t>
-<t>
-   If multiple selected responses are available (potentially including
-   responses without a Vary header field), the cache will need to choose one to use.
-   When a selecting header field has a known mechanism for doing so (e.g., qvalues on
-   <x:ref>Accept</x:ref> and similar request header fields), that mechanism &MAY; be
-   used to select a preferred response. If such a mechanism is not available, or leads to equally preferred responses, the most recent
-   response (as determined by the <x:ref>Date</x:ref> header field) is used, as
+   If multiple stored responses match the request header fields nominated by Vary
+   (potentially including responses without a Vary header field),
```
royfielding, line 621:
```suggestion
   If multiple stored responses match,
```

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -616,33 +616,34 @@
...
-   more preferable responses are available. When a cache has multiple responses for a
-   target URI and one or more omits the Vary header field, it &SHOULD; use the
-   most recent (see <xref target="age.calculations"/>) valid Vary field value available to select an appropriate response
-   for the request.
+   response (i.e., the one sent when the request does not express any preferences),
+   with the effect of choosing it for subsequent requests to that resource
+   even when more preferable responses are available. When a cache has
+   multiple stored responses for a target URI and one or more omits the Vary
+   header field, the cache &SHOULD; choose the most recent
+   see <xref target="age.calculations"/>) stored response with a valid Vary
```
royfielding, line 639:
```suggestion
   header field, the cache &SHOULD; choose the most recent
   (see <xref target="age.calculations"/>) stored response with a valid Vary
```

draft-ietf-httpbis-cache-latest.xml
royfielding: Also, the first sentence would be better as

>   A cache might store multiple responses for a request target that is subject to content negotiation.

royfielding (changes requested):

draft-ietf-httpbis-cache-latest.xml
```diff
@@ -278,7 +278,7 @@
 <t>
    If a request target is subject to content negotiation, the cache might
    store multiple responses for it. Caches differentiate these responses
```
mnot, line 280:
```suggestion
   A cache might store multiple responses for a request target that is subject 
   to content negotiation. Caches differentiate these responses
```

royfielding (approved):

mosi2021v (commented):
#

Issue #966, opened 2021-09-04
Artart last call review of draft-ietf-httpbis-messaging-18


reschke:
Reviewer: Marco Tiloca
Review result: Ready with Nits

Thanks for this document! I have found it very well written and I believe it's
basically ready.

Please, see below some minor comments and nits.

Best,
/Marco

 - [X]    [Section 1.2]

* As to "absolute-path", it is more precise to point to Section 4.1 of [HTTP].

**Editors:** https://github.com/httpwg/http-core/commit/26a3a5e5e81b74a4b73e0bd54c29de0a04090b70

 - [X]    [Section 3]

* "HTTP does not place a predefined limit on the length of a request-line, as
described in Section 2 of [HTTP]"

   This can better point to Section 2.3 of [HTTP].

**Editors**: https://github.com/httpwg/http-core/pull/967/commits/17d19589c21c1434bd98c4d1083e1ef539adf469

 - [X]    [Section 3.2]

* "A client MUST send a Host header field in all HTTP/1.1 request messages."

   This sentence can be expanded to point to Section 7.2 of [HTTP].

* "... excluding any userinfo subcomponent and its "@" delimiter ..."

   This should point to Section 4.2.4 of [HTTP].

**Editors**: https://github.com/httpwg/http-core/commit/dbee9841936b5c2bbe98f2560b8c87ce4e875eb7

 - [X]    [Section 3.3]

* "Alternatively, it might be better to redirect the request to a safe resource
that explains how to obtain a new client."

   Is "client" actually the intended word here? Or is it about using
   redirection to explain the client how to obtain something else (e.g. a
   proper authority component for a follow-up request) ?

**Editors**: client is the intended word here: this case is only possible if the client sent an HTTP/0.9 or HTTP/1.0 request without a Host header field, which should not be in any deployed client since 1995.

 - [X]    [Section 7.1.2]

* I believe it's better for the reader if the last paragraph is split into 2 or
3 sentences 

**Editors**: #969 

 - [X]    [Section 9.8]

* "When encountering an incomplete close, a client SHOULD treat as completed
all requests for which it has received ..."

   Shouldn't this be about received responses? Or does it refer to the
   completion of the exchanges started by the mentioned requests?

**Editors**: yes, it's about the completion of the exchanges.

 - [X]    [Appendix A]

* The first paragraph can better point to Section 5.6.1 of [HTTP].

[...]

PR #969, opened 2021-09-07
split the long requirement on decoding chunked trailer fields into …

royfielding:
…three separate requirements specific to each decision

For #966

### DIFF
--- a/draft-ietf-httpbis-messaging-latest.xml
+++ b/draft-ietf-httpbis-messaging-latest.xml
@@ -1383,12 +1383,12 @@ Transfer-Encoding: gzip, chunked
    A recipient that removes the chunked encoding from a message &MAY;
    selectively retain or discard the received trailer fields. A recipient
    that retains a received trailer field &MUST; either store/forward the
-   trailer field separately from the received header fields (e.g., within a
-   separate table or trailer section) or merge the received trailer field
-   into the header section. A recipient &MUST-NOT; merge a received trailer
-   field into the header section unless its corresponding header field
-   definition explicitly permits and instructs how the trailer field value
-   can be safely merged.
+   trailer field separately from the received header fields or merge the
+   received trailer field into the header section.
+   A recipient &MUST-NOT; merge a received trailer field into the
+   header section unless its corresponding header field definition
+   explicitly permits and instructs how the trailer field value can be
+   safely merged.
 </t>
 </section>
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1380,13 +1380,15 @@ Transfer-Encoding: gzip, chunked
...
-   any received trailer fields, store/forward them separately from the header
-   fields, or selectively merge into the header section only those trailer
-   fields corresponding to header field definitions that are understood by
-   the recipient to explicitly permit and define how their corresponding
-   trailer field value can be safely merged.
+   A recipient that removes the chunked encoding from a message &MAY;
+   selectively retain or discard the received trailer fields. A recipient
+   that retains a received trailer field &MUST; either store/forward the
+   trailer field separately from the received header fields (e.g., within a
+   separate table or trailer section) or merge the received trailer field
```
reschke, line 1387: separate "table"?

reschke (changes requested):

draft-ietf-httpbis-messaging-latest.xml
royfielding: yes, that's one example

royfielding: they can be stored as a separate lookup table or forwarded as a separate trailer section

draft-ietf-httpbis-messaging-latest.xml
```diff
@@ -1380,13 +1380,15 @@ Transfer-Encoding: gzip, chunked
...
+   into the header section. A recipient &MUST-NOT; merge a received trailer
+   field into the header section unless its corresponding header field
+   definition explicitly permits and instructs how the trailer field value
+   can be safely merged.
```
royfielding, line 1391:
```suggestion
   A recipient that removes the chunked encoding from a message &MAY;
   selectively retain or discard the received trailer fields. A recipient
   that retains a received trailer field &MUST; either store/forward the
   trailer field separately from the received header fields or merge the
   received trailer field into the header section.
   A recipient &MUST-NOT; merge a received trailer field into the
   header section unless its corresponding header field definition
   explicitly permits and instructs how the trailer field value can be
   safely merged.
```

draft-ietf-httpbis-messaging-latest.xml
royfielding: easier to just remove the example

reschke (approved):

mnot (approved):

Issue #914, opened 2021-07-29
Ben Kaduk's COMMENT on semantics


mnot:
- [x] 10.2 note

In Section 10.2 we had some
good text cleanup (I think, prompted by one of my comments -- thank you!),
but the removed text included a note about how the semantics of a response
header field might be refined by the semantics of the request method and/or
the response status code.  That seems like it would be useful to have
mentioned, and I'm not sure if this text was replicated elsewhere.

**Editors**: The problem with that wording was that a field has defined semantics that include variations based on the context in which it appears, and those variations are included as part of the field definition (not refined by other parts of the specification). Ultimately, we decided that this doesn't need to be said in the section intro.

---
- [x] updating 3864

This document updates RFC 3864, which is part of BCP 90.
However, this document is targeting Proposed Standard status, which
means it cannot become a part of BCP 90 as part of that update.
Did we consider splitting out the RFC 3864 updates into a separate,
BCP-level, document, that would become part of BCP 90?

**Editors**: The intended status for this document is full standard. The reason it updates RFC 3864 (which previously defined the registry for HTTP header fields as part of a registry for all application-level IMF-like protocols) is that IETF thinking has changed since 3864. Having a single IMF-wide definition of fields was unsuccessful and led to more confusion when fields diverged. Hence, this document is obsoleting only the HTTP parts of RFC 3864 by moving them back to the standards track. It is an update only because there is no status for partial obsoleting.

---
- [x] Section 1.2

>   HTTP/2 ([RFC7540]) introduced a multiplexed session layer on top of
   the existing TLS and TCP protocols for exchanging concurrent HTTP
   messages with efficient field compression and server push.  HTTP/3
   ([HTTP3]) provides greater independence for concurrent messages by
   using QUIC as a secure multiplexed transport over UDP instead of TCP.

My understanding was that h2 and h3 also use non-text-based headers, in
contrast to HTTP/1.1's "text-based messaging syntax" that we mention
earlier.  Is that non-text nature worth noting here?

**Editors**: No. This is not an overview of the differences between the protocols; just a brief introduction.

---
- [x] Section 3.7

>  Proxies are often used to group an
   organization's HTTP requests through a common intermediary for the
   sake of security, annotation services, or shared caching.  [...]

The term "security" can mean so many different things to different
audiences that its meaning in isolation is pretty minimal.  I suggest
finding a more specific term for the intended usage, perhaps relating to
an auditing, exfiltration protection, and/or content-filtering function.

[...]

PR #974, opened 2021-09-09
Avoid bare 'for the sake of security'

kaduk:
Inspired by some discussion on #914 : while just the word "security"
is indeed used in marketing literature for proxies, it's meaning to different
parties is so varied so as to not really convey much useful information.  In the
IETF we try hard to be clear about what security provides we need and/or provide,
avoiding the vague catch-all term.  However (as I am reminded out of band), the
audience of this document is not exactly security experts, and it is easy to go
overboard trying to be precise, at least in this instance.

[actual commit message body retained below]

The mere act of inserting a proxy into the chain does not, in and of
itself, do much of anything for the security of the system (and a badly
implemented proxy can make the security of the system much worse).
A proxy can, however, provide security services such as auditing
access, annotating content from untrustworthy sources, exfiltration
avoidance, etc.  It is these services that are the security-related
motivation for using a proxy, so say that "security services", rather
than just "security", are being provided by the proxy.

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -818,9 +818,9 @@
    other requests might require translation to and from entirely different
    application-level protocols. Proxies are often used to group an
    organization's HTTP requests through a common intermediary for the
-   sake of security, annotation services, or shared caching. Some proxies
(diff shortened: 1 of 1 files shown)


royfielding (approved):

mosi2021v (commented):
-

PR #975, opened 2021-09-09
Wordsmith future guidance on reactive negotiation

kaduk:
It may or may not be useful to have an HTTP extension in order to
perform reactive negotiation -- one possible approach that does not
require an extension would be policy in the user-agent.

In the spirit of "less is more", and to not overly constrain our
forward-looking statement, just remove the phrase "as an extension".

(inspired by follow-up discussions on #914)


royfielding (approved):
I almost made the same edit earlier.

PR #976, opened 2021-09-09
wordsmith automatic-link-editing guidance

kaduk:
The current text of "ought to ... where possible" is easy to read
as saying "if you have the technical capability, do it".
There are, however, some subtleties, such as if you get the 308 response
over http-not-s, in which case the question of whether or not to
re-link is not so clear-cut.

Change from "where possible" to "where appropriate" to hint that there
is some logic needed here beyond "blindly accept".

(inspired by follow-up discussions from #914)

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -13836,6 +13836,7 @@ Content-Type: text/plain
 <ul x:when-empty="None yet.">
   <li>In <xref target="field.accept"/>, align text about "q" parameter with recent changes to IANA media types registry,
   and instruct IANA to reference this document with respect to the "q" special case (<eref target="https://github.com/httpwg/http-core/issues/970"/>)</li>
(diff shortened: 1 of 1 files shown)


royfielding (changes requested):
I think that begs a follow-on question about when it is appropriate.
I would be more comfortable with changing the entire sentence to be

```
   The server is suggesting that a user agent with link-editing capability
   can permanently replace references to the target URI with one of the
   new references sent by the server. However, this suggestion is usually
   ignored unless the user agent is actively editing references
   (e.g., engaged in authoring content), the connection is secured, and
   the origin server is a trusted authority for the content being edited.
```

kaduk:
I like Roy's version better than mine :)
I guess I was too conservative...

mnot (approved):

reschke (changes requested):
The change itself is good, but it needs to be done for 301 as well.

PR #977, opened 2021-09-09
strawman for referencing the URI security considerations

kaduk:
Inspired by follow-up discussions on #914 , make a concrete proposal for how the RFC 3986 security considerations might be mentioned in our own security considerations.
I understand the editors' stance on adding substantial new text at this point, and will defer to them on whether this constitutes "substantial".

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -10470,7 +10470,7 @@ Content-Type: text/plain
    received via HTTP, or secure use of the Internet in general, rather than
    security of the protocol. The security considerations for URIs, which
    are fundamental to HTTP operation, are discussed in
-   <xref target="URI" x:fmt="," x:sec="7"/>. Various organizations maintain
+   <xref target="URI" x:fmt="of" x:sec="7"/>. Various organizations maintain
    topical information and links to current research on Web application
    security (e.g., <xref target="OWASP"/>).
 </t>
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -10468,8 +10468,8 @@ Content-Type: text/plain
    related to HTTP semantics are about securing server-side applications (code
    behind the HTTP interface), securing user agent processing of content
    received via HTTP, or secure use of the Internet in general, rather than
-   security of the protocol. URIs are fundamental to HTTP operation; the
-   security considerations of URIs are discussed in
+   security of the protocol. The security considerations for URIs, which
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -10468,9 +10468,11 @@ Content-Type: text/plain
...
    behind the HTTP interface), securing user agent processing of content
    received via HTTP, or secure use of the Internet in general, rather than
-   security of the protocol. Various organizations maintain topical
-   information and links to current research on Web application security
-   (e.g., <xref target="OWASP"/>).
+   security of the protocol. URIs are fundamental to HTTP operation; the
+   security considerations of URIs are discussed in
+   <xref target="URI" x:fmt="," x:sec="7"/>. Various organizations maintain
+   topical information and links to current research on Web application
+   security (e.g., <xref target="OWASP"/>).
```
royfielding, line 10475: I think it would read better as
```suggestion
   security of the protocol. The security considerations for URIs, which
   are fundamental to HTTP operation, are discussed in
   <xref target="URI" x:fmt="," x:sec="7"/>. Various organizations maintain
   topical information and links to current research on Web application
   security (e.g., <xref target="OWASP"/>).
```

royfielding (changes requested):

draft-ietf-httpbis-semantics-latest.xml
kaduk: Leading with "security considerations" does feel more parallel to the previous sentences.
There is perhaps a theoretical ambiguity in this phrasing as to whether it is URIs that are fundamental to HTTP or the security considerations thereof ... which suggests the even shorter "The security considerations for URIs are discussed in [ref]".

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -10468,9 +10468,11 @@ Content-Type: text/plain
    related to HTTP semantics are about securing server-side applications (code
    behind the HTTP interface), securing user agent processing of content
    received via HTTP, or secure use of the Internet in general, rather than
-   security of the protocol. Various organizations maintain topical
-   information and links to current research on Web application security
-   (e.g., <xref target="OWASP"/>).
+   security of the protocol. The security considerations for URIs, which
+   are fundamental to HTTP operation, are discussed in
+   <xref target="URI" x:fmt="," x:sec="7"/>. Various organizations maintain
```
reschke, line 10473:
```suggestion
   <xref target="URI" x:fmt="of" x:sec="7"/>. Various organizations maintain
```

PR #981, opened 2021-10-04
Standardize hyphenation for "case-sensitive" and "case-insensitive"

danfoygel:


reschke:
The document is already in the publication queue. The RFC Editor team will take care of remaining consistency issues.

PR #985, opened 2021-10-27
Slightly relax requirements for handling Accept-Encoding field values (closes #980)

### LINKED ISSUE #980: Recommended response behavior when identity encoding is unacceptable
[RFC 2616](https://www.rfc-editor.org/rfc/rfc2616.html#section-14.3) specified that identity content-coding could be made unacceptable (e.g., indicating very strong preference for response compression) by means of an Accept-Encoding request header field value like `identity;q=0` or `*;q=0` without an identity-specific preference, and then recommended that the server SHOULD respond with 406 (Not Acceptable) when it cannot send an acceptable response. But [RFC 7231](https://www.rfc-editor.org/rfc/rfc7231.html#section-5.3.4) and the [current draft](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.12.5.3.p.9) update that last point to "SHOULD send a response without any content-coding", presumably primarily intended to avoid sending 406 (Not Acceptable) when the acceptability of identity content-coding is unspecified. The text still includes a description of when identity content-coding is unacceptable and still **accommodates** a "compression is mandatory" interpretation of `Accept-Encoding: gzip, identity;q=0`, but seems to strongly discourage it in the sense of [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.html) ("_there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course_").

I suspect this to be unintentional, and would propose clarification in semantics:
```patch
    An Accept-Encoding header field with a field value that is empty
    implies that the user agent does not want any content coding in response.
    If an Accept-Encoding header field is present in a request and none of the
    available representations for the response have a content coding that
    is listed as acceptable, the origin server &SHOULD; send a response
-   without any content coding.
+   without any content coding unless absence of content coding is itself
+   unacceptable (in which case the origin server &SHOULD; honor that
+   preference).
```

In the absence of such clarification, the text describing when identity content-coding is unacceptable seems to be largely pointless.

reschke:


yadij (commented):
LGTM

PR #1010, opened 2022-02-04
slight editorial tweaks to Connection

### LINKED ISSUE #1009: Editorial tweaks to Connection
The Connection header field is oddly inconsistent with the other field definitions in how it defines the grammar late in the description. I think it is worth an editorial fix.

royfielding:
Fixes #1009

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -2936,6 +2936,13 @@ Host: www.example.org
    The "Connection" header field allows the sender to list desired
    control options for the current connection.
 </t>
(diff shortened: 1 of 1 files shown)


mnot (approved):

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -2949,7 +2956,7 @@ Host: www.example.org
    connection-option in this field, remove any header or trailer field(s) from
    the message with the same name as the connection-option, and then
    remove the Connection header field itself (or replace it with the
-   intermediary's own connection options for the forwarded message).
+   intermediary's own control options for the forwarded message).
```
reschke, line 2959: @royfielding - was that change intentional?

royfielding: Yes, they are referred to as control options in the definition. A connection-option is the ABNF term.

PR #1012, opened 2022-02-28
clarify handling of invalid or unsatisfiable range requests

### LINKED ISSUE #1011: Clarify range validity/satisfiability
A couple of related issues:
First, from section 14.1.1:

>    An int-range is invalid if the last-pos value is present and less
>    than the first-pos.

and from 14.1.2:

>    If a valid bytes range-set includes at least one range-spec with a
>    first-pos that is less than the current length of the representation,
>    or at least one suffix-range with a non-zero suffix-length, then the
>    bytes range-set is satisfiable.  Otherwise, the bytes range-set is
>    unsatisfiable.

So the first bug is that by this definition the invalid range-set "6-2" for a representation of length 10 would be considered satisfiable since the first-pos is less than the current length, even though the range-spec is invalid. It should probably say "at least one _valid_ range-spec".

In the latter quote, what does "valid bytes range-set" mean? A range-set's validity or invalidity is not explicitly defined, unless it refers here only to the syntactic correctness (no illegal chars etc.) regardless of whether any range-specs in it are invalid.

More generally, it is unclear what the relationship between an invalid range-spec and invalid range-set is: is the set invalid if at least one spec is invalid, or should invalid specs just be ignored as long as there are other valid ones in the set?

This is similarly confusing in section 14.2:

>    If all of the preconditions are true, the server supports the Range
>    header field for the target resource, and the specified range(s) are
>    valid and satisfiable (as defined in Section 14.1.2), the server
>    SHOULD send a 206 (Partial Content) response with a content
>    containing one or more partial representations that correspond to the
>    satisfiable ranges requested.
> 
>    ...
> 
>    If all of the preconditions are true, the server supports the Range
>    header field for the target resource, and the specified range(s) are
>    invalid or unsatisfiable, the server SHOULD send a 416 (Range Not
>    Satisfiable) response.

it mentions range(s) being valid/invalid or satisfiable/unsatisfiable, but valid/invalid is only defined for a range-spec, and satisfiable/unsatisfiable is only defined for a range-set (and not individual ranges). btw there are later references to a single range being satisfiable/unsatisfiable, although this is undefined.

Here too there is ambiguousness - it's unclear what happens if e.g. there is one invalid range-spec in the set (such as "1-5,9-7") - does that mean the "the specified range(s) are valid and satisfiable" (not all of them are) or "the specified range(s) are invalid or unsatisfiable" (not all of them are).

I think this should be clarified - either explicitly define valid/invalid/satisfiable/unsatisfiable separately for range-spec and for range-set, and then use the appropriate terminology in section 14.2 (range-spec or range-set rather than "range(s)"), or just explicitly define how an invalid range-spec affects the range-set (ignore it or make the whole set invalid), and/or qualify in both paragraphs in 14.2 whether it is 'all ranges' or 'at least one range' or whatever is deemed correct.

royfielding:
Fixes #1011 

This is longer than I wanted but has the benefit of defining a satisfiable ranges-specifier in general (instead of only for bytes requests), how to handle an invalid ranges-specifier as a bad request, and defines which code to use for an unsupported range-unit.

I believe all of this was implied by the existing text and how we have used valid/invalid, but I think it is clearer stated this way instead of the prior generalizations.

YMMV. This probably should be reviewed by the WG (just to be sure).

### DIFF
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -7731,8 +7731,8 @@ If-Unmodified-Since: Sat, 29 Oct 1994 19:43:31 GMT
 </t>
 
 <section title="Range Specifiers" anchor="range.specifiers">
-  <iref primary="true" item="satisfiable"/>
-  <iref primary="true" item="unsatisfiable"/>
+  <iref primary="true" item="satisfiable range"/>
+  <iref primary="true" item="unsatisfiable range"/>
 <t>
    Ranges are expressed in terms of a range unit paired with a set of range
    specifiers. The range unit name determines what kinds of range-spec
@@ -7858,7 +7858,7 @@ bytes=500-999
    is one less than the current length of the selected representation).
 </t>
 <t>
-   A client can request the last N bytes (N &gt; 0) of the selected
+   A client can refer to the last N bytes (N &gt; 0) of the selected
    representation using a <x:ref>suffix-range</x:ref>.
    If the selected representation is shorter than the specified
    <x:ref>suffix-length</x:ref>, the entire representation is used.
@@ -7895,7 +7895,8 @@ bytes=500-700,601-999
    </li>
 </ul>
 <t>
-   A valid bytes <x:ref>range-spec</x:ref> is satisfiable if it is either:
+   For a <x:ref>GET</x:ref> request, a valid bytes <x:ref>range-spec</x:ref>
+   is <x:ref>satisfiable<x:ref> if it is either:
 </t>
 <ul>
    <li>an <x:ref>int-range</x:ref> with a <x:ref>first-pos</x:ref> that
@@ -7905,9 +7906,10 @@ bytes=500-700,601-999
        <x:ref>suffix-length</x:ref>.</li>
 </ul>
 <t>
-   If the selected representation has zero length, the only satisfiable form of
-   <x:ref>range-spec</x:ref> is a <x:ref>suffix-range</x:ref> with a non-zero
-   <x:ref>suffix-length</x:ref>.
+   When a selected representation has zero length, the only
+   <x:ref>satisfiable<x:ref> form of <x:ref>range-spec</x:ref> in a
+   <x:ref>GET</x:ref> request is a <x:ref>suffix-range</x:ref> with a
+   non-zero <x:ref>suffix-length</x:ref>.
 </t>
 <t>
    In the byte-range syntax, <x:ref>first-pos</x:ref>,
--- a/draft-ietf-httpbis-semantics-latest.xml
+++ b/draft-ietf-httpbis-semantics-latest.xml
@@ -7954,10 +7954,12 @@ bytes=500-700,601-999
 </t>
 <t>
    A server that supports range requests &MAY; ignore or reject a
-   <x:ref>Range</x:ref> header field that consists of more than two
-   overlapping ranges, or a set of many small ranges that are not listed
-   in ascending order, since both are indications of either a broken client or
-   a deliberate denial-of-service attack (<xref target="overlapping.ranges"/>).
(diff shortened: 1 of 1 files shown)


mnot (approved):
Think this is ok, but admittedly I don't have all of the state for ranges paged in at the moment.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7972,8 +7991,17 @@ bytes=500-700,601-999
 </t>
 <t>
    If all of the preconditions are true, the server supports the Range header
-   field for the target resource, and the specified range(s) are invalid or
-   unsatisfiable, the server &SHOULD; send a
+   field for the target resource, and the received Range field-value contains
+   an invalid <x:ref>ranges-specifier</x:ref>,
+   the server &SHOULD; send a <x:ref>400 (Bad Request)</x:ref> response.
+</t>
```
reschke, line 7997: I'm not entirely happy that we seem to add a nwe normative requirement to produce a 400 here.

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7717,6 +7717,8 @@ If-Unmodified-Since: Sat, 29 Oct 1994 19:43:31 GMT
 </t>
 
 <section title="Range Specifiers" anchor="range.specifiers">
+  <iref primary="true" item="satisfiable"/>
```
reschke, line 7720: maybe "satisfiable range"

draft-ietf-httpbis-semantics-latest.xml
```diff
@@ -7867,13 +7881,15 @@ bytes=500-700,601-999
...
-   then the bytes <x:ref>range-set</x:ref> is satisfiable. Otherwise,
-   the bytes <x:ref>range-set</x:ref> is unsatisfiable.
+   A valid bytes <x:ref>range-spec</x:ref> is satisfiable if it is either:
 </t>
+<ul>
+   <li><t>an <x:ref>int-range</x:ref> with a <x:ref>first-pos</x:ref> that
+       is less than the current length of the selected representation;
+       or,</t></li>
+   <li><t>a <x:ref>suffix-range</x:ref> with a non-zero
+       <x:ref>suffix-length</x:ref>.</t></li>
```
reschke, line 7891:
```suggestion
   <li>an <x:ref>int-range</x:ref> with a <x:ref>first-pos</x:ref> that
       is less than the current length of the selected representation;
       or,</li>
   <li>a <x:ref>suffix-range</x:ref> with a non-zero
       <x:ref>suffix-length</x:ref>.</li>
```

reschke (changes requested):
I htink this would be an irmprovement, but we need to think about the new 400 requirement.

draft-ietf-httpbis-semantics-latest.xml
royfielding: I don't consider that to be a new requirement. An invalid ranges specifier is something that the specification already disallowed as a bad message, and the standard code response for "client sent a bad message" is 400. The SHOULD is because servers are allowed to ignore Range before parsing it.

draft-ietf-httpbis-semantics-latest.xml
royfielding: done

draft-ietf-httpbis-semantics-latest.xml
reschke: I do not disagree, but we have a new SHOULD here :-).

royfielding: okay, have a look at https://github.com/httpwg/http-core/pull/1012/commits/f654eff36c7de2dc0d383e303ab3842a4dc9493a

reschke (approved):
Thanks!

